PR #2, opened 2015-02-09
Fixes #1 Replace occurrences of let with var

### LINKED ISSUE #1: Replace occurrences of let with var
The current version uses `let`. Unfortunately `let` is not available on WebKit so this issue is about replacing all occurrences of `let` with `var`.

This makes me a little sad but I think there is no other way around this.

st3fan:
This patch replaces all occurences of let with var.

I'm open to suggestions to handle this differently but I honestly don't think there is a way around this. WebKit simply does not support let.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -94,7 +94,7 @@ Readability.prototype = {
    * @return void
   **/
   _postProcessContent: function(articleContent) {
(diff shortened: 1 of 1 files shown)


leibovic:
This looks good to me. r+!

PR #4, opened 2015-02-09
Readability.parse() also return the uri

st3fan:
This patch lets `Readability.parse()` also return the uri that it was passed on creation time. This is a backward compatible change since it is just a new unique name in the response dictionary.


leibovic:
I also think this is fine. This would actually help the desktop/Android code as well, since there are currently places in the code where we create a new URI from the article URL.

PR #6, opened 2015-02-14
Bug 958735 Function purgeNode moved

srlakhe:
Submitting the formatted patch

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -412,7 +412,7 @@ Readability.prototype = {
 
     //helper function used below in the 'while' loop:
     function purgeNode(node, allElements) {
-      for (let i = node.childNodes.length; --i >= 0;) {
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -410,6 +410,14 @@ Readability.prototype = {
     // Check if any "dir" is set on the toplevel document element
     this._articleDir = doc.documentElement.getAttribute("dir");
 
+    //helper function used below in the 'while' loop:
+    function purgeNode(node, allElements) {
+      for (let i = node.childNodes.length; --i >= 0;) {
```
leibovic: Please use 'var' instead of 'let' (I know you used 'let' in your original patch, but since then we've decided to make this shared script work in a more generic web context, where 'let' isn't supported).

srlakhe:
Updated the patch.
Can you please check if it was done properly?

PR #7, opened 2015-02-22
Create "compare view" test page for easier development

leibovic:
I created a side-by-side view of an original page and its readerized version. Because of same-origin policy restrictions, test pages need to be hosted on the same domain, so I also created a directory where we can start putting testcases to just develop locally. I started with a copy of a page from my blog, since I know the author (me) is fine with us including this in our test content.

screen shot 2015-02-22 at 11 07 01 am


leibovic:
@st3fan I don't think we have any official review policy is in this repo, but this doesn't actually change Readability.js at all, so I'm just going to merge it :)

I think that any changes to Readability.js itself should go through a peer review, but changes that just strictly improve the development environment can just land, and we can always improve things as we go.

Issue #8, opened 2015-02-22
Clarify licence situation


DavidBruant:
The original code is Apache2, but I imagine additional work by Moz folks will be MPL2

leibovic:
I think that what starts as Apache2 stays Apache2, even if it has been further developed in mozilla-central. But let's ask @gerv :)

gerv:
This issue lacks context, but the general rule is: if Mozilla people enhance some e.g. Apache2 code, we make our changes Apache2 so they can be reincorporated into the upstream project. This applies to all permissive licenses, e.g. MIT or BSD, which are on code we get from elsewhere.

If it's our code (e.g. from Gaia), then yes, we keep it Apache2 as well.

Does that help?

leibovic:
Thanks for the response, and I apologize for not explaining the situation more clearly.

Here's some more context: A few years ago we landed this Apache2 code (Readbility.js) in mozilla-central, and since then we have heavily modified it. The whole time we kept the original Apache2 license header in the file, and then now we've moved this file back into its own repo (this repo), so that it can be shared by other applications (e.g. Firefox for iOS).

So it sounds like the answer here is that this library is (and always has been) Apache2. We should add a LICENSE file to the repo to make that more clear.

gerv:
leibovic: yes, that sounds absolutely right.

Gerv

PR #14, opened 2015-02-25
Update README to include license

leibovic:
I took this license block from the Readability.js license header.

This should address this issue #8.


fwenzel:
driveby r+

leibovic:
I probably could have just landed this without a review, but I appreciate it :)

Issue #9, opened 2015-02-22
Document how the algorithm works


DavidBruant:
Even if only very briefly in the readme

kwill:
Perhaps include an example of a minimal semantic page structure that would get Readerized? (For people that want to explicitly target Reader.)

e.g. (assuming this were correct)

```
<html>
<title>...</title>
<body>
<article>
<h1>...</h1> <!-- included -->
<p>...</p>  <!-- included -->
</article>
<nav>...</nav>  <!-- excluded -->
</body>
</html>
```

gijsk:
The problem is that it's not that simple... the algorithm is largely heuristics-based, on class names and IDs and the like. We can't rely on using semantic page structures because most websites don't use them, and some actively abuse them, e.g. using dozens or even hundreds of `<article>` elements to indicate links to articles on an overview page, people using `<h2>` to use larger text for the 'lead' paragraph of an article (issue #281) , etc.

kwill:
Thanks for the info, I get that you're bravely parsing the real, messy web. I guess I'm wondering if the heuristics imply a family of ideal templates, that could in theory be targeted? And answer questions like these https://stackoverflow.com/questions/30730300/optimize-website-to-show-reader-view-in-firefox more concretely. (btw semantic stuff is just my preference)

I guess I'm concerned that if my page is Readerable today, but the heuristics evolve, it might be un-Readerable tomorrow. And my pages are not important in a global sense (compared to, say, Wikipedia or NY Times), so I wouldn't want to write a ticket just for them! It'd be easier to just target an "ideal" template (even if the ideal evolves).

Is there a concern that Readerability itself could/would be abused? (I can't imagine how/why, but I guess spammer-types have an evil ingenuity.)

dgellow:
FYI, this article does a good job as describing how the library works from a high level: https://videoinu.com/blog/firefox-reader-view-heuristics/

PR #20, opened 2015-03-13
Added basic functional test + travis setup.

n1k0:
This is WiP adding basic functional testing setup and travis-ci integration.

screen shot 2015-03-13 at 17 12 52

I'm not super happy with the way it's done; comparing huge html strings is never easy. Though this might serve our purpose on the short term, what do you think?

Note that we should add many other reference HTML pages to make this interesting, though this is admittedly a painful process, especially as many pages are actually broken using Readability, and we won't want to add them as reference :(

@leibovic thoughts,feedback=?

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,37 @@
+var scrape = require("readable-proxy").scrape;
+var path = require("path");
+var fs = require("fs");
+var prettyPrint = require("html").prettyPrint;
+var chai = require("chai");
+var chaiAsPromised = require("chai-as-promised");
+chai.should();
+chai.use(chaiAsPromised);
+var expect = chai.expect;
+
+var testPageRoot = path.join(__dirname, "test-pages");
+var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
+  return {
+    dir: dir,
+    source: path.join(testPageRoot, dir, "source.html"),
+    expected: path.join(testPageRoot, dir, "expected.html"),
+  };
+});
+
+describe("Test page", function() {
+  testPages.forEach(function(testPage) {
+    describe(testPage.dir, function() {
+      it("should render as expected", function() {
+        // Allows up to 10 seconds for parsing to complete.
+        // XXX: Scraping is damn slow. Investigate.
+        this.timeout(10000);
+        var expected = fs.readFileSync(testPage.expected, {encoding: "utf-8"});
+        return scrape("file://" + testPage.source).catch(function(err) {
+          throw err;
+        }).then(function(result) {
+          // normalize html
+          return prettyPrint(result.content);
+        }).should.eventually.become(prettyPrint(expected));
+      });
+    });
+  })
+});
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .DS_Store
+npm-debug.log
+node_modules
 .metadata
 *.pyc
 *~
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: node_js
+sudo: false
+node_js:
+  - '0.10'
--- a/package.json
+++ b/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "readability",
+  "version": "0.0.1",
+  "description": "A standalone version of the readability library used for Firefox Reader View.",
+  "main": "Readability.js",
+  "scripts": {
+    "test": "mocha"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mozilla/readability"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/mozilla/readability/issues"
+  },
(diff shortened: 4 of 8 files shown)


leibovic:
I agree this isn't great, since these tests can fail very easily with a small change to Readability.js. That being said, we can use a failure as a "something changed!" signal, take a look at the output diff ourselves, and if to a human it looks to be okay (perhaps even improved), we could update the expected value for that testcase. Yes, this is a very human process, and not scalable in the long term, but it will at least give us *something* to work with.

I think that this is strictly better than our current situation (no tests), so we should merge this and see what kinds of failures we get when we start changing Readability.js. Seeing what happens in practice will give us a clearer sense of what kinds of changes result in what kinds of differences, and maybe we can account for that in the test (something like the "fuzzy diff" that dolske suggested).

test/index.js
```diff
@@ -0,0 +1,37 @@
...
+  };
+});
+
+describe("Test page", function() {
+  testPages.forEach(function(testPage) {
+    describe(testPage.dir, function() {
+      it("should render as expected", function() {
+        // Allows up to 10 seconds for parsing to complete.
+        // XXX: Scraping is damn slow. Investigate.
+        this.timeout(10000);
```
leibovic: Can you elaborate on why we need this scraping? I'm not as familiar with server-side JS, but is there a way to create a document from a local HTML file, and then run Readability on that directly?

Also, will errors in Readability.js propagate here? I could see us failing tests because of accidentally introducing JS errors, and it would be nice to be able to see those.

DavidBruant:
> I'm not super happy with the way it's done; comparing huge html strings is never easy

What about (DOM) tree diffing? Maybe using one of these:
https://www.npmjs.com/package/virtual-dom#diff-computation
https://www.npmjs.com/package/jqgram

test/index.js
DavidBruant: >  I'm not as familiar with server-side JS, but is there a way to create a document from a local HTML file, and then run Readability on that directly?

Indeed. It's a bit strange. I think the main issue is that there is a circular dependency. With this PR, readability depends on readable-proxy and readable-proxy depends on readability (via the `READABILITY_LIB_PATH` env variable).
IMO the first thing to do would be making Mozilla's readability its own NPM package. Is it planned?
If it isn't, I will do it myself, because I will need it very soon (although it'd be completely inappropriate given Mozilla should be the entity releasing readability on NPM :-/ )

Then the tests should be self-contained to this repo (without dep to readable-proxy) via reading files from the file system. Among other things, this will ensure the latest version of readability will be the one tested (and not the one linked from by readable-proxy which might be different)

n1k0: To be honest that was the shortest plan I could find to move fast, as most of required code was already available in readable-proxy.

Also don't forget we need a real DOM to run Readability, and node doesn't ship with one. So while I could possibly fire jsdom, karma, selenium (!) and so on, I found easier to reuse what's been made available by the proxy.

To me the main issue is the quality of Readability itself, both from an API PoV (which is debatable) and an accuracy one. Actually I'm finding more and more case where it behaves completely wrong, eg. stripping or *reordering* large portions of text, eg.:

screen shot 2015-03-14 at 21 37 38

This is like voodoo to me, the most worrying part is that after a few days trying to figure out how the lib works precisely, I'm still struggling. That's no good sign.

Anyway, I'll try to come with a better testing infrastructure on Monday, for now I'll try enjoying my week-end :p

n1k0:
> What about (DOM) tree diffing?

You know what, I wasted hours trying to find something suitable for this, and never founds these two guys. On one hand, I'm happy you did, on the other I'm a little pissed ^^'

Tree diffing is definitely the way to go.

Oh and btw, if you're willing to contribute to this thing you're *more* than welcome! :)

DavidBruant:
> Oh and btw, if you're willing to contribute to this thing you're more than welcome! :)

I'm planning to as it's a crucial part of https://github.com/MyWebIntelligence/MyWebIntelligence
We're currently the [Readability Parser API](https://www.readability.com/developers/api/parser), but this is an unnecessary 3rd-party dependency. We'd feel much safer using a standalone lib. Even more one actually used broadly like Firefox's one is de facto required to.

> This is like voodoo to me, the most worrying part is that after a few days trying to figure out how the lib works precisely, I'm still struggling. That's no good sign.

I didn't spend hours on it, but I did spend some time and got to the same conclusion. This is what led me to opening #9 
A few weeks back, I dove in to understand the whole problem. Ended up finding http://stackoverflow.com/a/4240037 which is a very good start to understand how this all works (the person answering wrote a PhD on the topic, there is even a video of him talking about it).

I certainly will have time to look into this in April. Maybe we should schedule a call to coordinate (well.. or at least try to understand how the code works together :-p).

pdehaan:
There is also [**dom-compare**](https://www.npmjs.com/package/dom-compare), which looks pretty interesting:
> _NodeJS module to compare two DOM trees_

leibovic:
I'm going to go ahead and merge, since it's better than what we currently have (nothing).

PR #21, opened 2015-03-17
Fix bug 1127778 - paragraph reordering caused by removing items in-loop

gijsk:


gijsk:
The fact that this is relying on readability-proxy.js doesn't make this a lot easier, and it's not super-clear to me why this is necessary... but that can be a separate issue.

PR #28, opened 2015-03-18
Switched to using jsdom for tests.

n1k0:
Faster, stronger, better =)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,9 +18,8 @@
   "homepage": "https://github.com/mozilla/readability",
   "devDependencies": {
     "chai": "^2.1.*",
-    "chai-as-promised": "^4.3.*",
     "html": "0.0.*",
-    "mocha": "^2.2.*",
-    "readable-proxy": "1.3.*"
+    "jsdom": "^3.1.2",
+    "mocha": "^2.2.*"
   }
 }
--- a/test/index.js
+++ b/test/index.js
@@ -1,11 +1,8 @@
 var path = require("path");
 var fs = require("fs");
+var jsdom = require("jsdom");
 var prettyPrint = require("html").prettyPrint;
-var chai = require("chai");
-var chaiAsPromised = require("chai-as-promised");
-chai.should();
-chai.use(chaiAsPromised);
-var expect = chai.expect;
+var expect = require("chai").expect;
 
 var testPageRoot = path.join(__dirname, "test-pages");
 var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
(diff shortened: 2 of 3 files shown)


test/index.js
```diff
@@ -17,32 +14,30 @@ var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
...
+        jsdom.env(
+          fs.readFileSync(testPage.source, {encoding: "utf-8"}),
+          [path.join(__dirname, "..", "Readability.js")],
+          function (errors, window) {
+            if (errors) {
+              throw new Error(errors);
+            }
+            var uri = {
+              spec: "http://fakehost/test/page.html",
+              host: "fake-host",
```
pdehaan: "fake-host", or "fakehost" (no dash)?

n1k0: Good catch!

test/index.js
```diff
@@ -17,32 +14,30 @@ var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
...
+            if (errors) {
+              throw new Error(errors);
+            }
+            var uri = {
+              spec: "http://fakehost/test/page.html",
+              host: "fakehost",
+              prePath: "http://fakehost",
+              scheme: "http",
+              pathBase: "http://fakehost/test"
+            };
```
pdehaan: Unrelated, but I guess I've always been a bit confused by this `uri` object thing.
It seems that all the fields can be extracted from the `spec` string. It'd be nice if there were just some helper function which converts a fully qualified URL into a `host`, `prePath`, `scheme`, and `pathBase` for you.

I'm guessing this would be pretty trivial w/ Node.js' `url.parse()` function, but it may fit better in the Readability.js library.

```js
var url = require('url');

function uriFromString(str) {
  var urlObj = url.parse(str, true);
  var protocol = urlObj.protocol.replace(':', '');
  var prePath = protocol + '://' + urlObj.hostname;
  var uri = {
    spec: urlObj.href,
    host: urlObj.hostname,
    pathBase: prePath + urlObj.pathname.replace(/[^\/]+$/, ''),
    prePath: prePath,
    scheme: protocol
  };
  return uri;
}

console.log(JSON.stringify(uriFromString(uri.spec), null, 2));
```

I think we've also used http://medialize.github.io/URI.js/ in other projects (via Bower).

gijsk: I think considering this is test infrastructure and we're heavily doubling down on actually improving the lib, we can leave it like this for now and have further discussions regarding how best to generate this JSON blurb with URL info in a separate issue, if that sounds OK. :-)

pdehaan: Absolutely! Filed as #29.

package.json
```diff
@@ -18,9 +18,8 @@
   "homepage": "https://github.com/mozilla/readability",
   "devDependencies": {
     "chai": "^2.1.*",
-    "chai-as-promised": "^4.3.*",
     "html": "0.0.*",
-    "mocha": "^2.2.*",
-    "readable-proxy": "1.3.*"
+    "jsdom": "^3.1.2",
+    "mocha": "^2.2.*"
```
gijsk: Nit: keep trailing comma unless package.json's spec is stupid and doesn't allow you to do that or whatever

n1k0: package.json is stupid :/

n1k0:
Now that I've added support for log messages printing onto the test console.I think this is ready for review.

I'd like both r+ from @leibovic and @gijsk :)

PR #31, opened 2015-03-19
Allow generating tests from the web, make testing more closely match Firefox

gijsk:
r? @leibovic @n1k0 

I realize that this perpetuates the use of a less-powerful version of jsdom, but unfortunately it seems like jsdom is (a) huge and (b) breaks assumptions that readability.js has (about node lists being immutable), and (c) is geared to node.js, so importing that into m-c seems like the worse option at least for the short-to-medium term.

Then of course I had to make JSDOMParser actually work with node.js, which was "fun". I found some issues with script parsing and whitespace handling which I fixed. I've tried to make it obvious what I changed by separating out the commits.


test/test-pages/wapo-1/expected.html
```diff
@@ -0,0 +1 @@
...nistry should be trained and equipped to fight and counter this militancy.”</p> <p>The last major attack on a civilian target in Tunisia was in 2002, when al-Qaeda militants killed more than 20 people in a car bombing outside a synagogue in the city of Djerba.</p> <p>Heba Habib contributed to this report.</p>    <p channel="wp.com"> <b>Read more:</b> </p> <p channel="wp.com"> <a href="http://www.washingtonpost.com/world/middle_east/tunisias-islamists-get-sobering-lesson-in-governing/2014/11/20/b6fc8988-65ad-11e4-ab86-46000e1d0035_story.html" title="www.washingtonpost.com">Tunisia’s Islamists get a sobering lesson in governing</a> </p> <p channel="wp.com"> <a href="http://www.washingtonpost.com/world/national-security/tunisia-after-igniting-arab-spring-sends-the-most-fighters-to-islamic-state-in-syria/2014/10/28/b5db4faa-5971-11e4-8264-deed989ae9a2_story.html">Tunisia sends most foreign fighters to Islamic State in Syria</a> </p> <p channel="wp.com"> <a href="http://www.washingtonpost.com/blogs/worldviews/wp/2015/03/18/tunisias-bardo-museum-attacked-by-terrorists-is-home-to-amazing-roman-treasures/">Tunisia’s Bardo museum is home to amazing Roman treasures</a> </p>  </article></div>
```
n1k0: Couldn't we have these indented? Humans might want to read them too =)

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+ctx.console = console;
+vm.runInNewContext(fs.readFileSync(jsdomPath), ctx, jsdomPath);
+vm.runInNewContext(fs.readFileSync(readabilityPath), ctx, readabilityPath);
+var Readability = ctx.Readability;
+var JSDOMParser = ctx.JSDOMParser;
+
+
+if (process.argv.length < 3) {
+  console.log("Need at least a destination slug and potentially a URL (if the slug doesn't have source).");
+  return 0;
```
n1k0: Nit: `process.exit(0)` is usually more idiomatic in node-land

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+var slug = process.argv[2];
+var url = process.argv[3];
+
+var destRoot = path.join(__dirname, "test-pages", slug);
+
+fs.mkdir(destRoot, function(err) {
```
n1k0: Any specific reason we're not using `mkdirSync` here?

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+  if (err) {
+    var sourceFile = path.join(destRoot, "source.html");
+    fs.exists(sourceFile, function(exists) {
+      if (exists) {
+        fs.readFile(sourceFile, {encoding: "utf-8"}, function(err, data) {
```
n1k0: Same remark, `fs.readFileSync` might be jyst god enough while more readable.

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+  fetchSource();
+});
+
+function fetchSource() {
+  if (!url) {
+    console.error("You should pass a URL if the source doesn't exist yet!");
+    return;
+  }
+  var client = http;
```
n1k0: Nit: I'd keep `http`.

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+  if (url.indexOf("https") == 0) {
+    client = require("https");
+  }
+  client.get(url, function(response) {
+    console.log("STATUS:", response.statusCode);
+    console.log("HEADERS:", JSON.stringify(response.headers));
```
n1k0: Nit: Would be nice to have debug logging optional.

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+      rv += chunk;
+    });
+    response.on("end", function() {
+      console.log("End received");
+      onGotSource(rv);
+    });
+  });
+}
+
+function onGotSource(source) {
```
n1k0: Nit: onResponseReveived

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+  try {
+    var result = new Readability(uri, doc).parse();
+  } catch (ex) {
+    console.error(ex);
+  }
+  if (!result) {
+    console.error("No content generated by readability, not going to write expected.html!");
+    return;
+  }
+  fs.writeFile(destPath, result.content, function(err) {
```
n1k0: Nit: writeFileSync?

test/generate-testcase.js
```diff
@@ -0,0 +1,117 @@
...
+          }
+          onGotSource(data);
+        });
+      } else {
+        fetchSource();
+      }
+    });
+    return;
+  }
+  fetchSource();
```
n1k0: I was expecting to see onGotSource passed as a callback here. I suspect this would ease reading the async flow, though that may be just me.

JSDOMParser.js
```diff
@@ -0,0 +1,950 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
n1k0: Feels like this should go on its own node package eventually, and readability should have it as a dependency. Not now for sure ;)

Readability.js
```diff
@@ -668,7 +668,8 @@ Readability.prototype = {
           s -= 1;
           sl -= 1;
 
-          if (siblingNode.nodeName !== "DIV" && siblingNode.nodeName !== "P") {
+          if (siblingNode.nodeName !== "DIV" && siblingNode.nodeName !== "P" &&
+              siblingNode.nodeName !== "ARTICLE" && siblingNode.nodeName !== "SECTION") {
```
n1k0: Nit: how about:

```js
"DIV,P,ARTICLE,SECTION".split(",").indexOf(siblingNode.nodeName) === -1;
```

test/index.js
```diff
@@ -1,9 +1,19 @@
 var path = require("path");
 var fs = require("fs");
-var jsdom = require("jsdom");
 var prettyPrint = require("html").prettyPrint;
 var expect = require("chai").expect;
 
+var vm = require("vm");
+var readabilityPath = path.join(__dirname, "..", "Readability.js");
+var jsdomPath = path.join(__dirname, "..", "JSDOMParser.js");
```
n1k0: The three lines above are worth a comment methinks.

n1k0:
> jsdom is (a) huge (b) breaks assumptions that readability.js has (about node lists being immutable), and (c) is geared to node.js, so importing that into m-c seems like the worse option at least for the short-to-medium term.

TBH, I wasn't expecting we land Readability tests into m-c. Do you plan to run them as part of the build? Because it requires nodejs anyway, and imoe that's asking for trouble (been there, that's a tough topic) :p

Also, while I much like the idea of generating test cases, that requires ensuring that what's generated is accurate; adding green tests for broken pages could be the worst possible idea. Though with proper checks, documentation and reviews, that shouldn't be much of an issue.

The patch looks generally good, r=@n1k0 with comments addressed, though I'd like a counter-review by @leibovic before we land this.

test/generate-testcase.js
gijsk: Sync IO is evil. :-)

test/test-pages/wapo-1/expected.html
gijsk: I'm pretty sure this is what the Washington Post sends us (minus the going through jsdom and back), so I just didn't touch it. Is it noticeably worse than the actual article?

test/generate-testcase.js
gijsk: I don't understand what you mean by this comment, can you clarify?

test/generate-testcase.js
gijsk: This isn't only called for http responses.

gijsk:
> Also, while I much like the idea of generating test cases, that requires ensuring that what's generated is accurate; adding green tests for broken pages could be the worst possible idea. Though with proper checks, documentation and reviews, that shouldn't be much of an issue.

Without tests for existing behaviour, even if the behaviour is suboptimal, we don't know what we break if we change anything in the algorithm. That makes changing the algorithm super scary. If we can see the changes (so for instance, this showed me that the different DOM processing meant we suddenly changed ARTICLE tags into DIVs), then we can decide whether the changes are OK or not. IMO that is always better than not knowing.

test/generate-testcase.js
n1k0: For an util script like this, I don't think blocking is much of a concern. Well, your call.

gijsk:
> TBH, I wasn't expecting we land Readability tests into m-c. Do you plan to run them as part of the build? Because it requires nodejs anyway, and imoe that's asking for trouble (been there, that's a tough topic) :p

My point regarding not wanting to import jsdom was more to justify why I'm importing the mozilla thing into github and not the other way around. I want the behaviour to be the same as much as possible (ie if it works in readability in Firefox, it should work in the testcase. That wasn't the case before my changes, which was causing issues with my trying to create some set of comparison pages so we can evaluate algorithm changes.

test/generate-testcase.js
n1k0: I wouldn't alias `http` as `client`. No big deal, really.

test/test-pages/wapo-1/expected.html
n1k0: All generated html is onelined. I suspect not every website sends content that way.

n1k0:
> My point regarding not wanting to import jsdom was more to justify why I'm importing the mozilla thing into github and not the other way around

Ah ok, makes much sense now.

test/test-pages/wapo-1/expected.html
gijsk: Huh, that's arguably a bug in jsdom. I will look at using prettyPrint to fix this.

Aside: I found prettyPrint to not be idempotent (ie prettyPrint(x) != prettyPrint(prettyPrint(x))) which is annoying...

test/generate-testcase.js
gijsk: Someone downvoted me for suggesting it on stackoverflow a long time ago. It still hurts. ;-)

test/generate-testcase.js
gijsk: Ah, but a few lines down we use "https" if the URL requires it (because http.get complains if you pass it an https URL and refuses to work, because... reasons, I guess?). So I figured I would make the object we use to do the request use a different name than one specific to the protocol...

test/test-pages/wapo-1/expected.html
n1k0: > Aside: I found prettyPrint to not be idempotent (ie prettyPrint(x) != prettyPrint(prettyPrint(x))) which is annoying...

Found this as well, I'm working on integrating domcompare as suggested by @pdehaan.

test/generate-testcase.js
n1k0: Missed that, makes sense!

gijsk:
Kept the sync stuff for now so as not to change too much in something that works otherwise, adopted the other comments, I think (renamed the response handling function, passed some more stuff around instead of relying on globals, added comments and `process.exit` calls).

JSDOMParser.js
```diff
@@ -0,0 +1,950 @@
...
+        }
+      }
+
+      // If this is a self-closing tag, read '/>'
+      var closed = tag in voidElems;
+      if (c === "/") {
+        closed = true;
+        c = this.nextChar();
+        if (c !== ">") {
+          debugger;
```
leibovic: Should this debugger statement be removed?

leibovic:
This looks good to me. Importing JSDOMParser from mozilla-central makes sense for the short term, where our goal is to support Readability in Firefox.

However, we should file an issue to look into how Readability performs on iOS, and whether any assumptions about the DOM we get from JSDOMParser cause problems there (/cc @st3fan).

PR #35, opened 2015-03-19
Check metadata in tests

leibovic:
r? @gijsk 

I just generated the expected-metadata.json files with `JSON.stringify(result)` (after deleting some things we didn't care about).


test/test-pages/missing-paragraphs/expected-metadata.json
```diff
@@ -0,0 +1 @@
+{"title":"Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy\n            eirmod tempor invidunt","byline":null,"excerpt":"Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy\n            eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam\n            voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet\n            clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit\n            amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam\n            nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat,\n            sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum.\n            Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor\n            sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed\n            diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat,\n            sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum.\n            Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor\n            sit amet."}
```
pdehaan: Disclaimer: I don't know anything about anything, but why all the newlines (`\n`) in the title and excerpt?

test/test-pages/002/expected-metadata.json
```diff
@@ -0,0 +1 @@
+{"title":"This API is so Fetching! ✩ Mozilla Hacks – the Web developer blog","byline":"blog.nikhilism.com","excerpt":"For more than a decade the Web has used XMLHttpRequest (XHR) to achieve asynchronous requests in JavaScript. While very useful, XHR is not a very ..."}
```
pdehaan: Probably not worth the effort, but these may be more human readable if we pretty-fy them (ie: `JSON.stringify(data, null, 2)`).

leibovic: Good call, I just updated the PR to do that.

test/test-pages/001/expected-metadata.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Get your Frontend JavaScript Code Covered | Code",
+  "byline": "Nicolas Perriault —",
+  "excerpt": "width=device-width"
```
n1k0: Feels odd to mark this kind of erroneous result as "expected". I get that we want a starting point, though I hope we won't forget to fix these ;)

n1k0:
FWIW, I love this patch.

test/test-pages/001/expected-metadata.json
leibovic: Yeah, this exposes some real issues :/ I've seen some similar problems in the reading list UI on Fennec. Well we can start by trying to fix these testcases!

n1k0: This one shouldn't be too hard to fix*, as the html for this page is:

```html
    <meta name="description" content="Nicolas Perriault's homepage.">
    <meta name="viewport" content="width=device-width">
```
It seems the lib is simply targeting the wrong meta tag.

* famous last words

leibovic: Doh: https://github.com/mozilla/readability/pull/36

This code was originally written with `let`, but this broke when we switched everything to `var`.

PR #37, opened 2015-03-20
Remove reliance on inert nodecollection

gijsk:
This gets rid of the node collection stuff for the main loop, which should solve a lot of issues. The rest of the code (that uses getElementsByTagName) doesn't actually try to remove things from the array, so should be OK.

There are other things we can do now that the loop uses some more sane constructs and the DOM is more powerful, but I've held off on that for now.

This change in and of itself passes all our existing tests.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -285,18 +285,23 @@
     nodeName: null,
     parentNode: null,
     textContent: null,
+    nextSibling: null,
+    previousSibling: null,
 
     get firstChild() {
       return this.childNodes[0] || null;
     },
 
-    get nextSibling() {
-      if (this.parentNode) {
-        var childNodes = this.parentNode.childNodes;
-        return childNodes[childNodes.indexOf(this) + 1] || null;
-      }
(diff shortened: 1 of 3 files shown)


gijsk:
(oh, and purely based on some totally unscientific tests in my console, it seems to be approximately 30% faster at running all our tests)

gijsk:
@leibovic, @n1k0 thoughts/reviews? :-)

leibovic:
@thebnich, can you take a look at these JSDOMParser changes?

leibovic:
We discussed this in person, but for posterity: I think it's fine to merge this, since it's a strict improvement over what we currently have, and we can continue to hack on this tomorrow before merging things into fx-team.

leibovic:
Aside: I wonder if it would be worth some simple unit tests for JSDOMParser. Maybe not, if we're going to move away from using this, but it could be nice to test the new node manipulation logic.

gijsk:
Per IRL discussion, merging first, doing tests/checks later!

PR #40, opened 2015-03-20
Improve byline algorithm

leibovic:
Still a WIP, as I'll have to rebase on top of https://github.com/mozilla/readability/pull/37

This fixes at least these two testcases:
http://www.salon.com/2015/01/02/coca_colas_anti_american_outsourcing_scheme_how_big_soda_gets_the_public_to_shoulder_its_costs/
http://alistapart.com/article/designing-for-easy-interaction

Using the author meta tag is definitely a strict improvement, as is checking the nodes before throwing them away. In order to get the salon.com testcase to work, I added logic for preferring a rel="author" element, but I don't know if that's too specific to that testcase.

Also, I don't like the `var self = this;` crap, but it was the quickest way to getting this to work to test :)

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -434,6 +434,23 @@ Readability.prototype = {
     return node && node.nextElementSibling;
   },
 
+  _checkByline: function(node, matchString) {
+    if (this._articleByline) {
+      return false;
+    }
+
+    if (node.getAttribute !== undefined) {
+      var rel = node.getAttribute("rel");
+    }
+
+    if ((rel === "author" || this.REGEXPS.byline.test(matchString)) && this._isValidByline(node.textContent)) {
+      this._articleByline = node.textContent.trim();
+      return true;
+    }
+
+    return false;
+  },
+
   /***
    * grabArticle - Using a variety of metrics (content score, classname, element types), find the content that is
    *         most likely to be the stuff a user wants to read. Then return it wrapped up in a div.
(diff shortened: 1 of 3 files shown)


leibovic:
I updated this to not favor the rel="author", since that was causing one of the tests to fail (a valid failure).

leibovic:
With the new `_removeAndGetNext` logic, I realize this still has the problem of not checking the children of nodes that we're about to remove (e.g. digging through a header node to look for a byline). Perhaps we should try recursively looking through nodes before removing them if we haven't found a byline yet.

gijsk:
> With the new _removeAndGetNext logic, I realize this still has the problem of not checking the children of nodes that we're about to remove (e.g. digging through a header node to look for a byline). Perhaps we should try recursively looking through nodes before removing them if we haven't found a byline yet.

This sounds like a good idea for a followup, as discussed IRL.

Readability.js
```diff
@@ -765,10 +782,10 @@ Readability.prototype = {
    * If no excerpt is found, an empty string will be
    * returned.
    *
-   * @param Element - root element of the processed version page
    * @return String - excerpt of the article
   **/
-  _getExcerpt: function(articleContent) {
+  _getArticleMetadata: function() {
```
gijsk: Please update the comment. :-)

Readability.js
```diff
@@ -434,6 +434,24 @@ Readability.prototype = {
...
   },
 
+  _checkByline: function(node, matchString) {
+    if (this._articleByline) {
+      return false;
+    }
+
+    var rel = null;
+    if (node.getAttribute !== undefined) {
+      rel = node.getAttribute("rel");
```
gijsk: Because this is `var` and not `let` you can just use `var rel =` in the loop and it'll be `undefined` otherwise (which is OK).

Readability.js
```diff
@@ -1543,14 +1557,20 @@ Readability.prototype = {
     //   }).bind(this), 500);
     // }
 
-    var excerpt = this._getExcerpt(articleContent);
+    // No description meta tags, use the article's first paragraph.
```
gijsk: Maybe update this comment to indicate what we're using it for?

gijsk:
This generally looks great, just a few nits, for which, see my earlier comments. :-)

PR #42, opened 2015-03-20
Better test docs.

n1k0:
Removed the `process.env.TESTFILTER` trick as it's natively provided by mocha with its --grep option. Updated README, added Travis badge.


n1k0:
@gijsk @leibovic r?

gijsk:
LGTM!

PR #46, opened 2015-03-20
Write a bunch of tests for jsdomparser

gijsk:
Fewer bugs is good!


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A standalone version of the readability library used for Firefox Reader View.",
   "main": "Readability.js",
   "scripts": {
-    "test": "mocha test/index.js",
+    "test": "mocha test/index.js test/test-jsdomparser.js",
```
n1k0: How about `mocha test/*_test.js` and append `_test.js` to the filename of whatever is a test file? That would ease adding more test files in the future.

gijsk: ... or prepend test- ? I guess it doesn't matter much, but either will mean moving index.js...

test/test-jsdomparser.js
```diff
@@ -0,0 +1,194 @@
...
+    expect(foo.previousSibling).eql(beforeFoo);
+    expect(foo.nextSibling).eql(afterFoo);
+    expect(foo.previousElementSibling).eql(beforeFoo);
+    expect(foo.nextElementSibling).eql(afterFoo);
+
+    expect(foo.previousSibling.nextSibling).eql(foo);
+    expect(foo.nextSibling.previousSibling).eql(foo);
+    expect(foo.nextSibling).eql(foo.nextElementSibling);
+    expect(foo.previousSibling).eql(foo.previousElementSibling);
+    done();
```
n1k0: As all operations within this test are synchronous, using `done` isn't required.

test/test-jsdomparser.js
```diff
@@ -0,0 +1,194 @@
...
+describe("Test JSDOM functionality", function() {
+  it("should work for basic operations with elements", function(done) {
+    expect(baseDoc.childNodes.length).eql(1);
+    expect(baseDoc.getElementsByTagName("*").length).eql(10);
+    var foo = baseDoc.getElementById("foo");
+    expect(foo.id).eql(foo.getAttribute("id"));
+    expect(foo.parentNode.localName).eql("body");
+    expect(baseDoc.body).eql(foo.parentNode);
+    expect(baseDoc.body.parentNode).eql(baseDoc.documentElement);
+    expect(baseDoc.body.childNodes.length).eql(3);
```
n1k0: How about increasing test granularity and organize the tests a little? That's usually helpful to get contextualized and descriptive information:

```js
describe("JSDOMParser", function() {
  describe("#parse()", function() {
    it("should return a Document", function() {
      expect(new JSDOMParser().parse(BASETESTCASE)).to.be.an.instanceOf(Document);
    });
  });
});

describe("Document", function() {
  var baseDoc;
  
  beforeEach(function() {
    baseDoc = new JSDOMParser().parse(BASETESTCASE);
  });

  describe("get childNodes()", function() {
    it("should return the expected number of child nodes", function() {
      expect(baseDoc.childNodes.length).eql(1);
    });
  });

  describe("#getElementsByTagName", function() {
    it("should return the expected number of tags", function() {
      expect(baseDoc.getElementsByTagName("*").length).eql(10);
    });
  });
  
  etc.

Comment (gijsk) [inline test/test-jsdomparser.js]: To me, it doesn't seem that useful tbh. It adds a lot of code and the behaviour isn't much better... I can try and add more `it()` wrappers, but adding the describe toplevel things as well seems not-super-useful?

**Comment by n1k0** [comment]
LGTM! I'd like @leibovic to give it a look too

File: test/test-jsdomparser.js [hunk 4]
```diff
@@ -0,0 +1,210 @@
...
+
+var baseDoc = new JSDOMParser().parse(BASETESTCASE);
+
+describe("Test JSDOM functionality", function() {
+  function nodeExpect(actual, expected) {
+    try {
+      expect(actual).eql(expected);
+    } catch (ex) {
+      throw ex.message;
+    }
```
leibovic: Why do we need this helper method instead of just calling expect directly? Does this have to do with those recursive problems you mentioned to me earlier?

leibovic:
Nice, this is great. I'm excited to have this test coverage!

test/test-jsdomparser.js
gijsk: Yes, NiKo and I discussed this on IRC. This seems to fix the issue with the test timing out / not completing if there are failures (which I get isn't the case right now, but it helps if stuff does break!).

PR #47, opened 2015-03-20
Drop whitespace-only textnodes immediately

gijsk:
This doesn't seem to require changes to the tests, which means they're basically already getting dropped later on (not sure where yet) - fine, but let's not waste time adding elements (which we'll then recurse into!).


gijsk:
@leibovic how does this look?

Issue #43, opened 2015-03-20
Publish to npm


DavidBruant:
So that other projects can install it via 
`npm install pick-your-name`

gijsk:
Considering the script itself isn't at all usable through node, I'm not convinced this is a good idea, at least right now...

DavidBruant:
> Considering the script itself isn't at all usable through node

You've got [Travis running the tests for Node 0.10](https://github.com/mozilla/readability/blob/master/.travis.yml), so in one way or another, the code is usable in Node (unless I missed some part where Firefox is started from Node to run the tests).

Also, looking at https://github.com/mozilla/readability/blob/d0df9d8479f40189fb39037c78574cc5c420f075/test/index.js#L11-L29 it looks like it would be fairly easy to wrap the existing code to make it work in Node (just isolate these lines in their own file, make this new file the `main` endpoint in `package.json` and publish that).
... wait... I'll just do it and send a PR...

gijsk:
Yeah, I mean that the vm stuff isn't particularly "nice"... you can't require("readability.js") or whatever.

DavidBruant:
I need this sort of things in a project I'm working on. Something that works would be better than nothing at all. The thing we use now is the [Readability Parser API](https://www.readability.com/developers/api/parser). As you can imagine, the vm stuff will certainly be a major improvement over this HTTP API in terms of perfs.

pdehaan:
If we put it on **npm** (YAY!!) can we please put it on **bower** too?

asbjornenge:
What is the status on this? Quite a lot of us are now using `npm` for client side package management.

gijsk:
I no longer think this is a bad idea, but equally I don't have time to push this forward significantly. @pdehaan, what would need to happen to get this listed on npm/bower as a mozilla thing?

pdehaan:
@gijsk Make sure you're logged into npm in your Terminal. Type `npm publish` in your local repo folder. That should do it for npm. Not sure about bower, I haven't published there in a while -- but I can take a look.

pdehaan:
Actually, the [**readability**](https://www.npmjs.com/package/readability) name is already used by arc90's module on npm, so we'd have to rename our module to avoid a naming conflict.
We could use something like **moz-readability**, or whatever.

gijsk:
Is there no way to associate the published module with Mozilla-the-org rather than with my personal NPM username ?

pdehaan:
[...]

Issue #29, opened 2015-03-18
Add method to convert a URI string to a uri object


pdehaan:
I've always been a bit confused by this `uri` object thing:

```js
var uri = {
  spec: "http://fakehost/test/page.html",
  host: "fakehost",
  prePath: "http://fakehost",
  scheme: "http",
  pathBase: "http://fakehost/test"
};
var result = new window.Readability(uri, window.document).parse();
```

It seems that all the fields can be extracted from the `spec` string. It'd be nice if there were just some helper function which converts a fully qualified URL into a `host`, `prePath`, `scheme`, and `pathBase` for you.

Or maybe modify the `window.Readability()` method to accept an object or string, and if it's a URI string, try and auto-convert it to an object.

I'm guessing this would be pretty trivial w/ Node.js' `url.parse()` function, but it may fit better in the Readability.js library as plain olde JavaScript.

```js
var url = require('url');

function uriFromString(str) {
  var urlObj = url.parse(str, true);
  var protocol = urlObj.protocol.replace(':', '');
  var prePath = protocol + '://' + urlObj.hostname;
  var uri = {
    spec: urlObj.href,
    host: urlObj.hostname,
    pathBase: prePath + urlObj.pathname.replace(/[^\/]+$/, ''),
    prePath: prePath,
    scheme: protocol
  };
  return uri;
}

console.log(JSON.stringify(uriFromString(uri.spec), null, 2));
```

I think we've also used http://medialize.github.io/URI.js/ in other projects (via Bower).

gijsk:
URI parsing is surprisingly treacherous ground (cf. https://bugzilla.mozilla.org/show_bug.cgi?id=483304 https://bugzilla.mozilla.org/show_bug.cgi?id=1093611 ). If the tests are the only place where we need to make one of these (because otherwise consumers are responsible for creating these), I'd be inclined to just leave it. OTOH, security/correctness concerns might be limited for test-only use as well?

Is there anything using this / some variant of this that isn't the tests?

(don't care about the real code *reading* the object - worried about how they're getting created though...)

PR #48, opened 2015-03-20
Package is now in a publishable state.

DavidBruant:
Moves #43 a step closer

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,40 @@
+"use strict";
+
+var path = require("path");
+var vm = require("vm");
+var fs = require("fs");
+
+var readabilityPath = path.join(__dirname, "Readability.js");
+var jsdomPath = path.join(__dirname, "JSDOMParser.js");
+
+var scopeContext = {};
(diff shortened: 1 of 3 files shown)


DavidBruant:
I'm happy to re-discuss the exposed API if necessary.

gijsk:
I think we should only expose + require Readability.js from the package file. Ideally in N days/weeks/months from now we can ditch having jsdomparser in this repo (either it can be in its own repo or we can start relying on jsdom or whatever).

DavidBruant:
Ok for having a document as argument.
I guess I'll make a request to change the API of mine as well. The uri argument should be a string, not the non-standard, non-obvious URI object.
I'll try to cook something from the `url` node module.

pdehaan:
@DavidBruant,

Re: `url` node module, see https://github.com/mozilla/readability/issues/29. I wrote it quickly, but it may work (or maybe not, thats the fun of software!)

gijsk:
This has conflicts and is superseded by #608, so I'm gonna close this out.

PR #49, opened 2015-03-20
Improve logic to rely on children instead of childNodes

gijsk:
@leibovic this knocks another 50-ish ms off our test runs, which is nice, and cleans things up a bit. Will be looking at some more logic/perf improvements after this.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -99,7 +99,8 @@ Readability.prototype = {
     videos: /https?:\/\/(www\.)?(youtube|vimeo)\.com/i,
     nextLink: /(next|weiter|continue|>([^\|]|$)|»([^\|]|$))/i,
     prevLink: /(prev|earl|old|new|<|«)/i,
-    whitespace: /^\s*$/
+    whitespace: /^\s*$/,
+    hasContent: /\S$/,
   },
 
   DIV_TO_P_ELEMS: [ "A", "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL", "SELECT" ],
@@ -480,7 +481,7 @@ Readability.prototype = {
       // First, node prepping. Trash nodes that look cruddy (like ones with the
       // class name "comment", etc), and turn divs into P tags where they have been
       // used inappropriately (as in, where they contain no other block level elements.)
-      var nodesToScore = [];
+      var elementsToScore = [];
       var node = this._doc.documentElement;
 
       while (node) {
@@ -504,7 +505,7 @@ Readability.prototype = {
         }
 
         if (node.tagName === "P" || node.tagName === "TD" || node.tagName === "PRE")
-          nodesToScore.push(node);
+          elementsToScore.push(node);
 
         // Turn all divs that don't have children block level elements into p's
         if (node.tagName === "DIV") {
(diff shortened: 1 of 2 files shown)


Readability.js
```diff
@@ -847,33 +846,28 @@ Readability.prototype = {
...
-          return -1;
-
-        if (pIndex >= 0)
-          return -1;
-
-        pIndex = i;
-      } else if (node.nodeType == Node.TEXT_NODE && this._getInnerText(node, false)) {
-        return -1;
+      if (node.nodeType == Node.TEXT_NODE &&
+          this.REGEXPS.hasContent.test(node.textContent)) {
```
leibovic: So you ended up using this regex for something after all :)

Readability.js
```diff
@@ -624,10 +623,10 @@ Readability.prototype = {
         // Move all of the page's children into topCandidate
         topCandidate = doc.createElement("DIV");
         neededToCreateTopCandidate = true;
-        var children = page.childNodes;
-        while (children.length) {
-          this.log("Moving child out:", children[0]);
-          topCandidate.appendChild(children[0]);
+        var kids = page.childNodes;
```
leibovic: Why are we still using childNodes here instead of children? If this is correct, I think we should update the comment above to talk about nodes instead of "children".

Readability.js
```diff
@@ -643,72 +642,71 @@ Readability.prototype = {
...
-          if (nodeLength > 80 && linkDensity < 0.25) {
-            append = true;
-          } else if (nodeLength < 80 && linkDensity === 0 && nodeContent.search(/\.( |$)/) !== -1) {
+          if (sibling.readability &&
+              ((sibling.readability.contentScore + contentBonus) >= siblingScoreThreshold)) {
             append = true;
+          } else if (sibling.nodeName === "P") {
+            var linkDensity = this._getLinkDensity(sibling);
+            var nodeContent = this._getInnerText(sibling);
+            var nodeLength = nodeContent.length;
```
leibovic: Let's be careful with these `var`s to make sure they don't hoist their way into causing problems.

Readability.js
```diff
@@ -643,72 +642,71 @@ Readability.prototype = {
...
+            this.log("Altering sibling:", sibling, 'to div.');
 
-            this._setNodeTag(siblingNode, "DIV");
+            this._setNodeTag(sibling, "DIV");
           }
 
           // To ensure a node does not interfere with readability styles,
           // remove its classnames.
-          siblingNode.removeAttribute("class");
+          sibling.removeAttribute("class");
```
leibovic: This is odd because we definitely don't do this for all elements.

gijsk: Yeah, I think it's just for the toplevel nodes that get stuck into the article content. :-\

Readability.js
gijsk: Because all the text nodes get stuck in the topCandidate too, which is important if, for instance, we just have a <body> with a bunch of text in it.

I don't know how much we care about that case, though...

Readability.js
```diff
@@ -643,72 +642,71 @@ Readability.prototype = {
...
+          // sibling is removed from the array when we call appendChild().
+          // As a result, we must revisit this index since the nodes
+          // have been shifted.
+          s -= 1;
+          sl -= 1;
         }
       }
 
-      this.log("Article content pre-prep: " + articleContent.innerHTML);
+      if (this.ENABLE_LOGGING)
```
leibovic: Isn't the log function itself already hidden behind a `this.ENABLE_LOGGING` check?

leibovic: Oh, or do you want to avoid making these `innerHTML` calls?

gijsk: > Oh, or do you want to avoid making these `innerHTML` calls?
Right in two. :-)

leibovic:
Looks good to me. Small steps towards making this all more readable! :+1:

Readability.js
n1k0: +1, btw why don't we use `forEach` for loops?

PR #50, opened 2015-03-21
Bug 1143725 - fix the Herald Sun website

gijsk:
@mhammond Especially for you! ;-)

@leibovic This adds a little extra work which on our set of testcases basically loses me the 50-ish ms that I gained in the previous commit... I think it's worth it but maybe we should add a cut-off for the parentNode searching or something (there is an implicit one right now in that we will stop when we hit nodes with no readability info)? Or instead look at the scores of the other candidates and look for a common ancestor that we can use if there's a bunch at pretty similar levels or something?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -634,6 +634,29 @@ Readability.prototype = {
         page.appendChild(topCandidate);
 
         this._initializeNode(topCandidate);
+      } else if (topCandidate) {
+        // Because of our bonus system, parents of candidates might have scores
+        // themselves. They get half of the node. There won't be nodes with higher
+        // scores than our topCandidate, but if we see the score going *up* in the first
+        // few steps up the tree, that's a decent sign that there might be more content
+        // lurking in other places that we want to unify in. The sibling stuff
+        // below does some of that - but only if we've looked high enough up the DOM
+        // tree.
+        var parentOfTopCandidate = topCandidate.parentNode;
+        // The scores shouldn't get too low.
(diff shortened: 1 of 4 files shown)


Readability.js
```diff
@@ -634,6 +634,29 @@ Readability.prototype = {
...
+        // The scores shouldn't get too low.
+        var scoreThreshold = topCandidate.readability.contentScore / 3;
+        var lastScore = parentOfTopCandidate.readability.contentScore;
+        while (parentOfTopCandidate && parentOfTopCandidate.readability) {
+          var parentScore = parentOfTopCandidate.readability.contentScore;
+          if (parentScore < scoreThreshold)
+            break;
+          if (parentScore > lastScore) {
+            // Alright! We found a better parent to use.
+            topCandidate = parentOfTopCandidate;
```
leibovic: If I'm understanding this correctly, this can only happen if this parent node wasn't part of the original set of candidates, right? However, it does seem like we push parentNode/grandParentNode into the candidates array in the scoring loop. Are you sure something just isn't going wrong in our logic to find the top candidates? Why do we even have two for loops in there?

leibovic: I was confused, I see now that we're looking for a top candidate that actually has a lower score than our current top candidate, but is a parent of the top candidate. This looks fine to me, as it means we will get more content.

mhammond:
> @mhammond Especially for you! ;-)

Awww, shucks :)

PR #51, opened 2015-03-21
Fix the title to not update when we already have one

gijsk:
This of course assumes that we get well-formed pages with a title element in the head, so that we ignore random SVG <title> things in the body. We could also add code for that, but that code was more complex and this seemed just as effective.


gijsk:
@fqueze Florian, can you have a look, please? :-)

PR #52, opened 2015-03-21
Use forEach when it makes sense.

n1k0:
I'm finding all these for loops hard to read and error prone since we use `var` instead of `let` statements.

I'm suggesting we loop using `forEach` whenever it makes sense to address both issues.

@leibovic @gijsk thoughts?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -915,7 +915,7 @@ Readability.prototype = {
   _hasChildBlockElement: function (element) {
     return this._someNode(element.childNodes, function(node) {
       return this.DIV_TO_P_ELEMS.indexOf(node.tagName) !== -1 ||
-             this._hasChildBlockElement(node)
+             this._hasChildBlockElement(node);
     });
   },
 
@@ -928,9 +928,10 @@ Readability.prototype = {
    * @return string
   **/
   _getInnerText: function(e, normalizeSpaces) {
+    normalizeSpaces = (typeof normalizeSpaces === 'undefined') ? true : normalizeSpaces;
     var textContent = e.textContent.trim();
 
-    if (!Boolean(normalizeSpaces)) {
+    if (normalizeSpaces) {
       return textContent.replace(this.REGEXPS.normalize, " ");
     } else {
       return textContent;
@@ -989,6 +990,9 @@ Readability.prototype = {
   **/
   _getLinkDensity: function(element) {
     var textLength = this._getInnerText(element).length;
+    if (textLength === 0)
+      return;
+
     var linkLength = 0;
 
     // XXX implement _reduceNodeList?
@@ -996,7 +1000,7 @@ Readability.prototype = {
       linkLength += this._getInnerText(linkNode).length;
     });
 
-    return textLength !== 0 ? linkLength / textLength : 0;
+    return linkLength / textLength;
   },
 
   /**
--- a/Readability.js
+++ b/Readability.js
@@ -119,6 +119,36 @@ Readability.prototype = {
   },
 
   /**
+   * Iterate over a NodeList, which doesn't natively fully implement the Array
+   * interface.
+   *
+   * For convenience, the current object context is applied to the provided
+   * iterate function.
+   *
+   * @param  NodeList nodeList The NodeList.
+   * @param  Function fn       The iterate function.
+   * @return void
+   */
+  _forEachNode: function(nodeList, fn) {
+    return Array.prototype.forEach.call(nodeList, fn, this);
+  },
+
+  /**
+   * Iterate over a NodeList, return true if any of the provided iterate
+   * function calls returns true, false otherwise.
+   *
+   * For convenience, the current object context is applied to the
+   * provided iterate function.
+   *
+   * @param  NodeList nodeList The NodeList.
+   * @param  Function fn       The iterate function.
+   * @return Boolean
+   */
+  _someNode: function(nodeList, fn) {
+    return Array.prototype.some.call(nodeList, fn, this);
+  },
+
+  /**
    * Converts each <a> and <img> uri in the given element to an absolute URI.
    *
    * @param Element
@@ -149,19 +179,18 @@ Readability.prototype = {
 
     function convertRelativeURIs(tagName, propName) {
       var elems = articleContent.getElementsByTagName(tagName);
-      for (var i = elems.length; --i >= 0;) {
-        var elem = elems[i];
+      this._forEachNode(elems, function(elem) {
         var relativeURI = elem.getAttribute(propName);
         if (relativeURI != null)
-          elems[i].setAttribute(propName, toAbsoluteURI(relativeURI));
-      }
+          elem.setAttribute(propName, toAbsoluteURI(relativeURI));
+      });
     }
 
      // Fix links.
-    convertRelativeURIs("a", "href");
+    convertRelativeURIs.call(this, "a", "href");
 
      // Fix images.
-    convertRelativeURIs("img", "src");
+    convertRelativeURIs.call(this, "img", "src");
   },
 
   /**
@@ -217,19 +246,17 @@ Readability.prototype = {
     var doc = this._doc;
 
     // Remove all style tags in head
-    var styleTags = doc.getElementsByTagName("style");
-    for (var st = styleTags.length - 1; st >= 0; st -= 1) {
-      styleTags[st].parentNode.removeChild(styleTags[st]);
-    }
+    this._forEachNode(doc.getElementsByTagName("style"), function(styleNode) {
+      styleNode.parentNode.removeChild(styleNode);
+    });
 
     if (doc.body) {
       this._replaceBrs(doc.body);
     }
 
-    var fonts = doc.getElementsByTagName("FONT");
-    for (var i = fonts.length; --i >=0;) {
-      this._setNodeTag(fonts[i], "SPAN");
-    }
+    this._forEachNode(doc.getElementsByTagName("font"), function(fontNode) {
+      this._setNodeTag(fontNode, "SPAN");
+    });
   },
 
   /**
@@ -255,9 +282,7 @@ Readability.prototype = {
    *   <div>foo<br>bar<p>abc</p></div>
    */
   _replaceBrs: function (elem) {
-    var brs = elem.getElementsByTagName("br");
-    for (var i = 0; i < brs.length; i++) {
-      var br = brs[i];
+    this._forEachNode(elem.getElementsByTagName("br"), function(br) {
       var next = br.nextSibling;
 
       // Whether 2 or more <br> elements have been found and replaced with a
@@ -296,7 +321,7 @@ Readability.prototype = {
           next = sibling;
         }
       }
-    }
+    });
   },
 
   _setNodeTag: function (node, tag) {
@@ -336,26 +361,21 @@ Readability.prototype = {
     this._cleanConditionally(articleContent, "div");
 
     // Remove extra paragraphs
-    var articleParagraphs = articleContent.getElementsByTagName('p');
-    for (var i = articleParagraphs.length - 1; i >= 0; i -= 1) {
-      var imgCount = articleParagraphs[i].getElementsByTagName('img').length;
-      var embedCount = articleParagraphs[i].getElementsByTagName('embed').length;
-      var objectCount = articleParagraphs[i].getElementsByTagName('object').length;
+    this._forEachNode(articleContent.getElementsByTagName('p'), function(paragraph) {
+      var imgCount = paragraph.getElementsByTagName('img').length;
+      var embedCount = paragraph.getElementsByTagName('embed').length;
+      var objectCount = paragraph.getElementsByTagName('object').length;
+      var totalCount = imgCount + embedCount + objectCount;
 
-      if (imgCount === 0 &&
-        embedCount === 0 &&
-        objectCount === 0 &&
-        this._getInnerText(articleParagraphs[i], false) === '')
-        articleParagraphs[i].parentNode.removeChild(articleParagraphs[i]);
-    }
+      if (totalCount === 0 && !this._getInnerText(paragraph, false))
+        paragraph.parentNode.removeChild(paragraph);
+    });
 
-    var brs = articleContent.getElementsByTagName("BR");
-    for (var i = brs.length; --i >= 0;) {
-      var br = brs[i];
+    this._forEachNode(articleContent.getElementsByTagName("br"), function(br) {
       var next = this._nextElement(br.nextSibling);
       if (next && next.tagName == "P")
         br.parentNode.removeChild(br);
-    }
+    });
   },
 
   /**
@@ -522,8 +542,7 @@ Readability.prototype = {
             elementsToScore.push(node);
           } else {
             // EXPERIMENTAL
-            for (var i = 0, il = node.childNodes.length; i < il; i += 1) {
-              var childNode = node.childNodes[i];
+            this._forEachNode(node.childNodes, function(childNode) {
               if (childNode.nodeType === Node.TEXT_NODE) {
                 var p = doc.createElement('p');
                 p.textContent = childNode.textContent;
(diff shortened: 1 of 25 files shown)


Readability.js
```diff
@@ -119,6 +119,36 @@ Readability.prototype = {
   },
 
   /**
+   * Iterate over a NodeList, which doesn't natively fully implement the Array
+   * interface.
```
DavidBruant: [NodeLists](https://dom.spec.whatwg.org/#interface-nodelist) are `iterable<node>`.
This means they can be iterated over with for..of loops.
It could result in the following code:
````js
function convertRelativeURIs(tagName, propName) {
    for(var elem of articleContent.getElementsByTagName(tagName)){
        var relativeURI = elem.getAttribute(propName);
        if (relativeURI != null)
            elem.setAttribute(propName, toAbsoluteURI(relativeURI));
    }
}
````js

This would work in Firefox, but wouldn't in Node 0.10. It can be made to work in Node 0.12 and io.js with some experimental flags on I believe. The future (hopefully coming soon) is that ES6 will be deployed consistently everywhere, so it will eventually work everywhere.

I don't know what's best, but felt it's worth sharing options :-)

Comment (n1k0) [inline Readability.js]: Readability's code needs to run on iOS, so I suspect what you suggest is not an option (sadly).

File: Readability.js [hunk 2]
```diff
@@ -1058,7 +1067,6 @@ Readability.prototype = {
   _findNextPageLink: function(elem) {
     var uri = this._uri;
     var possiblePages = {};
-    var allLinks = elem.getElementsByTagName('a');
```
gijsk: Err, this is used a few lines down... I mean, the code isn't currently being called at all, but we should try not to break it like this. :-)

n1k0: Fixed, I've been a little too prone at cutting things it seems :)

I've added more tests to cover things I've refactored.

Readability.js
gijsk: The downside of forEach used to be that the extra function calls produced overhead, so the performance is worse. I don't know if that's still (meaningfully) the case.

Readability.js
```diff
@@ -217,19 +246,17 @@ Readability.prototype = {
     var doc = this._doc;
 
     // Remove all style tags in head
-    var styleTags = doc.getElementsByTagName("style");
-    for (var st = styleTags.length - 1; st >= 0; st -= 1) {
-      styleTags[st].parentNode.removeChild(styleTags[st]);
-    }
+    this._forEachNode(doc.getElementsByTagName("style"), function(styleNode) {
+      styleNode.parentNode.removeChild(styleNode);
+    });
```
gijsk: Certainly here though, the issue I'd be worried about is that Array.forEach can't deal with removing items from the array while iterating. The reason this loop iterated backwards is to deal with NodeLists that are dynamic, ie when you call removeChild, the item will disappear from the doc.getElementsByTagName result, and therefore from the nodelist.

Your change does not handle that, so on compliant DOM implementations it would break this code. It doesn't with JSDOMParser, but I believe at least iOS will want to use Readability with a "real" DOM.

Readability.js
n1k0: I'm always prepending test suite run calls with the `time` exec, I didn't notice any performance impact. I think js engines are now optimizing this properly, and even if it doesn't perform as good I think more readable code is worth the price.

DavidBruant: Functions calls are removed by the JIT compiler, especially if the collection type is stable.
If you notice a significant difference (which is unlikely), file a bug, that's the sort of things JS engine folks want to know about.
More on the topic http://rfrn.org/~shu/2013/03/20/two-reasons-functional-style-is-slow-in-spidermonkey.html

gijsk: > Functions calls are removed by the JIT compiler, especially if the collection type is stable.

Which isn't the case here for real DOMs, right? :-)

Readability.js
n1k0:
```js
var a = [1, 2, 3];
a.forEach(function(x, i) {
  if (x === 2)
    a.splice(i, 1)
});
console.log(a) // [1,3]
```

I agree this isn't the most easy code to deal with, but in my world still better than within traditional for loops (I've probably been too much into FP latesly, trying to avoid mutable state changes as much as I can, or when I can't trying to limit side effects as much as possible :/)

n1k0: Hmm, you're probably right that dealing with a mutable DOM that way is probably not a good idea. I should just cancel this PR then.

Readability.js
n1k0: Guys, please note we're speaking milliseconds here. I think the argument of the impossibility to deal with DOM mutations within function based loops is much more worth spending our time.

Readability.js
gijsk: Hm, no, actually the docs say that I am just wrong:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

> The range of elements processed by forEach() is set before the first invocation of callback. Elements that are appended to the array after the call to forEach() begins will not be visited by callback. If the values of existing elements of the array are changed, the value passed to callback will be the value at the time forEach() visits them; elements that are deleted before being visited are not visited.

I don't think any of these loops care about added items?

n1k0: > I don't think any of these loops care about added items?

It doesn't seem so, though that means we won't ever be able to introduce such a behavior; on the other hand, appending items to a currently iterated collection is definitely asking for trouble imoe :D

test/test-pages/style-tags-removal/source.html
```diff
@@ -0,0 +1,42 @@
...
+      Tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+      quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+      consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
+      cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+      proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+    </div>
+  </article>
+  <style>
+    * {
+      color: yelow;
```
gijsk: Nit: yellow with two ls

test/test-pages/basic-tags-cleaning/source.html
```diff
@@ -0,0 +1,35 @@
...
+      cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+      proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+    </div>
+    <h2>Foo</h2>
+    <div>
+      <p>Tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
+      quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+      consequat.</p>
+      <object data="foo.swf" type="application/x-shockwave-flash" width="88" height="31">
+        <param movie="foo.swf" />
```
gijsk: Can you add a youtube example to this that is kept, please?

Readability.js
```diff
@@ -1571,7 +1579,7 @@ Readability.prototype = {
     if (!metadata.excerpt) {
       var paragraphs = articleContent.getElementsByTagName("p");
       if (paragraphs.length > 0) {
```
gijsk: Huh, can you file a followup issue to make this less dumb? If we only care about the first P, we should just do our own iteration.

Readability.js
```diff
@@ -978,16 +987,16 @@ Readability.prototype = {
...
-    for (var i = 0, il = links.length; i < il; i += 1) {
-      linkLength += this._getInnerText(links[i]).length;
-    }
+    // XXX implement _reduceNodeList?
+    this._forEachNode(element.getElementsByTagName("a"), function(linkNode) {
+      linkLength += this._getInnerText(linkNode).length;
+    });
 
-    return linkLength / textLength;
+    return textLength !== 0 ? linkLength / textLength : 0;
```
gijsk: Why was this necessary?

Readability.js
```diff
@@ -915,13 +924,13 @@ Readability.prototype = {
...
    * @param Element
+   * @param Boolean normalizeSpaces (default: true)
    * @return string
   **/
   _getInnerText: function(e, normalizeSpaces) {
     var textContent = e.textContent.trim();
-    normalizeSpaces = (typeof normalizeSpaces === 'undefined') ? true : normalizeSpaces;
 
-    if (normalizeSpaces) {
+    if (!Boolean(normalizeSpaces)) {
```
gijsk: Ewww. No need to convert.

Readability.js
```diff
@@ -900,14 +912,11 @@ Readability.prototype = {
...
-    for (var i = 0; i < length; i++) {
-      var child = e.children[i];
-      if (this.DIV_TO_P_ELEMS.indexOf(child.tagName) !== -1 || this._hasChildBlockElement(child))
-        return true;
-    }
-    return false;
+  _hasChildBlockElement: function (element) {
+    return this._someNode(element.childNodes, function(node) {
+      return this.DIV_TO_P_ELEMS.indexOf(node.tagName) !== -1 ||
+             this._hasChildBlockElement(node)
```
gijsk: Nit: semicolon.

Readability.js
```diff
@@ -860,14 +878,13 @@ Readability.prototype = {
    * @param Element
   **/
   _removeScripts: function(doc) {
-    var scripts = doc.getElementsByTagName('script');
-    for (var i = scripts.length - 1; i >= 0; i -= 1) {
-      scripts[i].nodeValue="";
-      scripts[i].removeAttribute('src');
+    this._forEachNode(doc.getElementsByTagName('script'), function(scriptNode) {
+      scriptNode.nodeValue = "";
```
gijsk: I wonder why we set `nodeValue` at all, and why we nullcheck the element's `parentNode`... Well, best leave that as-is for now, and file another followup, please? :-)

Readability.js
```diff
@@ -544,17 +563,17 @@ Readability.prototype = {
        * A score is determined by things like number of commas, class names, etc. Maybe eventually link density.
       **/
       var candidates = [];
-      for (var pt = 0; pt < elementsToScore.length; pt += 1) {
-        var parentNode = elementsToScore[pt].parentNode;
+      this._forEachNode(elementsToScore, function(elementToScore) {
```
gijsk: So, here's something which I've been wondering about... are you trying to be consistent for its own sake? Because you could just call `elementsToScore.forEach(...)`, right? Or update `_forEachNode` to check `Array.isArray` before doing the `.call` thing...

Readability.js
```diff
@@ -336,26 +361,21 @@ Readability.prototype = {
...
-        articleParagraphs[i].parentNode.removeChild(articleParagraphs[i]);
-    }
+    this._forEachNode(articleContent.getElementsByTagName('p'), function(paragraph) {
+      var imgCount = paragraph.getElementsByTagName('img').length;
+      var embedCount = paragraph.getElementsByTagName('embed').length;
+      var objectCount = paragraph.getElementsByTagName('object').length;
+      var totalCount = imgCount + embedCount + objectCount;
+
+      if (totalCount === 0 && !this._getInnerText(paragraph, false))
+        paragraph.parentNode.removeChild(paragraph);
```
gijsk: Hrmpf. All of this is pretty inefficient, but that can be another followup...

Readability.js
n1k0: Possible division by zero. Defensive programming, admittedly.

Readability.js
gijsk: Hrmpf. I thought this was all going to go OK, and then I realized we'll still need to fix the `_setNodeTag` stuff for non-JSDOMParser, which will involve reparenting stuff (ie, below, move everything from a `font` tag into a different `span` tag and then replace the `font` with the `span`). This is unfortunate because if there are nested tags, we will be adding + removing things, which will mess up the loop. For instance:

`<font><font>Hello</font></font>`

We'll hit the first font first with the forEach loop, which then means we'll reparent the second one, which *might* mean we won't iterate over the second node? Not sure... :-\

Tested in Firefox's devtools console:
```
>>> document.getElementsByTagName("font").forEach
undefined
>>> Array.prototype.forEach.call(document.getElementsByTagName("font"), function(f) { var x = document.createElement("span"); Array.prototype.forEach.call(f.childNodes, function(fc) { x.appendChild(fc); }); f.parentNode.replaceChild(x, f); });
undefined
>>> document.getElementsByTagName("font")
HTMLCollection [ <font> ]
>>> document.getElementsByTagName("font")[0].parentNode
<span>
```

Readability.js
gijsk: I see. In that case, can we just return early before calling getElementsByTagName if `textLength === 0`? If that is the case, clearly `linkLength` is also going to be 0 anyway...

Readability.js
n1k0: > Or update _forEachNode to check Array.isArray before doing the .call thing...

Would this be to address a possible performance penalty? Because applying the method from the prototype on an array works exactly the same way, and has nearly no impact on performances…

gijsk: No, I think I was just confused by the fact that we're using this helper function all the time, which says it's there to make things work with NodeList, but we use it for other things, too... :-)

I don't really mind either way, up to you!

Readability.js
n1k0: Heh. You can't believe the wtf-per-minute I've been producing while initially reading the code. We'll address all this iteratively, if you don't mind.

gijsk: Yeah, it was the same on my side. And yes, iteratively for sure. See also https://twitter.com/mleibovic/status/578676569582907392 for the state of some of the people here...

Readability.js
n1k0: I like the idea of avoiding calling expensive functions for no reason :) Will fix.

gijsk:
@n1k0 I need to head out and likely won't be back until the evening here in SF.

I think this is basically OK but we need to be careful with the forEach stuff in cases where we manipulate the things we're looping. Can you just file a followup that as well? I think the clarity this change adds is something we should take while we're still working on the compat with live DOMs.

gijsk:
(by which I mean, we can postpone the hard work of figuring out those changes later, and potentially change them back to manual loops if that fixes bugs at that point in time)

Readability.js
n1k0: Damn. Sounds so much complicated. Indirectly related, just out of curiosity: why do we want to replace font tags with spans? couldn't we just remove these font tags? Note that we currently keep the font tag attributes, which makes generated HTML looking a little silly.

gijsk: It's to avoid the styles of the readable page being messed up by the page's (font tag's) styles, I think. Like I said, we can leave it for a followup, it's not currently broken with this change.

test/test-pages/basic-tags-cleaning/source.html
n1k0: Good point, especially as youtube embedded code is now iframe-based only.

n1k0: Filed https://github.com/mozilla/readability/issues/56

Readability.js
n1k0: Filed https://github.com/mozilla/readability/issues/57

Readability.js
n1k0: I don't get why this code is a problem. Hints?

n1k0:
> Can you just file a followup that as well? 
https://github.com/mozilla/readability/issues/59

Readability.js
gijsk: It gets all the paragraphs (walks the entire tree), but then we only really want the first one. It could just stop looking after the first paragraph, ie the equivalent of querySelector. :-)

gijsk:
Great, this is ready to be merged now, right, @n1k0 ?

n1k0:
> Great, this is ready to be merged now, right, @n1k0 ?

I think so! :)

Readability.js
gijsk: Filed #60

n1k0:
\o/

Issue #58, opened 2015-03-22
<embed> tags aren't stripped out


n1k0:
```html
<div>
  …
  <embed src="foo.swf>
  …
</div>
```
That `embed` tag  won't be stripped out, while it probably should.

n1k0:
Fixed by cf3dce6cf21983c0b0a189d21e701ea34d261aab

PR #61, opened 2015-03-22
Refs #58 - Stripped embed tags.

n1k0:
(WiP)

Naive approach… that works.


gijsk:
Happy with this as a naive approach, although arguably we should refactor this into the main loop for efficiency at some point. Assuming this is final, please hit the merge button (I need to run out soon, may be back in 4-5 hours or so)

n1k0:
Yeah, efficiency is the big next topic for sure. For now let's have as many tests as we can so we get in a place where we can safely start to heavily refactor things.

PR #63, opened 2015-03-22
Preserve inline tweets as they're part of article contents.

n1k0:
I can see no good reason to filter out inline tweets, as they're most likely to contribute to article meaning.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -88,7 +88,7 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
-    unlikelyCandidates: /combx|comment|community|disqus|extra|foot|header|menu|remark|rss|shoutbox|sidebar|sponsor|ad-break|agegate|pagination|pager|popup|tweet|twitter/i,
+    unlikelyCandidates: /combx|comment|community|disqus|extra|foot|header|menu|remark|rss|shoutbox|sidebar|sponsor|ad-break|agegate|pagination|pager|popup/i,
     okMaybeItsACandidate: /and|article|body|column|main|shadow/i,
     positive: /article|body|content|entry|hentry|main|page|pagination|post|text|blog|story/i,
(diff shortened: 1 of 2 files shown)


gijsk:
If we want to do this we should probably have some kind of CSS for it too.

I'd like other people's feedback about whether we should be doing this though. @leibovic  ?

n1k0:
I'd be surprised we couldn't style a `blockquote` and some paragraphs within? Or are you suggesting we should have specific styling for tweets?

gijsk:
> I'd be surprised we couldn't style a blockquote and some paragraphs within? Or are you suggesting we should have specific styling for tweets?

We'd need to do /something/ . :-)

leibovic:
We do currently have styles for blockquotes in aboutReader.css. We could definitely add something for tweets, I think they're common enough that it would be fine (we actually already have some wordpress-specific styles in our mobile aboutReader.css, so there's precedent for site-specific styles).

n1k0:
I'm suggesting we shouldn't bother styling tweets specifically; twitter embedded code uses some generic enough semantics, eg.:

```html
<blockquote class="twitter-tweet tw-align-center">
    <p>Drinking game for web devs:
        <br>(1) Think of a noun
        <br>(2) Google "&lt;noun&gt;.js"
        <br>(3) If a library with that name exists - drink</p>— Shay Friedman (@ironshay)
    <a
    href="https://twitter.com/ironshay/statuses/370525864523743232">August 22, 2013</a>
</blockquote>
```

This works perfectly fine even in raw html. Marking it specifically as a *tweet* using CSS (eg. small twitter icon in the margin or smthg) would be *nice to have*, though this shouldn't be a requirement if you as me :)

On the contrary, stripping tweets entirely may completely change the meaning of an article. Well, your call.

n1k0:
If you're curious, have a look at how a page like https://storify.com/YoungInvincible/millennial-monday-special-enrollment renders without this patch.

gijsk:
Alright, so let's merge this and file a followup for tweet styles, I guess?

n1k0:
Sounds good to me :)

PR #65, opened 2015-03-23
Fixes #56 - Updated support for embedded Youtube & Vimeo videos.

### LINKED ISSUE #56: Most recent Youtube embedded markup isn't supported
Sample most recent embedded markup format:

```html
<iframe width="560" height="315" 
  src="https://www.youtube.com/embed/zOduYvDupXQ" 
  frameborder="0" 
  allowfullscreen></iframe>
```

As we're stripping out all iframes, this code is removed as well, resulting in end user missing embedded youtube videos in generated pages.

Note that Youtube offers a privacy-compliant version of the markup, which uses another domain to serve the iframe content:

```html
<iframe width="560" height="315" 
  src="https://www.youtube-nocookie.com/embed/zOduYvDupXQ" 
  frameborder="0" 
  allowfullscreen></iframe>
```

We should ensure that both versions are supported. I'd go as far as forcing the switch to the privacy-compliant version by default for generated readable content, though that's worth discussing.

n1k0:
Refs #56.

r=? @leibovic @gijsk

### DIFF
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -63,7 +63,7 @@ describe("Test page", function() {
         pathBase: "http://fakehost/test"
       };
 
-      beforeEach(function() {
+      before(function() {
         doc = new JSDOMParser().parse(source);
         result = new Readability(uri, doc).parse();
       });
--- a/Readability.js
+++ b/Readability.js
@@ -96,7 +96,7 @@ Readability.prototype = {
     byline: /byline|author|dateline|writtenby/i,
     replaceFonts: /<(\/?)font[^>]*>/gi,
     normalize: /\s{2,}/g,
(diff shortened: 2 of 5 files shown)


test/test-readability.js
```diff
@@ -63,7 +63,7 @@ describe("Test page", function() {
         pathBase: "http://fakehost/test"
       };
 
-      beforeEach(function() {
+      before(function() {
```
n1k0: Note on this: we were parsing before *each* test, while they all rely on the same result; that was unnecessary overhead, hence this change.

n1k0:
Also, we should probably add support for some other major online video providers, though we should address this in another bug.

PR #67, opened 2015-03-23
Fixes #66 - Keep tabular data as they're most likely to be part of content

### LINKED ISSUE #66: We shouldn't strip tables agressively, as they're most often actually part of content
Tabular data are valid content and shouldn't be stripped out entirely. If we're afraid of websites using tables as a layout strategy, once the main article content is properly identified that shouldn't be an issue.

Sample example: https://hsivonen.fi/webm-share/

n1k0:
The original Arc90's approach at dealing with table conditional cleaning seems really strong (see provided test case extracted from [this page](https://hsivonen.fi/webm-share/), where the table is entirely removed atm). 

The more I think about it, the less I can see concrete use cases where we'd actually want to remove tables entirely. Nowadays most sites use tables to structure tabular data, which can often contain a few amount of text though these are still really part of the article content. Finding a good algorithm to identify where it would make sense to conditionally strip tables seems highly fragile.

I suggest we just keep tables by default. That also saves us some CPU cycles, and we can always revisit this decision later if needed (bring use cases though!).

r=? @gijsk @leibovic

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -357,7 +357,6 @@ Readability.prototype = {
 
     // Do these last as the previous stuff may have removed junk
     // that will affect these
-    this._cleanConditionally(articleContent, "table");
     this._cleanConditionally(articleContent, "ul");
     this._cleanConditionally(articleContent, "div");
 
--- a/test/test-pages/keep-tabular-data/expected-metadata.json
+++ b/test/test-pages/keep-tabular-data/expected-metadata.json
@@ -1,0 +1,5 @@
+{
+  "title": "Keep tabular data test",
+  "byline": null,
+  "excerpt": "Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod\n      tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,\n      quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo\n      consequat."
+}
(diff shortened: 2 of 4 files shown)


gijsk:
I think a lot of "legacy" pages still use tables for layout (sigh). For instance, a quick look shows Yahoo's main page still does this. (And they seem to actually be generated by the YUI framework!) :-(

Accessibility code uses an algorithm to detect whether a table is "presentational" - ie not a semantic table. It checks for the `role="presentation"` attribute, which authors can use to explicitly mark a table as such, and it has some heuristics based on whether there are table headers and/or how many rows/columns there are. I can check the specifics with Marco Zehe and/or other Firefox a11y team members.

I also think that the number of "real" articles that do use data tables is probably comparatively small. Most of the testcases on the etherpad ( https://firefox-ux.etherpad.mozilla.org/reader-comparisons ) that I've looked at don't use data tables.

gijsk:
(in other words: based on all of those reasons, I think the risk of regressing big-but-stupid pages outweighs the benefit of fixing Henri's beautifully correct handcrafted HTML... at least for "let's just not do anything" option)

OTOH... maybe these tables are less likely to be inside the candidates we pick anyway? We should test that, ideally with a variety of pages that do have these silly table things.

gijsk:
Closing this in favour of #523 which landed recently and fixed this! :-)

PR #78, opened 2015-03-25
Fix readability to work with real dom (fixes #72)

### LINKED ISSUE #72: Avoid using and/or provide alternative implementation for _setNodeTag
While JSDOMParser is happy to just pretend a node just changed its tagName, you can't do that with a real DOM.

The main options are calling `appendChild` for all the node's kids, or setting `.innerHTML`.

gijsk:
This really needs careful review and probably another iteration or two.

There is 1 failing test, the TMZ one, with jsdom, because it contains a comment.

I don't want to add comment parsing to JSDOMParser.
I don't want to have different testcase output for jsdom.
I don't really want to strip all the comments explicitly because it'll need another DOM walk and is essentially useless.

Thoughts? @leibovic @n1k0 @DavidBruant

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -329,10 +329,25 @@ Readability.prototype = {
   },
 
   _setNodeTag: function (node, tag) {
-    // FIXME this doesn't work on anything but JSDOMParser (ie the node's tag
-    // won't actually be set).
-    node.localName = tag.toLowerCase();
-    node.tagName = tag.toUpperCase();
+    this.log("_setNodeTag", node, tag);
+    if (node.__JSDOMParser__) {
+      node.localName = tag.toLowerCase();
+      node.tagName = tag.toUpperCase();
+      return node;
+    }
+
+    var replacement = node.ownerDocument.createElement(tag);
+    while (node.firstChild) {
+      replacement.appendChild(node.firstChild);
+    }
+    node.parentNode.replaceChild(replacement, node);
+    if (node.readability)
+      replacement.readability = node.readability;
+
+    for (var i = 0; i < node.attributes.length; i++) {
+      replacement.setAttribute(node.attributes[i].name, node.attributes[i].value);
+    }
+    return replacement;
   },
 
   /**
@@ -462,6 +477,37 @@ Readability.prototype = {
     return node && node.nextElementSibling;
   },
 
+  /**
+   * Like _getNextNode, but for DOM implementations with no
+   * firstElementChild/nextElementSibling functionality...
+   */
+  _getNextNodeNoElementProperties: function(node, ignoreSelfAndKids) {
+    function nextSiblingEl(n) {
+      do {
+        n = n.nextSibling;
+      } while (n && n.nodeType !== n.ELEMENT_NODE);
+      return n;
+    }
+    // First check for kids if those aren't being ignored
+    if (!ignoreSelfAndKids && node.children[0]) {
+      return node.children[0];
+    }
+    // Then for siblings...
+    var next = nextSiblingEl(node);
+    if (next) {
+      return next;
+    }
+    // And finally, move up the parent chain *and* find a sibling
+    // (because this is depth-first traversal, we will have already
+    // seen the parent nodes themselves).
+    do {
+      node = node.parentNode;
+      if (node)
+        next = nextSiblingEl(node);
+    } while (node && !next);
+    return node && next;
+  },
+
   _checkByline: function(node, matchString) {
     if (this._articleByline) {
       return false;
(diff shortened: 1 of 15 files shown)


gijsk:
Huh, so the travis stuff rebases, which I've not done, and so there's another test failure because of the twitter addition that I just merged. Easy to fix, though...

Readability.js
```diff
@@ -329,10 +329,25 @@ Readability.prototype = {
...
-    // FIXME this doesn't work on anything but JSDOMParser (ie the node's tag
-    // won't actually be set).
-    node.localName = tag.toLowerCase();
-    node.tagName = tag.toUpperCase();
+    this.log("_setNodeTag", node, tag);
+    if (node.__JSDOMParser__) {
+      node.localName = tag.toLowerCase();
+      node.tagName = tag.toUpperCase();
+      return node;
+    }
```
n1k0: Hmm, I could totally see this going away in order to avoid starting dealing with implementation-specific behavior; though, if you think this has a very significant impact on performances we should probably keep it.

Readability.js
```diff
@@ -462,6 +477,37 @@ Readability.prototype = {
     return node && node.nextElementSibling;
   },
 
+  /**
+   * Like _getNextNode, but for DOM implementations with no
+   * firstElementChild/nextElementSibling functionality...
```
n1k0: Does this only concern jsdom? If so we might want to revisit using it at all if it's broken…

n1k0:
Looks good, though the `tmz-1` test is failing, because of some unstripped HTML comment it seems; shouldn't be too hard to fix.

While betting on jsdom as a representative real-world DOM impl. should probably be challenged at some point, we're not in a position to fiddle a lot and I think we should land this.

Thoughts and feedback from @leibovic & @DavidBruant would be appreciated :)

gijsk:
@n1k0 Thanks for reviewing! Yes, I think it'd be good if others would weigh in, too. How would you fix the comment issue? Strip it on the real DOM implementations? Implement in JSDOMParser so we don't kill it there? (otherwise, if we add it to the testcase, the jsdom-based test will succeed and the JSDOMParser-based one will fail...)

Readability.js
gijsk: I expect so, but I've not measured (which is generally true for this patch, and should probably happen before we decide to land it). I will be away from tomorrow until the second half of Monday, so if you have time to look at how this performs that would be helpful. I'm not sure what the best way of measuring this is now that both tests run in the same test framework...

n1k0: I'll try to come with a simple [matcha](https://github.com/logicalparadox/matcha) banchmark to measure this.

n1k0:
I wish we had more serialization options here… Adding support for them to JSDOMParser seems unavoidable, though that'd be only to make tests pass :(

I'll look into this tomorrow.

test/test-pages/embedded-videos/expected.html
```diff
@@ -10,20 +10,21 @@
...
         <p>At root</p>
         <iframe width="560" height="315" src="https://www.youtube.com/embed/LtOGa5M8AuU"
-        frameborder="0"></iframe>
+        frameborder="0" allowfullscreen=""></iframe>
         <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/LtOGa5M8AuU"
-        frameborder="0"></iframe>
-        <iframe src="https://player.vimeo.com/video/32246206?color=ffffff&title=0&byline=0&portrait=0"
-        width="500" height="281" frameborder="0"></iframe>
+        frameborder="0" allowfullscreen=""></iframe>
+        <iframe src="https://player.vimeo.com/video/32246206?color=ffffff&amp;title=0&amp;byline=0&amp;portrait=0"
```
DavidBruant: are the `&amp;` normal? I'd write the URL with and `&` directly (but maybe I'd be wrong)

DavidBruant:
Beyond the nits and questions, it's hard for me to give a review as I'm not super-familiar with the code. In any case, I believe this change is important.

test/test-pages/embedded-videos/expected.html
gijsk: IIRC they are not strictly necessary in HTML unless there is a `;` before the next non-alphanumeric-non-# character, but I could be wrong. Having them, however, is not incorrect.

Generally, the encoding/decoding of attributes that this patch does could do with another look, too.

Readability.js
gijsk: Not sure if xmldom will work with a pure HTML-based DOM... and yes, the fix is only for jsdom; Firefox will cope (as will webkit, I believe).

gijsk:
Closing this in favour of pull request #80 . Let's sync up about that tomorrow.

PR #80, opened 2015-03-27
Fix readability to work with real dom (fixes #72)

### LINKED ISSUE #72: Avoid using and/or provide alternative implementation for _setNodeTag

n1k0:
This patch replaces #78. It removes HTML comments from JSDOM generated document.

r=? @gijsk @leibovic


gijsk:
I'd take the comment fix; it's really just for the benefit of the testcase.

I'm still worried about perf here, though...

gijsk:
Summarizing issues from #78:

* perf
* HTML encoding issues
* ???

test/test-readability.js
```diff
@@ -63,29 +97,19 @@ describe("Test page", function() {
...
 
-      before(function() {
-        doc = new JSDOMParser().parse(source);
-        result = new Readability(uri, doc).parse();
-      });
-
-      it("should return a result object", function() {
-        expect(result).to.include.keys("content", "title", "excerpt", "byline");
-      });
+      describe("JSDOM", function() {
```
DavidBruant: Nit: `jsdom`? (lowercase)

test/test-readability.js
```diff
@@ -47,11 +48,44 @@ var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
   };
 });
 
+function suite(result, expectedContent, expectedMetadata) {
+  it("should return a result object", function() {
+    expect(result).to.include.keys("content", "title", "excerpt", "byline");
+  });
+
+  it("should extract expected content with jsdom", function() {
```
DavidBruant: Nit: This prints out `should extract expected content with jsdom` whether jsdom or JSDOMParser is used.

DavidBruant:
Regarding perf, I don't understand the problem.
From what I understand, at the branch in this PR, there are 22 test pages. Each page has its DOM built via both jsdom and JSDOMParser and the each of these 2 Documents goes through the `Readability().parse()`, so it's 44 DOM building + readability (+22 removing comments). On my machine, running all that takes under 700ms. Granted, it's in command line, not in the browser fighting against other threads and addons for CPU, but I find that extremely fast.
What is the performance problem? Is there a page or a use case in particular that has performance issues?

test/test-readability.js
```diff
@@ -47,11 +48,44 @@ var testPages = fs.readdirSync(testPageRoot).map(function(dir) {
...
+
+  it("should extract expected excerpt", function() {
+    expect(expectedMetadata.excerpt).eql(result.excerpt);
+  });
+}
+
+function removeCommentNodesRecursively(node) {
+  for (var n = 0, length = node.childNodes.length; n < length; n++) {
+    var child = node.childNodes[n];
+    if (!child) continue; // JSDOM is dumb sometimes :/
```
DavidBruant: Can you elaborate? :-) I'm sure they'll want to know.
I can reduce a test case if I'm told where to begin.

n1k0: > Can you elaborate? :-)
For some reason, jsdom retrieves a bunch of `undefined` items in `childNodes` list for no obvious reason.
> I'm sure they'll want to know.

They don't. The 3.x branch of jsdom is claimed [not officially maintained anymore since they moved to the io.js platform for 4.x](https://github.com/tmpvar/jsdom/tree/3.x#jsdom-3x-branch).

But you're invited to fork and maintain your own version if you will (hint: don't).

DavidBruant: > (please note I'm not fascinated with the idea of investigating this further atm, but maybe you are?).

I am. And it was stupid of me asking to be told where to begin. I should just comment your line, notice the bug in the tests and start from there.
I am interested, because jsdom is the closest attempt I'm aware of of a conformant standalone DOM in Node.js. 

> But you're invited to fork and maintain your own version if you will (hint: don't).

:-p I'll try to repro the bug you found, find out whether it's still in 4.x and report to them if so.

DavidBruant: ... wait... isn't it just a bug due to removing elements in a for-loop?
You cache the length, start the loop, remove elements. If one element got removed, at the end of the loop, `node.childNodes[n]` can only return `undefined`, because the `node.childNodes` isn't of the same length that it was at the beginning of the loop any longer. Possibly, you're also forgetting to remove some comments, by skipping one element after one was removed.

You can clone the list via Array.prototype.slice.call (or Array.from) and iterate over that (with forEach, Yay \o/)

DavidBruant: I'm wrong about forgetting some elements because of the `n--` below, but I think the length explanation still stands.

n1k0: Fixed using a forEach loop.

gijsk:
> What is the performance problem?

I want at least some idea of how the JSDOMParser/Readability.js changes here impact performance, in relative terms instead of "on our really fast machines, when run in isolation on node.js, it takes 20ms so there's no problem". :-)

gijsk:
In particular, I suspect the HTML encoding I added in JSDOMParser.js could be very slow.

gijsk:
Now that we have benchmarks, it's pretty easy to see that this indeed slows down JSDOMParser, though it seems to speed up Readability a little, which is somewhat surprising. I'm going to look at using some more laziness in the textContent/innerHTML stuff, as well as ensuring entities like `&#xa8;` and `&#2552;` all work.

gijsk:
So, this version makes JSDOMParser faster than it currently is, and Readability slightly slower. All in all, I'm pretty happy with the code here at this point. @n1k0, @leibovic can one of you review? :-)

Readability.js
```diff
@@ -329,10 +329,25 @@ Readability.prototype = {
   },
 
   _setNodeTag: function (node, tag) {
-    // FIXME this doesn't work on anything but JSDOMParser (ie the node's tag
-    // won't actually be set).
-    node.localName = tag.toLowerCase();
-    node.tagName = tag.toUpperCase();
+    this.log("_setNodeTag", node, tag);
+    if (node.__JSDOMParser__) {
```
n1k0: Now we have benchmarks, is performance really bad if we remove this and use the generic behavior? I still think we shouldn't distinguish between parsers within Readability…

n1k0: Hmm okay got it, JSDOMParser doesn't support what's required below.

test/test-jsdomparser.js
```diff
@@ -190,6 +195,19 @@ describe("Test JSDOM functionality", function() {
...
   });
+
+  it("should handle encoding HTML correctly", function() {
+    var baseStr = "<p>Hello, everyone &amp; all their friends, &lt;this&gt; is a &quot; test.</p>";
+    var doc = new JSDOMParser().parse(baseStr);
+    var p = doc.getElementsByTagName("p")[0];
+    expect("<p>" + p.innerHTML + "</p>").eql(baseStr);
+
+    p.textContent = p.textContent + " ";
+    expect("<p>" + p.innerHTML + "</p>").eql(baseStr.replace("</p>", " </p>"));
```
n1k0: What's the exact purpose of this assertion?

test/test-jsdomparser.js
```diff
@@ -190,6 +195,19 @@ describe("Test JSDOM functionality", function() {
...
+
+    var doc = new JSDOMParser().parse("<p>&#32;&#x20;</p>");
+    expect(doc.getElementsByTagName("p")[0].textContent).eql("  ");
```
n1k0: I suggest we generally split tests containing many assertions into smaller ones, so when the suite fails we have a more precise idea what is exactly failing.

JSDOMParser.js
```diff
@@ -472,7 +524,27 @@
 
     nodeName: "#text",
     nodeType: Node.TEXT_NODE,
-    textContent: ""
+    get textContent() {
+      if (typeof this._textContent === "undefined") {
+        this._textContent = decodeHTML(this._innerHTML || "");
```
n1k0: Nit: `this._textContent = decodeHTML(this._innerHTML)  || "";` might be some ns faster

JSDOMParser.js
```diff
@@ -452,6 +487,23 @@
     this.value = value;
   };
 
+  Attribute.prototype = {
+    setValue: function(newValue) {
```
n1k0: This is not used and can probably be safely removed.

JSDOMParser.js
```diff
@@ -575,13 +647,14 @@
...
               var attr = child.attributes[j];
+              // attr.value is HTML escaped.
               var quote = (attr.value.indexOf('"') === -1 ? '"' : "'");
               arr.push(" " + attr.name + '=' + quote + attr.value + quote);
             }
 
             if (child.localName in voidElems) {
               // if this is a self-closing element, end it here
-              arr.push("/>");
+              arr.push(">");
```
n1k0: Out of curiosity, why this specific change?

gijsk: Because `jsdom` doesn't use self-closing tag thingies, and so the expected output was different. I standardized on the jsdom behaviour and updated JSDOMParser to match it.

JSDOMParser.js
gijsk: I don't want to do that, because before this, consumers assigned directly into `attr.value`. That shouldn't happen because then `this._decodedValue` isn't deleted and will have a stale value.

gijsk: Maybe I should prefix `value` to be `_value` so as to make this more clear...

n1k0:
Looks good to me. Some comments you might want to address before landing.

test/test-jsdomparser.js
gijsk: So JSDOMParser actually stores `innerHTML` straight from the parser and then regurgitates it for the first assertion.

Assigning to `textContent` will break that cache and force it to re-encode from the escaped version.

JSDOMParser.js
n1k0: Ok makes sense.

JSDOMParser.js
n1k0: Ah yes, good idea, this is far from obvious just reading through the code :)

test/test-jsdomparser.js
n1k0: Oooooh ok. Maybe that's worth a comment.

Readability.js
gijsk: Which part does it not support? I think it should work in theory, but I'd be very surprised if it wasn't slower. That said, the benchmarks are super noisy on my machine. I get results anywhere between 170 and 250 ops/s for the JSDOMParser part on the reference benchmark, with an average of about 200, all on the same changeset.

gijsk: Trying to fix that by increasing the iterations massively increases the number of ops/s, which makes me think we're hitting some kind of JIT optimization, which isn't really realistic (we won't normally readerize the same document a bunch of times in a row).

n1k0: JSDOMParser doesn't seem to support `ownerDocument`, though that's probably no bof deal adding support for it.

n1k0: > we won't normally readerize the same document a bunch of times in a row

True :)

PR #87, opened 2015-03-31
Fix script parsing to ignore closing tags in comments

### LINKED ISSUE #77: JSDOMParser errors on http://www.newyorker.com/magazine/2014/12/01/quiet-german
```
JSDOMParser error: expected '</lineargradient>'
JSDOMParser error: expected '</defs>'
JSDOMParser error: expected '</svg>'
JSDOMParser error: expected '</h1>'
JSDOMParser error: expected '</div>'
JSDOMParser error: expected '</div>'
JSDOMParser error: expected '</div>'
JSDOMParser error: expected '</div>'
JSDOMParser error: expected '</div>'
JSDOMParser error: expected '</body>'
JSDOMParser error: expected '</html>'
```

gijsk:
r? @n1k0 and/or @leibovic

This should fix #77, but I've not tested that, just the reduced testcase in this commit.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -919,14 +919,59 @@
     },
 
     readScript: function (node) {
-      var index = this.html.indexOf("</script>", this.currentChar);
-      if (index === -1) {
-        index = this.html.length;
+      while (this.currentChar < this.html.length) {
+        var c = this.nextChar();
+        var nextC = this.peekNext();
+        if (c === "<") {
+          if (nextC === "!" || nextC === "?") {
+            // We're still before the ! or ? that is starting this comment:
+            this.currentChar++;
+            node.appendChild(this.discardNextComment());
+            continue;
+          }
+          if (nextC === "/" && this.html.substr(this.currentChar, 8 /*"/script>".length */).toLowerCase() == "/script>") {
+            // Go back before the '<' so we find the end tag.
+            this.currentChar--;
+            // Done with this script tag, the caller will close:
+            return;
+          }
+        }
+        // Either c wasn't a '<' or it was but we couldn't find either a comment
+        // or a closing script tag, so we should just parse as text until the next one
+        // comes along:
+
+        var haveTextNode = node.lastChild && node.lastChild.nodeType === Node.TEXT_NODE;
+        var textNode = haveTextNode ? node.lastChild : new Text();
+        var n = this.html.indexOf("<", this.currentChar);
+        // Decrement this to include the current character *afterwards* so we don't get stuck
+        // looking for the same < all the time.
+        this.currentChar--;
+        if (n === -1) {
+          textNode.textContent += this.html.substring(this.currentChar, this.html.length);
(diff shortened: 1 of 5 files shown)


JSDOMParser.js
```diff
@@ -919,14 +919,59 @@
...
 
     readScript: function (node) {
-      var index = this.html.indexOf("</script>", this.currentChar);
-      if (index === -1) {
-        index = this.html.length;
+      while (this.currentChar < this.html.length) {
+        var c = this.nextChar();
+        var nextC = this.peekNext();
+        if (c === "<") {
+          if (nextC === "!" || nextC === "?") {
```
n1k0: I may be missing something obvious, but what type of comments `?` does match? :)

n1k0: Bah, just learned about the `<?…?>` html comment type. Never too late I guess.

n1k0:
Note: I'd expect tests to be added to the JSDOMParser suite instead of only relying on Readability ones.

gijsk:
> Note: I'd expect tests to be added to the JSDOMParser suite instead of only relying on Readability ones.

Good point, let me write some and get back here.

JSDOMParser.js
```diff
@@ -919,14 +919,59 @@
...
-      var txt = new Text();
-      txt.textContent = this.html.substring(this.currentChar, index === -1 ? this.html.length : index);
-      node.appendChild(txt);
-      this.currentChar = index;
+    },
+
+    discardNextComment: function() {
+      if (this.match("--")) {
+        this.discardTo("-->");
+      } else {
```
n1k0: Shouldn't we discardTo(`?>`) as well if `?` is found? This test case fails:

```js
  it("should strip comments within script tags", function() {
    var html = '<script><?Silly test > <script src="foo.js"></script>?></script>';
    var doc = new JSDOMParser().parse(html);
    expect(doc.firstChild.tagName).eql("SCRIPT");
    expect(doc.firstChild.textContent).eql("");
  });
```

gijsk: Probably! This is copied from the existing code, actually... so I guess that was also broken before. Writing tests atm, that'll help here.

gijsk: Err, actually the parser spec says: http://www.w3.org/TR/html5/syntax.html#bogus-comment-state

> Consume every character up to and including the first ">" (U+003E) character or the end of the file (EOF), whichever comes first. Emit a comment token whose data is the concatenation of all the characters starting from and including the character that caused the state machine to switch into the bogus comment state, up to and including the character immediately before the last consumed character (i.e. up to the character just before the U+003E or EOF character)

And this is triggered here: http://www.w3.org/TR/html5/syntax.html#tag-open-state

gijsk: In other words, AFAICT the current behaviour is correct...

n1k0:
LGTM especially with the tests added and a green build :)

gijsk:
r+ from @n1k0 over IRC, so merging this in. :)

PR #88, opened 2015-04-01
Create a quick way to measure performance

gijsk:
This is a pretty bare-bones approach, but it spits out two numbers and that seems useful enough as a baseline. @leibovic, @n1k0 - thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "Readability.js",
   "scripts": {
     "test": "mocha test/test-*.js",
+    "perf": "node test/time-readability.js",
     "generate-testcase": "node test/generate-testcase.js"
   },
   "repository": {
--- a/test/time-readability.js
+++ b/test/time-readability.js
@@ -1,0 +1,72 @@
+var path = require("path");
+var fs = require("fs");
+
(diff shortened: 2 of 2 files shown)


DavidBruant:
2 numbers aren't enough to be meaningful (on my machine, `Readerized everything, took 286ms.`, yes, but for 1 file ? 10 ? 100 ?). At least print out the number of processes HTML files.
With [fast-stats](https://www.npmjs.com/package/fast-stats), you can very easily get the mean/median/sum of a serie of values.

For high-precision time measurement (not Date.now() specialty), you can use [process.hrtime()](https://nodejs.org/api/process.html#process_process_hrtime)

gijsk:
The noise is in the 1-2ms ballpark on my machine, so it doesn't seem useful to get more high-precision measures.

My usecase is "how much does patch X slow this down", for which the current state is fine.

I selected the "real" articles from our existing testcases, so I don't think mean/median is useful as they are different files.

gijsk:
The `process.hrtime()` doesn't seem to improve noise levels much for me, but perhaps on slower machines it'll help, so let's go with that.

gijsk:
Closing in favour of #90 .

PR #90, opened 2015-04-01
Added benchmarks.

n1k0:
r=? @gijsk @leibovic 

screen shot 2015-04-01 at 21 58 48

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,31 @@
+var path = require("path");
+var fs = require("fs");
+var prettyPrint = require("html").prettyPrint;
+var chai = require("chai");
+chai.config.includeStack = true;
+var expect = chai.expect;
+
+// We want to load Readability and JSDOMParser, which aren't set up as commonjs libraries,
+// and so we need to do some hocus-pocus with 'vm' to import them on a separate scope
+// (identical) scope context.
+var vm = require("vm");
+var readabilityPath = path.join(__dirname, "Readability.js");
+var jsdomPath = path.join(__dirname, "JSDOMParser.js");
+
+
+var scopeContext = {};
+// We generally expect dump() and console.{whatever} to work, so make these available
+// in the scope we're using:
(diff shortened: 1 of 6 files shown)


gijsk:
Can we get an average of a fixed set of those ops/s so it's easier to compare the output ?

gijsk:
(the total time for all benchmarks is less useful because new commits tend to add a new testcase)

n1k0:
> Can we get an average of a fixed set of those ops/s so it's easier to compare the output ?

I couldn't find an option for this, except exporting the result as CSV (brrr). We could create a separate suite with a fixed subset of these test pages, and mark it as a constant reference point, though.

gijsk:
>> Can we get an average of a fixed set of those ops/s so it's easier to compare the output ?

> I couldn't find an option for this, except exporting the result as CSV (brrr). We could create a separate suite with a fixed subset of these test pages, and mark it as a constant reference point, though.

That sounds like a good idea, yeah.

n1k0:
screen shot 2015-04-01 at 22 27 37

n1k0:
@gijsk now we have these two distinct commands, `npm run perf` and `npm run perf-reference`, but we need to select the [right subset of reference test pages](https://github.com/mozilla/readability/blob/33eb24f4070370d9778b53ecb337580d435586f6/benchmarks.js#L7-L12). Could you please take over this, then land the path when you're happy? It's a little late here :)

index.js
```diff
@@ -0,0 +1,31 @@
+var path = require("path");
+var fs = require("fs");
+var prettyPrint = require("html").prettyPrint;
+var chai = require("chai");
+chai.config.includeStack = true;
+var expect = chai.expect;
```
n1k0: Ah just spotted all this should be removed from this file, as we're not in a test context.

gijsk: Done.

PR #91, opened 2015-04-02
Add benchmark for readability and show averages by using custom reporter

gijsk:
@n1k0 noticed two issues:

1. We were only testing JSDOMParser instead of testing Readability
2. It was hard to interpret the results because the test time is meant to stay the same (that's what iterations are for) so we still had to inspect the number of operations/second for each file.

I added a custom reporter that fixes (2) and added a second benchmark suite that fixes (1). Does this all make sense?

PS: should we do a separate PR to move this all into a subdir like "benchmarks" or "lib" or whatever? Having this stuff in the toplevel seems ugly...

### DIFF
--- a/benchmark-reporter.js
+++ b/benchmark-reporter.js
@@ -1,0 +1,41 @@
+var matchaPath = require.resolve('matcha');
+var path = require('path');
+
+var clean = require('matcha/lib/matcha/reporters/clean'));
+
+function average(list) {
+  if (!list.length)
+    return 0;
+
+  var sum = list.reduce(function(previous, current) { return previous + current; });
+  return (sum / list.length).toFixed(0);
+}
+
+
+// Like clean, but also produces an average:
+module.exports = function(runner, utils) {
+  var humanize = utils.humanize;
+  var padBefore = utils.padBefore;
+  var color = utils.color;
+  var results = {};
+  var currentResults = [];
+  runner.on('bench end', function(results) {
+    currentResults.push(results.ops);
+  });
+  runner.on('suite end', function(suite) {
+    var avg = humanize(average(currentResults));
+    console.log(padBefore(avg + ' op/s', 22) + ' » ' + suite.title);
+    console.log();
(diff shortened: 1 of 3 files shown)


benchmark-reporter.js
```diff
@@ -0,0 +1,37 @@
+
+var matchaPath = require.resolve('matcha');
+var path = require('path');
+
+var clean = require(path.join(path.dirname(matchaPath), 'lib', 'matcha', 'reporters', 'clean'));
+
+function average(list) {
+  var sum = list.reduce(function(previous, current) { return previous + current; });
+  return (sum / list.length).toFixed(0);
```
pdehaan: could `list.length` ever be 0?

gijsk: For hypothetical other consumers, yes. For us, not unless we remove all our testcases...

benchmark-reporter.js
```diff
@@ -0,0 +1,37 @@
+
+var matchaPath = require.resolve('matcha');
+var path = require('path');
+
+var clean = require(path.join(path.dirname(matchaPath), 'lib', 'matcha', 'reporters', 'clean'));
```
n1k0: You can simply use

```js
var clean = require('matcha/lib/matcha/reporters/clean');
```

benchmark-reporter.js
```diff
@@ -0,0 +1,37 @@
...
+
+function average(list) {
+  var sum = list.reduce(function(previous, current) { return previous + current; });
+  return (sum / list.length).toFixed(0);
+}
```
n1k0: Nit: \n

benchmark-reporter.js
n1k0: We should return early if the list is empty

n1k0:
All this makes a lot of sense, r+ with comments addressed. 

We should definitely move everything benchmarking into its own directory, though that could be either done here or in a next PR: your call :)

Issue #85, opened 2015-03-29
Add the "simple check" script to the repo


DavidBruant:
From #77 
> Note that on Android we ran this parse on every page load (to see if it was worth/possible offering reader mode), and we couldn't manipulate the real DOM because that'd obviously change the page, so there was a need to serialize+reparse anyway.

> We don't do that anymore on desktop because doing that caused a boatload of perf regressions. So we have a simpler check that determines whether we show the button, which isn't 1:1 equivalent to whether the content is reader-able, but it's close. From that perspective, in principle we could reuse the DOM.

If Mozilla came to the conclusion that Readability is only practical if a simple check is used beforehand (instead of running readability on every page), I believe it might be worth adding this simpler check to this repo.
This will help other reusers too (including the iOS browser as I understand it).

leibovic:
I think this is a good point (/cc @st3fan @thebnich). We don't do this in Firefox for Android yet, but we will probably do this soon to match desktop (https://bugzilla.mozilla.org/show_bug.cgi?id=1046112#c10).

Right now this code lives in ReaderMode.jsm, which is a module shared between desktop/Android Firefox:
http://mxr.mozilla.org/mozilla-central/source/toolkit/components/reader/ReaderMode.jsm?force=1#70

There's also a bug filed about trying to make this method better, since it's currently a bit aggressive about saying pages aren't reader-able:
https://bugzilla.mozilla.org/show_bug.cgi?id=1147626

leibovic:
Yes, definitely, I want some tests for this. I could make a PR for this.

@n1k0, do you have ideas about where we should put this? Should it just go in its own file? The annoying bit about that would be trying to keep that in sync with our version in mozilla-central.

Alternately, we could make another method on the Readability object, but then we would need to update ReaderMode.jsm to import Readability.js. But I assume that's not a big deal.

n1k0:
> Alternately, we could make another method on the Readability object, but then we would need to update ReaderMode.jsm to import Readability.js. But I assume that's not a big deal.

If it's not a big deal, I think that's the way to go :)

n1k0:
Fixed by #96

PR #95, opened 2015-04-02
Bug 1150695 - Move isProbablyReaderable function to Readability.js

leibovic:
f? @n1k0 

This should address https://github.com/mozilla/readability/issues/85

I added a simple test for this, but we fail to detect that some of our test pages are "probably readerable", so we'll either need to update `isProbablyReaderable` to be more lenient (long term solution) or find a way to hide these test failures (short term solution).

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1561,6 +1561,39 @@ Readability.prototype = {
   },
 
   /**
+   * Decides whether or not the document is reader-able without parsing the whole thing.
+   *
+   * @return boolean Whether or not we suspect parse() will suceeed at returning an article object.
+   */
+  isProbablyReaderable: function() {
+    var nodes = this._doc.getElementsByTagName("p");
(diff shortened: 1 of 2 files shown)


n1k0:
I suggest we first add tests matching the current behavior (maybe by adding a new `readerable` prop to `expected-metadata.json`?), then we should fix the its behavior in a next PR.

gijsk:
> I suggest we first add tests matching the current behavior (maybe by adding a new `readerable` prop to `expected-metadata.json`?), then we should fix the its behavior in a next PR.

I agree, though I will also point out that the only currently failing things are the unit-test-y tests that we created ourselves. All the actual articles (from varied sources) are passing.

Finally, when pursuing the metadata solution, we should make sure that we update the testcase-generation script so it sets that value, too. :-)

PR #100, opened 2015-04-05
Fixes #99 - Fixed JSDOMParser tag name case handling.

### LINKED ISSUE #99: Readabily fails at parsing pages using svg camel-cased tags
Sample test case:

```html
<svg version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 50 50" height="50" width="50" style="position: absolute;">
  <g>
    <clipPath id="hex-mask-large"><polygon points="15,35 10,35 10,0 10,0 45,0 45,35 45,35 25,35 15,43"/></clipPath>
    <clipPath id="hex-mask-small"><polygon points="5,1 5,16 3,23 10,20 24,20 24,1"/></clipPath>
  </g>
</svg>
```

Gives:

```
JSDOMParser error: expected '</clippath>'
JSDOMParser error: expected '</svg>'
JSDOMParser error: expected '</body>'
JSDOMParser error: expected '</html>'
```

Replacing `clipPath` with `clippath` "fixes" the issue. There's definitely an issue with force-lowercasing tag names here.

n1k0:
Okay so this patch fixes #99 from the JSDOMParser side of things, but unfortunately it appears that jsdom related tests are now failing. This is because jsdom serializer forces lowercasing tag names:

```
$ node
> jsdom = require("jsdom").jsdom
> serialize = require("jsdom").serializeDocument
> serialize(jsdom("<html><body><svg><clipPath/></svg></body></html>"))
'<html><head></head><body><svg><clippath></clippath></svg></body></html>'
```

I suspect jsdom shouldn't alter tag name case that way. Dunno what to do, thoughts?

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -962,7 +962,7 @@
      */
     match: function (str) {
       var strlen = str.length;
-      if (this.html.substr(this.currentChar, strlen) === str) {
+      if (this.html.substr(this.currentChar, strlen).toLowerCase() === str.toLowerCase()) {
         this.currentChar += strlen;
         return true;
       }
@@ -1058,7 +1058,7 @@
      */
     readNode: function () {
       var c = this.nextChar();
- 
+
       if (c === undefined)
         return null;
 
--- a/test/test-jsdomparser.js
+++ b/test/test-jsdomparser.js
@@ -267,3 +267,16 @@ describe("Script parsing", function() {
     expect(doc.firstChild.childNodes.length).eql(1);
   });
 });
+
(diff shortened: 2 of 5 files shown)


gijsk:
Seems like we could just fix this by teaching JSDOMParser to expect a close tag name with any casing? That might be simpler (and more robust) than a whitelist.

Less sure what to do with jsdom...

n1k0:
@gijsk r=?

gijsk:
r=me with the commits squashed before landing.

Issue #117, opened 2015-04-07
Reader View displays strings related to social widgets for specific wordpress blogs


avaida:
**Reproducible on**:
Beta 38.0b2-build1 (20150406174117)

**Sample page showing this issue**:
http://blog.longreads.com/2015/03/24/i-would-prefer-not-to-the-origins-of-the-white-collar-worker/

**Screenshot**:
![screenshot from 2015-04-07 17 16 32](https://cloud.githubusercontent.com/assets/5707237/7025108/f7648304-dd49-11e4-8a3c-2455545d8810.png)

gijsk:
This is fixed on Nightly but I don't see this fixed on beta and I don't understand why...

gijsk:
... oh, I do! Uplifts will be in yesterday's beta which I don't seem to be getting yet, for some reason. :-\

PR #120, opened 2015-04-07
Refs #117 - Drop social/share buttons.

n1k0:
This is probably a little too aggressive :) though in my personal tests it's really efficient at removing social buttons/links/widgets.

Thoughts? @gijsk @leibovic


gijsk:
This looks OK to me, and we have enough time to revert if this turns out to be problematic.

PR #121, opened 2015-04-08
Fixed test result output was sent at once.

n1k0:
In b1f6c62b2512e3350f46bfd5ddb3761ad689667f we refactored the tests for readability so they share a suite between jsdom and JSDOMParser; though this introduced a strange behavior for the test runner, which is displaying all the test output at once since, instead of realtime as previously.

This patch addresses this issue.

r=? @leibovic @gijsk

### DIFF
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -10,32 +10,6 @@ var JSDOMParser = readability.JSDOMParser;
 
 var testPages = require("./bootstrap").getTestPages();
 
-function suite(result, expectedContent, expectedMetadata) {
-  it("should return a result object", function() {
-    expect(result).to.include.keys("content", "title", "excerpt", "byline");
-  });
-
-  it("should extract expected content", function() {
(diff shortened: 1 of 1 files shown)


gijsk:
Why does this work? I mean, I believe it does, I just don't understand it, and I don't like the fact that now we have a lot of duplicate code in there...

n1k0:
I don't know, and after spending a few minutes digging into the mocha codebase to figure out why, I decided it wasn't worth my time. Though I can revisit this decision if you insist.

gijsk:
> I don't know, and after spending a few minutes digging into the mocha codebase to figure out why, I decided it wasn't worth my time. Though I can revisit this decision if you insist.

I was unhappy about the duplication so I poked at it until I had #123 which seems to work. Can you check if you're OK landing that instead?

n1k0:
Fixed by #123 and d9fc21155e9186a179ea5576a99fc402f9c34890.

PR #127, opened 2015-04-08
Improved detection "algorithm", distinct tests for extraction and detection.

n1k0:
So this is is WiP, though I'd like some early feedback. Don't pay too much attention with the "algorithm" used in `isProbablyReaderable`, as I'm just drafting the foundation to have something workable first.

Feedback=? @gijsk @leibovic


n1k0:
Hmm, don't be afraid of the huuuge diff, it's basically mostly test pages moved around.

gijsk:
While I understand why you would want to do something like this, the algorithm you propose is going to be a *lot* slower, and because this runs on almost every page load, I don't think that's a good idea.

n1k0:
> the algorithm you propose is going to be a lot slower

Using `Array#some` will make the iteration stops on first matching test. In my own benchmarks, it's actually slightly faster than before (probably because of the cheaper test performed).

gijsk:
Isn't getElementsByTagName("*") going to be much slower than just getting all the `p' s ?

The other question is whether this won't overestimate where reader mode is available. While I understand that it is frustrating to have an article that you really want to see in reader mode not show the button, I think it is a worse user experience to have the button show and have it result in an error.

Not all pages should be readerable, and that is OK, as far as I'm concerned. :-)

n1k0:
> Isn't getElementsByTagName("*") going to be much slower than just getting all the p s ?

Well, this is what chome is already doing http://mxr.mozilla.org/mozilla-central/source/toolkit/components/reader/ReaderMode.jsm#230

> I think it is a worse user experience to have the button show and have it result in an error.

I kinda agree, sadly. Well, as stated in the bug description, my main goal here is first to setup a good test environment so we can start refining the algorithm a little more safely, a little less empirically (even though…)

leibovic:
>> Isn't getElementsByTagName("*") going to be much slower than just getting all the p s ?

> Well, this is what chome is already doing http://mxr.mozilla.org/mozilla-central/source/toolkit/components/reader/ReaderMode.jsm#230

This only happens after the user hits the button, where we're more okay with a performance hit.

gijsk:

PR #132, opened 2015-04-09
Don't look at banners and skyscrapers, remove <noscript> elements

gijsk:
This fixes [bug 1152817](https://bugzilla.mozilla.org/show_bug.cgi?id=1152817).


leibovic:
This looks reasonable to me, but travis is still failing, so I think more testcases need updating.

gijsk:
D'oh, thanks for the ping. Another jsdom vs. JSDOMParser difference regarding HTML encoding... fixed it by adjusting the testcase, it passes locally for me now.

PR #135, opened 2015-04-13
Bug 1147584 - Don't strip unlikely <a>s, and replace useless <a>s with textContent

gijsk:
@leibovic this WFM. The issue was using a hardcoded reference to `Text` which works with JSDOMParser but not with other DOMs. I implemented `createTextNode`, tweaked the escaping of `textContent` into `innerHTML` and now it works.


leibovic:
Thanks for figuring this out!

PR #146, opened 2015-04-16
Detect search engine results pages and make sure isProbablyReaderable returns false

gvanwaelvelde:
This should detect search engine results pages and make isProbablyReaderable return false.
It's not perfect though, as it will have some false positives (some search results will slip through).
But it seems to do the job quite well. Maybe it's worth taking a look at? Thanks.
https://bugzilla.mozilla.org/show_bug.cgi?id=1151087

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1642,6 +1642,13 @@ Readability.prototype = {
    * @return boolean Whether or not we suspect parse() will suceeed at returning an article object.
    */
   isProbablyReaderable: function() {
+    var nodes = this._doc.getElementsByTagName('input');
+    for (var i = 0; i < nodes.length; i++) {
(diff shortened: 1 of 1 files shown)


n1k0:
Hmm this breaks a lot of tests, though I can agree the reported error information isn't helpful at all. Filed #147 about this. 

Also this patch would add a very specific test for search engines, which would need a few test cases to be added to our current suite so we'd be confident we actually cover them without breaking other things.

leibovic:
I talked with @gijsk about this earlier today, and in Firefox code, we can know if a URL is a search result from one of your installed search engines, so we may have a better way to do this in the browser chrome. I realize that this doesn't solve this issue more generically for other consumers of this library, but `isProbablyReaderable` was originally added for the sake of browser UI anyway.

That being said, if we do use the Firefox search service to compare a URL, that logic wouldn't be able to live in this repo.

PR #170, opened 2015-04-17
Added support for getElementsByClassName to JSDOMParser.

n1k0:
While debugging _things_, I've added support for `getElementsByClassName` to JSDOMParser. It's convenient while not too slow, though I still miss a good use case for readability just yet.

Let me isolate and park this work here for now, so if we ever need it, there it is.


gijsk:

PR #172, opened 2015-04-21
Fixes #130 - Using js-beautify for HTML formatting.

### LINKED ISSUE #130: Consider switching from html to js-beautify for HTML prettification
In my latest experiments on other fronts, I've been pleased with the HTML reformatting provided by [js-beautifier](https://www.npmjs.com/package/js-beautify) (yeah the name is misleading, it actually supports HTML prettification).

I've never been really happy with our prettyPrint function from the html package (the strange resulting indentation more specifically), so I think it's a good opportunity to switch to something better.

n1k0:
Large diff, admittedly.

r=? @leibovic @Gijsk

### DIFF
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -1,4 +1,4 @@
-var prettyPrint = require("html").prettyPrint;
+var prettyPrint = require("./utils").prettyPrint;
 var jsdom = require("jsdom").jsdom;
 var chai = require("chai");
 chai.config.includeStack = true;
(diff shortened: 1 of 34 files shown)


test/test-readability.js
```diff
@@ -8,11 +8,14 @@ var readability = require("../index");
 var Readability = readability.Readability;
 var JSDOMParser = readability.JSDOMParser;
 
-var testPages = require("./bootstrap").getTestPages();
+var testPages = require("./utils").getTestPages();
 
 function runTestsWithItems(label, beforeFn, expectedContent, expectedMetadata) {
   describe(label, function() {
+    this.timeout(5000);
```
n1k0: Note: this was triggering intermittent failures before, there I fixed it.

test/utils.js
```diff
@@ -22,3 +22,19 @@ exports.getTestPages = function() {
...
+  return prettyPrint(html, {
+    "indent_size": 4,
+    "indent_char": " ",
+    "indent_level": 0,
+    "indent_with_tabs": false,
+    "preserve_newlines": false,
+    "break_chained_methods": false,
+    "eval_code": false,
+    "unescape_strings": false,
+    "wrap_line_length": 0,
```
gijsk: I would really like lines to wrap, actually... this diff is hard to review in part because of the lack of wrapping. Can we change this?

gijsk:
Discussed this on IRC, let's do this.

PR #167, opened 2015-04-17
Fixes #164 - Add support for title alt semantic metadata.

n1k0:
Let's favor semantic headlines when they're available. r=? @Gijs @leibovic

PR #175, opened 2015-04-21
Fixes #174 - Remove aggressive article title formatting rule.

### LINKED ISSUE #174: Parts of the headline are missing in Reader View for pages from webmd.com
**Affected**:
* 38.0b6-build1 (20150420134330)
* 39.0a2 (2015-04-20)
* 40.0a1 (2015-04-20)

**Sample page showing this issue**:
http://www.webmd.com/news/20150417/superbugs-what-they-are

**Notes**:
The headline for the sample is:
> Superbugs: What They Are and How You Get Them

but *Superbugs:* is missing from it in Reader View.

n1k0:
Refs https://github.com/mozilla/readability/issues/174#issuecomment-94831420

r=? @gijsk

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -165,6 +165,21 @@ Readability.prototype = {
   },
 
   /**
+   * Concat all nodelists passed as arguments.
+   *
+   * @return ...NodeList
+   * @return Array
+   */
+  _concatNodeLists: function() {
+    var slice = Array.prototype.slice;
+    var args = slice.call(arguments);
+    var nodeLists = args.map(function(list) {
+      return slice.call(list);
+    });
+    return Array.prototype.concat.apply([], nodeLists);
+  },
+
+  /**
    * Converts each <a> and <img> uri in the given element to an absolute URI.
    *
    * @param Element
@@ -245,10 +260,24 @@ Readability.prototype = {
       if (curTitle.split(' ').length < 3)
         curTitle = origTitle.replace(/[^\|\-]*[\|\-](.*)/gi,'$1');
     } else if (curTitle.indexOf(': ') !== -1) {
-      curTitle = origTitle.replace(/.*:(.*)/gi, '$1');
+      // Check if we have an heading containing this exact string, so we
+      // could assume it's the full title.
+      var headings = this._concatNodeLists(
+        doc.getElementsByTagName('h1'),
+        doc.getElementsByTagName('h2')
+      );
+      var match = this._someNode(headings, function(heading) {
+        return heading.textContent === curTitle;
+      });
 
-      if (curTitle.split(' ').length < 3)
-        curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
+      // If we don't, let's extract the title out of the original title string.
+      if (!match) {
+        curTitle = origTitle.substring(origTitle.lastIndexOf(':') + 1);
+
+        // If the title is now too short, try the first colon instead:
+        if (curTitle.split(' ').length < 3)
+          curTitle = origTitle.substring(origTitle.indexOf(':') + 1);
+      }
     } else if (curTitle.length > 150 || curTitle.length < 15) {
       var hOnes = doc.getElementsByTagName('h1');
 
--- a/test/test-pages/webmd-2/expected-metadata.json
+++ b/test/test-pages/webmd-2/expected-metadata.json
@@ -1,0 +1,6 @@
+{
+  "title": "Superbugs: What They Are and How You Get Them",
+  "byline": "By Kelli  Miller\n                                                    WebMD Health News",
+  "excerpt": "Drug-resistant bacteria, dubbed",
+  "readerable": true
+}
--- a/test/test-pages/webmd-2/expected.html
+++ b/test/test-pages/webmd-2/expected.html
@@ -1,0 +1,18 @@
+<div id="readability-page-1" class="page">
+    <div id="textArea">
+        <h3></h3>
+        <p>April 17, 2015 -- Imagine being sick in the hospital with a <a href="http://www.webmd.com/a-to-z-guides/bacterial-and-viral-infections" onclick="return sl(this,'','embd-lnk');" class="">bacterial infection</a> and doctors can't stop it from spreading. This so-called "superbug" scenario is not science fiction. It's an urgent, worldwide worry that is prompting swift action.</p>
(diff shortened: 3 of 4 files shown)


gijsk:
I'm reluctant to do this because I think that given that some sites will do this on all their titles, the number of pages where this rule is a good one is likely to be higher than the number of pages where this doesn't happen.

The ideal case would be doing some kind of analysis of the text-to-be-removed and comparing it to the site title, but I don't know where we'd get the site title from.

@leibovic @dolske , thoughts?

pdehaan:
Wouldn't it be easier to search for an "og:title" or meta title and use that instead (if available)?

For example, on http://www.huffingtonpost.com/stephen-balkam/ex-machina-the-consciousness-test_b_7096258.html?utm_hp_ref=technology&ir=Technology Reader View shows the title as:

> _'Ex Machina': The Consciousness Test | Stephen Balkam_

...Which happens to be the main page's `<title>`:
```html
<title>'Ex Machina': The Consciousness Test&nbsp;|&nbsp;Stephen Balkam</title>
```

But it may have been easier to get the parsed/specified title from the specified og:title metadata:
```html
<meta property="og:title" content="'Ex Machina': The Consciousness Test"/>
```

---

In fact, I believe this is the logic that embed.ly uses to grab  page titles:

> **`title`** &mdash; The title of the resource. It’s picked in the following order:
>
> - The rss entry’s title
> - The oEmbed title
> - The open graph title
> - The `meta` title tag
> - The `title` attribute in the `head` element

&mdash; via http://embed.ly/docs/api/extract/endpoints/1/extract#response-attributes

n1k0:
> But it may have been easier to get the parsed/specified title from the specified og:title metadata

This has been implemented in #167, which has landed. The problem is that there's no semantic metadata in this case.

We could add support for more fallback strategies, but for websites implementing none, we still need to do something.

n1k0:
> I'm reluctant to do this because I think that given that some sites will do this on all their titles, the number of pages where this rule is a good one is likely to be higher than the number of pages where this doesn't happen.

To be frank, I have no idea if this is actually true, as I don't have access to numbers/metrics about it. I just decided to go with more inclusion than damaging the meaning carried by the title. 

Indeed, I wish we had a way to extract the website name.

gijsk:
To be clear, I have no data either. I just assume that the people adding that logic knew of (high-profile) sites that formatted their titles in this way...

n1k0:
New attempt, paired with @ametaireau and @leplatrem.

Readability.js
```diff
@@ -245,10 +260,24 @@ Readability.prototype = {
...
+        doc.getElementsByTagName('h2')
+      );
+      var match = this._someNode(headings, function(heading) {
+        return heading.textContent === curTitle;
+      });
 
-      if (curTitle.split(' ').length < 3)
-        curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
+      if (!match) {
+        curTitle = origTitle.replace(/.*:(.*)/gi, '$1');
```
gijsk: `origTitle.replace(/^[^:]*:/, '');` seems more appropriate - don't need to do this more than once, don't need to keep the rest of the string and replace it back in - we can just strip the first bit. Could even do this with `indexOf` and `.substring`

Readability.js
```diff
@@ -245,10 +260,24 @@ Readability.prototype = {
...
+
+        if (curTitle.split(' ').length < 3)
+          curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
```
gijsk: I'm not sure why this needs to do two replaces... can you elaborate?

Thanks for sticking to this, btw!

Readability.js
```diff
@@ -245,10 +260,23 @@ Readability.prototype = {
...
+      );
+      var match = this._someNode(headings, function(heading) {
+        return heading.textContent === curTitle;
+      });
 
-      if (curTitle.split(' ').length < 3)
-        curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
+      // If we don't, let's extract the title out of the original title string.
+      if (!match) {
+        curTitle = origTitle.replace(/.*:(.*)/gi, '$1');
```
gijsk: `origTitle.replace(/^[^:]*:/, '');` seems more appropriate - don't need to do this more than once, don't need to keep the rest of the string and replace it back in - we can just strip the first bit. Could even do this with `indexOf` and `.substring`

(resurfacing this comment as the diff got updated...)

Readability.js
n1k0: We didn't change this line, just moved it.

Readability.js
n1k0: Same as above.

Readability.js
gijsk: I know that, but you added another replace before it... so now there are two `replace` calls that strip out "foo:" and the second one overrides the first, but it does almost the same thing because it uses the same source string (instead of being cumulative).

n1k0: We added no supplementary replace; previous code was:

```js
    } else if (curTitle.indexOf(': ') !== -1) {
      curTitle = origTitle.replace(/.*:(.*)/gi, '$1');

      if (curTitle.split(' ').length < 3)
        curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
    } else if (curTitle.length > 150 || curTitle.length < 15) {
```

New code is:

```js
    } else if (curTitle.indexOf(': ') !== -1) {
      // Wheck if we have an heading containing this exact string, so we
      // could assume it's the full title.
      var headings = this._concatNodeLists(
        doc.getElementsByTagName('h1'),
        doc.getElementsByTagName('h2')
      );
      var match = this._someNode(headings, function(heading) {
        return heading.textContent === curTitle;
      });

      // If we don't, let's extract the title out of the original title string.
      if (!match) {
        curTitle = origTitle.replace(/.*:(.*)/gi, '$1');

        if (curTitle.split(' ').length < 3)
          curTitle = origTitle.replace(/[^:]*[:](.*)/gi,'$1');
      }
    } else if (curTitle.length > 150 || curTitle.length < 15) {
```

The previous code colon-test codeblock has been simply moved into the new `if (!match)` one.

gijsk: Can you add a comment that the second thing is trying to detect a now-too-short title and then does a more conservative replace?

Also, this can be written more simply as:

```
        curTitle = origTitle.replace(/.*:/, '');

        if (curTitle.split(' ').length < 3)
          curTitle = origTitle.replace(/[^:]*:/,'');
```

TBH, I'm not sure that this code makes much sense the way it is, but let's leave it with this change.

gijsk: or, in fact, using indexOf/lastIndexOf, as we're already guaranteed there's at least one colon if we're even in this block.

PR #177, opened 2015-04-22
Add User-Agent when generating the test case

ametaireau:
Otherwise, some website don't answer correctly (for instance the IETF website wrt internet drafts). Paired with @n1k0

### DIFF
--- a/test/generate-testcase.js
+++ b/test/generate-testcase.js
@@ -6,6 +6,7 @@ var jsdom = require("jsdom").jsdom;
 var prettyPrint = require("./utils").prettyPrint;
 var serializeDocument = require("jsdom").serializeDocument;
 var http = require("http");
+var urlparse = require("url").parse;
 
 var readability = require("../index");
 var Readability = readability.Readability;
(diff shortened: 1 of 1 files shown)


test/generate-testcase.js
```diff
@@ -54,7 +55,10 @@ function fetchSource(url, callbackFn) {
   if (url.indexOf("https") == 0) {
     client = require("https");
   }
-  client.get(url, function(response) {
+  var options = urlparse(url);
+  options.headers = {'User-Agent': 'readability.js'}
```
ametaireau: I'm not actually sure we should be doing it this way, or should create a new object instead of adding properties to the one returned by urlparse.

gijsk: This seems fine to me, but I would prefer to send something more resembling Firefox. I know that it's lying, but a surprising number of sites adjust content that you get, and we should probably try to get content that vaguely resembles what Firefox gets.

n1k0:
Landed through 5912e0c87207af097df080b344ec36b5a1c007ec with @gijsk's comment addressed.

Issue #143, opened 2015-04-15
No readability results for https://www.mozilla.org/en-US/firefox/39.0a2/firstrun/


pdehaan:
Was testing something in Aurora and noticed that https://www.mozilla.org/en-US/firefox/39.0a2/firstrun/ (default page for new profile) shows me a reader view icon in the address bar, but clicking it shows:

> #### _"Failed to load article from page"_

---

![failed_to_load_article_from_page_and_new_issue_ _mozilla_readability_and_untitled_3](https://cloud.githubusercontent.com/assets/557895/7171003/14f08c6c-e388-11e4-85c6-a0c534aabfca.png)

n1k0:
My WIP patch in #142 seems to fix this:

![screen shot 2015-04-16 at 08 53 27](https://cloud.githubusercontent.com/assets/41547/7175551/27fdba94-e416-11e4-8d48-282a933a5d4e.png)

n1k0:
This has been fixed by #185.

PR #142, opened 2015-04-15
Better detection of natural main content node.

n1k0:
This is an early attempt at improving early detection of natural good candidate semantically identifying a main content node, when available.

Work in progress, don't review just yet.

PR #185, opened 2015-04-24
Fixes #139 #143: Added more weight to section tags.

### LINKED ISSUE #139: No reader view results for https://www.mozilla.org/en-US/firefox/desktop/customize/
### Steps to reproduce:
1. Go to https://www.mozilla.org/en-US/firefox/desktop/customize/ (one of the default bookmarks in a new profile).
2. Enter reader view.

### Actual results

> #### Failed to load article from page

### Expected results
Reader view not available, or a scraped page.

---

failed_to_load_article_from_page_and_new_issue_ _mozilla_readability

n1k0:
This patch fixes #139 and #143 by adding `<section>` tags to the list of elements to score by default, while not breaking other tests.

This is a far less invasive patch than in #142, and my manual testing using FFx didn't bring more failures than before.

Thoughts? @gijsk @leibovic


gijsk:
This looks great to me. I worry that I might need to turn these back into not-probably-readerable when I alter that algorithm, but otherwise it should work. I'll see if I can avoid that, and in any case, we can cross that bridge when we get to it.

gijsk:
@n1k0 fwiw, when I saw this (too much on my mind, so this is a bit late, sorry), I wondered if it would make sense to include `<article>` in the scored element things as well. Does that break a lot of things?

n1k0:
I tried, it basically adds more noise to extracted contents, eg. including surrounding tags more often (in medium-1), include comments (lifehacker-post-comment-load)… I attempted at fine tuning this, but didn't manage to do it in a reliable manner, so…

PR #186, opened 2015-04-24
Bug 800165 - Remove hidden elements from readability output

leibovic:
r? @gijsk @n1k0 

I still need to figure out how to solve the bustage this is causing with our jsdom tests. It looks like calling `node.style` sets an empty style attribute on the node, which is messing up our output. The thing is, we do have logic to remove style attributes in `_cleanStyles`, and I'm not sure why that's failing to remove these new attributes.


n1k0:
This is indeed an issue with jsdom:

```
> doc = jsdom.jsdom("<html><body><p>plop</p></body></html>")
> p = doc.querySelector("p")
> p.style
> jsdom.serializeDocument(doc)
'<html><head></head><body><p style="">plop</p></body></html>'
```

Maybe cloning the element node before checking for its style would solve the issue? Though that's not supported by JSDOMParser atm, and will much likely impact performance. Will think of a better solution.

n1k0:
Now I think about this feature, I'm not 100% sure we actually want to exclude whatever is hidden; think of an article of the following form:

```html
<article>
  <section>
    <p>Lorem…</p>
    <p><a href="#" onclick="$('#more-content').show()">Read more</a></p>
  </section>
  <section id="more-content" style="display:none">
    … ipsum dolor sit amet…
  </section>
```

Readability.js
```diff
@@ -613,6 +613,12 @@ Readability.prototype = {
       var node = this._doc.documentElement;
 
       while (node) {
+        // Remove hidden elements.
+        if (node.style && node.style.display === "none") {
```
DavidBruant: Maybe also test the presence of the `hidden` attribute?
https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute

gijsk:
> Now I think about this feature, I'm not 100% sure we actually want to exclude whatever is hidden; think of an article of the following form:

> ```html
<article>
  <section>
    <p>Lorem…</p>
    <p><a href="#" onclick="$('#more-content').show()">Read more</a></p>
  </section>
  <section id="more-content" style="display:none">
    … ipsum dolor sit amet…
  </section>
</article>
```

Yeah, this is tricky. I think we should err on the side of not including anything that was originally hidden. It's not possible for us to predict what content may become visible later, and clearly in this case the user can go to the website, click the link, and then click the reader button, and it'll Just Work. The reverse isn't true (ie if there's a bunch of hidden content, like the error messages on etherpad.m.o, for instance, then we don't do the right thing and the user can't fix that).

gijsk:
Regarding the test issue, we should figure out why the style removal code isn't working.

gijsk:
Huh, so changing @n1k0 's snippet to `p.removeAttribute('style')` does... nothing.

n1k0:
This is definitely a bug in jsdom. The issue here is that the 3.x branch isn't maintained anymore, while it's the last one being supported with NodeJS (later versions require io.js).

We could definitely move to io.js, hoping we could then see this bug eventually addressed (latest 5.1.0 still has the bug), though that seems highly overkill :/

n1k0:
Filed https://github.com/tmpvar/jsdom/issues/1109, let's see how it goes.

gijsk:
jsdom has since been fixed and should now work with node rather than io.js. We should look at this again. :-)

gijsk:
This got fixed in #283 / #285 . \o/

PR #187, opened 2015-04-24
Fixes #184 - Don't strip class names from article content

### LINKED ISSUE #184: Don't strip class names and ids from article content
Right now we inconsistently maintain class names on readability content. It looks like the original intention was to actually strip all these class names from content, but we've found that preserving them can be useful for applying styles to common class names (e.g. `.visibility-hidden` should be hidden).

Now that we landed https://bugzilla.mozilla.org/show_bug.cgi?id=1154028, these class names shouldn't interfere with the reader view controls, so we could try to preserve even more class names for more special CSS rules.

leibovic:
r? @gijsk @n1k0 

I realized that although we fixed bug 1154028 for desktop, we'll want to fix bug 1158194 for mobile as well to avoid problems that might arise from this change. I'll just raise the priority of that bug on my list to try to get it in before the next Readability merge.


gijsk:
LGTM!

Issue #157, opened 2015-04-17
Reader View is available for specific tumblr.com index pages


avaida:
**Affected**:
* 38.0b5-build1
* 39.0a2 (2015-04-16)
* 40.0a1 (2015-04-16)

**Sample pages**:
* https://www.tumblr.com/explore/text
* https://www.tumblr.com/explore/staff-picks

gijsk:
So this page is a pain. It basically has a lot of `<article>` tags and so we'll just end up picking whichever one seems biggest. I have no idea how to teach readability (or isProbablyReaderable) that this is a bad idea.

n1k0:
Silly idea, but would disabling readerability for pages where there are more than X `<article>` tags? I know… :/

Issue #149, opened 2015-04-16
'Failed to load article from page' when entering Reader View via bbc.com


adalucinet:
Affected branches:
Beta 38.0b4 (20150415030206), Aurora 39.0a2 (2015-04-16) and Nightly 40.0a1 (2015-04-15)

Sample pages showing this issue:
http://www.bbc.com/sport/0/
http://www.bbc.com/culture

n1k0:
These portal-type pages shouldn't be detected as *readerable* if you ask me.

FWIW, Safari doesn't.

n1k0:
Fixed by #188.

Issue #145, opened 2015-04-16
 No reader mode offered for IETF draft spec page


n1k0:
http://tools.ietf.org/html/draft-dejong-remotestorage-04

ametaireau:
It seems the issue is not with the readability.js implementation, but rather with the article detection. I don't know where this code leaves though, and any pointers would be appreciated (because obviously, this is behaving in a wrong way).

ametaireau:
(I cannot close here but please do ;))

leibovic:
This sounds similar to this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1154295

gijsk:
Yeah... not sure what to do about this. :-\

msujaws:
We can detect if it is a .txt document pretty easily. We will need the parser though to break up the blocks of text and create paragraphs out of them. Could we just look for consecutive line-breaks (more than 1)?

msujaws:
In privileged JS, you can check to see if document.mozSyntheticDocument is true. If so, then check that the body only has one child (<pre>) and that the <pre> has no children.

I don't think we should care about the length of these documents. Checking <pre>.textContent.length is gonna be a very long string for most .txt documents.

gijsk:
This isn't a .txt document though...

n1k0:
Fixed by #188.

Issue #141, opened 2015-04-15
Reader view available for Twitter, fails to load content


pdehaan:
### Steps to reproduce:
1. Go to any Twitter acct, like https://twitter.com/realandreasgal
2. Click the "Enter Reader View" button up in the address bar.

### Actual results:
> #### _Failed to load article from page_

### Expected results:
No reader view available for twitter.com.

---

![failed_to_load_article_from_page_and_new_issue_ _mozilla_readability_and____github_com___348__-_peterdehaan_gmail_com_-_gmail_and_el_presidente_gal___realandreasgal____twitter_and_issues_ _mozilla_readability_ _github](https://cloud.githubusercontent.com/assets/557895/7150262/62817d80-e2cd-11e4-9a6b-b54e611c42bf.png)

pdehaan:
Actually, to clarify, I don't think the _generic_ https://twitter.com/RealAndreasGal/ page should be reader-able, but I'd expect _specific_ Twitter tweets (ie: [twitter.com/RealAndreasGal/status/588582184258142208]( https://twitter.com/RealAndreasGal/status/588582184258142208)) to be available in reader view since they are static (although obviously not much content).

pdehaan:
Not sure how/if this would work with tweets that have lots of replies, such as https://twitter.com/elonmusk/status/588144086755999744

I think a lot of these questions come back to the larger question of having lots of site-specific logic/CSS in readability for a lot of the major sites.

gijsk:
Yeah, I really don't know how to make readability and/or isProbablyReaderable figure out what to do here. From a DOM structure perspective, there is almost no difference between the two "types" of tweet pages you listed. Suggestions?

pdehaan:
I only have one idea, but it/me is stupid:

Profile page [1]: https://twitter.com/elonmusk
Tweet page [2]: https://twitter.com/elonmusk/status/588144086755999744

I noticed that the profile page [1] doesn't seem to have any open graph metadata tags:
```
$('meta')
[
  <meta charset=​"utf-8">​,
  <meta name=​"msapplication-TileImage" content=​"/​/​abs.twimg.com/​favicons/​win8-tile-144.png">​,
  <meta name=​"msapplication-TileColor" content=​"#00aced">​,
  <meta name=​"swift-page-name" id=​"swift-page-name" content=​"profile">
​]
```

But if I go to a tweets/details page [2], I do get some more open graph metadata tags:
```
$('meta')
[
[...]

PR #188, opened 2015-04-24
Make isProbablyReaderable include <pre>, and deal with long <br>-separat...

gijsk:
...ed paragraphs and/or shorter-than-5-paragraph text and such.

Things this fixes:

* The remaining page on https://bugzilla.mozilla.org/show_bug.cgi?id=1153384
* #149 for BBC sport - culture normally returns something useful for me.
* #116 ( https://wordpress.com/fresh/ ) with the isVisible helper passed in by ReaderMode.jsm (separate hg change for which I'll need to file a bug)
* #145
* other things which have have really long paragraphs, or lists with paragraphs (that shouldn't be readerable), or things with a large number of paragraphs that used to all fall under the 200 line before https://bugzilla.mozilla.org/show_bug.cgi?id=1149859 .

Things this doesn't fix:
* twitter (#141)
* tumblr index pages (#157)

Things it breaks:
* one of the mozilla testcases will no longer offer reader mode. I think that's OK, looking at the page.
* maybe perf? I need to trypush for this.

In the meantime, I'd appreciate thoughts. @n1k0 / @leibovic  ?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -182,6 +182,15 @@ Readability.prototype = {
     return Array.prototype.concat.apply([], nodeLists);
   },
 
+  _getAllNodesWithTag: function(node, tagNames) {
+    if (node.querySelectorAll) {
+      return node.querySelectorAll(tagNames.join(','));
+    }
+    return [].concat.apply([], tagNames.map(function(tag) {
+      return node.getElementsByTagName(tag);
+    }));
+  },
+
   /**
    * Converts each <a> and <img> uri in the given element to an absolute URI.
    *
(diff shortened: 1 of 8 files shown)


gijsk:
ReaderMode.jsm change: https://gist.github.com/gijsk/6e6f3202616bddbd4e1a

Readability.js
```diff
@@ -1678,15 +1678,18 @@ Readability.prototype = {
...
    * @return boolean Whether or not we suspect parse() will suceeed at returning an article object.
    */
-  isProbablyReaderable: function() {
-    var nodes = this._doc.getElementsByTagName("p");
-    if (nodes.length < 5) {
-      return false;
-    }
-
-    var possibleParagraphs = 0;
+  isProbablyReaderable: function(helperIsVisible) {
```
n1k0: I wish we could have `helperIsVisible` part of the API, but I suspect this is impossible to achieve relying on JSDOMParser… 

Though for clients relying on a real DOM impl., maybe we could define some fallback? I've been using [this](https://github.com/n1k0/casperjs/blob/11bb84995fd543440891a14baff5b8e1bb9b7866/modules/clientutils.js#L156-L171) in casperjs with good reliability so far.

Readability.js
```diff
@@ -1678,15 +1678,18 @@ Readability.prototype = {
...
+    var nodes;
+    if (typeof this._doc.querySelectorAll != "undefined")
```
n1k0: This test should be moved within `_concatNodeList`.

Readability.js
```diff
@@ -1678,15 +1678,18 @@ Readability.prototype = {
...
+      nodes = this._doc.querySelectorAll("p, pre");
+    else
+      nodes = this._concatNodeLists(this._doc.getElementsByTagName("p"), this._doc.getElementsByTagName("pre"));
+
+    var score = 0;
     for (var i = 0; i < nodes.length; i++) {
```
n1k0: Nit: reuse `this._forEachNode` maybe?

n1k0: Actually, as we return as soon as we've found something, `_someNode` would be a much more appropriate (and fast) candidate.

n1k0:
> one of the mozilla testcases will no longer offer reader mode. I think that's OK, looking at the page.

I agree.

> maybe perf? I need to trypush for this.

Adding some benchmarks dedicated to measure `isProbablyReaderable` performance qo we could watch for regression would indeed be a good idea.

Readability.js
n1k0: Note that as stated in a general comment, we should add benchmarks for this specific function to watch for possible perf regressions.

Readability.js
gijsk: That's somewhat surprising to me... if you ask a child/descendant for display/visibility and its parent/ancestor is display: none, the child won't have that style (but obviously still won't be visible).

Readability.js
gijsk: I don't understand how you would do that - `_concatNodeList` takes two node lists; at that point it's too late for `querySelectorAll`...

n1k0: The idea would be to rely on a root node and an array of tag names passed as args instead of NodeList directly. Maybe that wouldn't be the most beautiful API on earth though:

```js
_concatNodeList: function(node, tagNames) {
  var nodes, doc = this._doc;
  if (typeof doc.querySelectorAll != "undefined") {
    nodes = doc.querySelectorAll(tagNames.join(","));
  } else {
    nodes = this._concatNodeLists.apply(this, tagNames.map(function(tag) {
      return doc.getElementsByTagName(tag);
    }));
  }
  // …
}
```

Readability.js
n1k0: Good point, dunno why I'm not looking for ancestor styles here… I'll file an issue. Thanks.

gijsk: I mean, we can write a fallback relying on `clientHeight`, but that requires not just a DOM implementation but a layout implementation, which implies a CSS implementation, which 'even' `jsdom` doesn't have, so I'm not sure how useful it is in the meantime. I guess it might be useful for the iOS version?

gijsk:
> Adding some benchmarks dedicated to measure `isProbablyReaderable` performance in this repo so we could watch for regression would indeed be a good idea.

Added one in this version of the PR. I ran it, and it seems this deteriorates performance by approximately 0.05ms (from 3400 ops/s to 2700 ops/s) on my machine. I think this is acceptable in the grand scheme of things, though we probably shouldn't make it much heavier than this...

n1k0:
LGTM, r+

PR #189, opened 2015-04-27
Fixes #150 - Keep article intermediary headings.

### LINKED ISSUE #150: LWN.net missing article headers
https://lwn.net/Articles/637393/

Filed from https://bugzilla.mozilla.org/show_bug.cgi?id=1154781

n1k0:
We were aggressively stripping out intermediary headings within extracted article contents if they had a link density greater than 1/3rd. 

In real world, I believe these headings can have as many links as they want (none, a single one to link to some other part/page, a few…), so I'm removing the link density rule entirely here, while keeping the class weight check.

This fixes #150.


gijsk:
Yeah, I don't see the point of including some headers and removing others just depending on whether they have links. Having just half the headers is unlikely to be useful, I'd imagine.

PR #190, opened 2015-04-27
Improved author metadata detection.

n1k0:
Some pages declare byline using a `<meta property="author" content="*">` tag, while we're only checking for `<meta name=` attr.

This patch addresses that and adds a test case.

r=? @gijsk @leibovic


leibovic:
LGTM!

PR #193, opened 2015-04-29
Fixes #113 - Score great grandparent nodes.

### LINKED ISSUE #113: Specific pages from ehow.com can't be displayed in Reader View
**Reproducible on**: 
Beta 38.0b1-build1 (20150330154247)

**Sample page showing this issue**:
http://www.ehow.com/how_2042752_build-terrarium.html

n1k0:
Spent a bunch of time investigating this one, went with scoring great grandparent nodes with good outcome — and no breakage. 

~~Note: I'm crying blood staring at the implementation, but I'm trying to focus on getting results in the shortest possible amount of time…~~

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -591,6 +591,18 @@ Readability.prototype = {
     return false;
   },
 
+  _getNodeAncestors: function(node, maxDepth) {
+    maxDepth = maxDepth || 0;
+    var i = 0, ancestors = [];
+    while (node.parentNode) {
+      ancestors.push(node.parentNode)
+      if (maxDepth && ++i === maxDepth)
+        break;
+      node = node.parentNode;
+    }
+    return ancestors;
+  },
+
   /***
    * grabArticle - Using a variety of metrics (content score, classname, element types), find the content that is
    *         most likely to be the stuff a user wants to read. Then return it wrapped up in a div.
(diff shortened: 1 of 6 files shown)


Readability.js
```diff
@@ -724,8 +734,12 @@ Readability.prototype = {
         // Add the score to the parent. The grandparent gets half.
         parentNode.readability.contentScore += contentScore;
 
-        if (grandParentNode)
+        if (grandParentNode && grandParentNode.readability) {
```
gijsk: `_initializeNode` should be making sure that grandParentNode always exists. Does it not?

Readability.js
```diff
@@ -724,8 +734,12 @@ Readability.prototype = {
...
           grandParentNode.readability.contentScore += contentScore / 2;
+          if (greatGrandParentNode && greatGrandParentNode.readability) {
+            greatGrandParentNode.readability.contentScore += contentScore / 4;
```
gijsk: Mm... I've wondered before now if we should just use some way of scoring that deals with this automagically (for whatever depth) but I've not figure out how best to do this. If this works, let's start here, I guess.

n1k0: Yeah, I think I can work something out here. Hold on merging :)

Readability.js
n1k0: That's betond my current understanding, but yes, in some circumstances I didn't take the time to fully investigate, `grandParentNode` can be `undefined`.

gijsk: Maybe if `typeof(grandParentNode.tagName)` is undefined... that'll not initialize readability for that node...

n1k0:
r=? @gijsk @leibovic

gijsk:
What's the perf implication here? Assuming that's OK and with the early return restored (see comment above), r=me!

gijsk:
(I won't be around to merge, so see above and merge when it's ready (and/or ask @leibovic ))

n1k0:
Perf is really similar as before :)

PR #194, opened 2015-04-30
Fixes #180 - Score intermediary headings.

### LINKED ISSUE #180: Images missing from specific buzfeed.com pages in Reader View
Affected:
38.0b6-build1 (20150420134330)
39.0a2 (2015-04-22)
40.0a1 (2015-04-22)

Page: http://www.buzzfeed.com/markdistefano/diet-pills-burns-up

n1k0:
Intermediary headings are part of content, and they most often carry weight; this patch extends the default tags to score to support them.

r=? @gijsk @leibovic

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -102,7 +102,7 @@ Readability.prototype = {
   DEFAULT_MAX_PAGES: 5,
 
   // Element tags to score by default.
-  DEFAULT_TAGS_TO_SCORE: ["SECTION", "P", "TD", "PRE"],
+  DEFAULT_TAGS_TO_SCORE: "section,h2,h3,h4,h5,h6,p,td,pre".toUpperCase().split(","),
 
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
--- a/test/test-pages/buzzfeed-1/source.html
+++ b/test/test-pages/buzzfeed-1/source.html
@@ -3550,7 +3550,6 @@
                                             <div class="sub_buzz_content">
                                             </div>
                                         </div>
-                                        <div class="grid_clear">&nbsp;</div>
                                     </div>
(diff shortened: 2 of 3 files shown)


test/test-pages/buzzfeed-1/source.html
```diff
@@ -3550,7 +3550,6 @@ <h1 id="post-title">Student Dies After Diet Pills She Bought Online “Burned He
                                             <div class="sub_buzz_content">
                                             </div>
                                         </div>
-                                        <div class="grid_clear">&nbsp;</div>
```
n1k0: Note regarding this: I had to remove this line from source as it was producing different outputs with jsdom. As we're not maintaining jsdom, I thought it was faire to simply adapt the source so we get consistent results.

Readability.js
```diff
@@ -102,7 +102,7 @@ Readability.prototype = {
   DEFAULT_MAX_PAGES: 5,
 
   // Element tags to score by default.
-  DEFAULT_TAGS_TO_SCORE: ["SECTION", "P", "TD", "PRE"],
+  DEFAULT_TAGS_TO_SCORE: "section,h2,h3,h4,h5,h6,p,td,pre".toUpperCase().split(","),
```
leibovic: It feels a bit suspicious to me that this fixes the issue of images not appearing. Is this just a coincidence because these images are siblings of some of these header elements? Should we instead be increasing the content score for elements that hold images?

I'm not too worried about landing since, since it doesn't regress the other test cases and does improve this case, but it just feels like it might be a coincidental fix.

n1k0: The idea is to score article intermediary headings in orther to better identify which part of page contents actually define the main article. We were only scoring P, TD and PRE tags; I think intermediary headings are definitely important parts of an article, carrying meaning and semantical weight. This patch simply makes the parser scoring them, and this fix the issue with the test page because the main content area is now properly detected, thanks to scoring these headings by default.

PR #203, opened 2015-05-04
Remove nav tags from article content.

n1k0:
The readable version of [this page](http://arstechnica.com/security/2015/04/16/just-released-minecraft-exploit-makes-it-easy-to-crash-game-servers/) was displaying a navigation menu (using a `nav` tag). I think it's safe to remove these by default from article content.

Note: I was surprised to see this page broken :/ Also, the readable version still includes garbage at the end. It seems we're now too optimistically declare ancestor nodes as main content root ones :/


n1k0:
Much better alternative strategy proposed in #204.

PR #204, opened 2015-05-05
Updated great grandparent node scoring.

n1k0:
This is a much better alternative to what is proposed in #203. The patch tweaks the great grandparent node scoring to ensure we're not too easily include wrapping article contents, with same test case provided and no other breakage.

r=? @leibovic @gijsk

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -732,7 +732,12 @@ Readability.prototype = {
             candidates.push(ancestor);
           }
 
-          ancestor.readability.contentScore += contentScore / (level === 0 ? 1 : level * 2);
+          // Node score divider:
+          // - parent:             1 (no division)
+          // - grandparent:        2
+          // - great grandparent+: ancestor level * 3
+          var scoreDivider = level === 0 ? 1 : level === 1 ? 2 : level * 3;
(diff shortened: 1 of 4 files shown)


Readability.js
```diff
@@ -732,7 +732,12 @@ Readability.prototype = {
             candidates.push(ancestor);
           }
 
-          ancestor.readability.contentScore += contentScore / (level === 0 ? 1 : level * 2);
+          // Node score divider:
+          // - parent:             1 (no division)
+          // - grandparent:        2
+          // - great grandparent+: 6
+          var scoreDivider = level === 0 ? 1 : level === 1 ? 2 : 6;
```
gijsk: I'm vaguely uncomfortable with the magic scoring here, but sure, we can do this...

gijsk: Actually, hang on... this means the scoring is constant for `level > 1`, right? Can we use `level * 3` instead? Does that break other testcases?

n1k0: > I'm vaguely uncomfortable with the magic scoring here, but sure, we can do this...

I'm definitely not proud of that code, though that fixes issues and I couldn't come with a better approach :/

> Can we use level * 3 instead? Does that break other testcases?

Good catch

PR #207, opened 2015-05-05
Improved embedded video elements detection.

n1k0:
This patch adds support for detecting embedded video markup where URLs are using the generic protocol scheme, eg:

```html
<iframe src="//www.dailymotion.com/embed/video/x2p552m?syndication=131181" 
  frameborder="0" width="534" height="320"></iframe>
```

r=? @gijsk @leibovic


gijsk:
rs=me

Issue #209, opened 2015-05-06
Bad parse results for blog.sfgate.com


pdehaan:
![blank_skitch_document](https://cloud.githubusercontent.com/assets/557895/7485619/a455521e-f34d-11e4-96a6-4afba88ea649.png)

## Steps to reproduce:
1. Go to http://blog.sfgate.com/ontheblock/2015/05/05/nob-hill-one-bedroom-sells-for-2-3-million/
2. Open in reader mode.

## Actual results:
It looks like reader mode is only parsing the gallery bits, but ignoring the text.

![nob_hill_one-bedroom_sells_for__2_3_million_-_on_the_block](https://cloud.githubusercontent.com/assets/557895/7485616/8dd20f5a-f34d-11e4-8892-0613cee4f7d9.png)

![nob_hill_one-bedroom_sells_for__2_3_million](https://cloud.githubusercontent.com/assets/557895/7485618/9a5e2bd2-f34d-11e4-87e8-a3d44c3a2349.png)

## Expected results:
Ignore the gallery, only give me the text.

n1k0:
> `<p class="MsoNormal">`

Do I need to say more?

PR #212, opened 2015-05-06
Refs #209 - Increase score for elements containing large amount of text.

n1k0:
Refs #209 and http://blog.sfgate.com/ontheblock/2015/05/05/nob-hill-one-bedroom-sells-for-2-3-million/

This page, visibly made using MSWord, seems to have generated, poorly structured semantics ; so this patch doesn't entirely fix the situation with it. Though:
- it boosts scores for traditional elements used to store text containing more than 300 characters, so at least we have the actual article contents now part of the readable version;
- it clears up part of the gallery thing by removing its control bar.

I know this approach is far from ideal (don't ask me why 1.5 or why 300 and not 200 or 400), we could also decide not caring too much for this specific range of websites.

### DIFF
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -18,7 +18,7 @@ function reformatError(err) {
 
 function runTestsWithItems(label, beforeFn, expectedContent, expectedMetadata) {
   describe(label, function() {
-    this.timeout(5000);
+    this.timeout(10000);
 
     var result;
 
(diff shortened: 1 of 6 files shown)


test/test-readability.js
```diff
@@ -18,7 +18,7 @@ function reformatError(err) {
 
 function runTestsWithItems(label, beforeFn, expectedContent, expectedMetadata) {
   describe(label, function() {
-    this.timeout(5000);
+    this.timeout(10000);
```
n1k0: 7000+ lines of html === burden to parse

gijsk:
Yeah, I'm wary of changing things in this way late in the game for a single article...

gijsk:
(the Ars changes this patch required are somewhat worrying, for instance)

gijsk:

PR #213, opened 2015-05-06
Fixes #198 - Avoid stripping linked list items.

### LINKED ISSUE #198: Bulleted lists disappear when switching to reader view on pocket ToS page
blank_skitch_document

### Steps to reproduce:
1. Go to https://getpocket.com/tos in latest FF Nightly.
2. Switch to readability mode.

#### Reader view
pocket__terms_of_service

#### HTML view
pocket__terms_of_service

### Expected results:
Keep the internal links.

---

Interestingly, it looks like Safari keeps the bulleted list nav, but strips the links (which makes them fairly useless as a way to quickly jump to internal bookmarks on the page):

pocket__terms_of_service_and_new_issue_ _mozilla_readability_and_pocket__terms_of_service

n1k0:
Refs #198. This also fixes an issue with the `buzzfeed-1` test case where we were erroneously stripping an image out.

r=? @gijsk @leibovic


n1k0:
ping @gijsk @leibovic

gijsk:
I... swear I commented on this already. Let me see if I can find it somewhere...

Readability.js
```diff
@@ -1655,8 +1655,6 @@ Readability.prototype = {
           toRemove = true;
         } else if (!isList && weight < 25 && linkDensity > 0.2) {
           toRemove = true;
-        } else if (weight >= 25 && linkDensity > 0.5) {
-          toRemove = true;
```
gijsk: This change affects more than just list items, as evidenced by the buzzfeed testcase. I think we need to tread slightly more carefully here.

I think an alternative fix that might well work pretty well is not counting links that are anchor links to the same page (so linkDensity will be lower). As a bonus, we should try to keep hash links non-absolute so they work in reader mode.

n1k0: > This change affects more than just list items, as evidenced by the buzzfeed testcase. I think we need to tread slightly more carefully here.

Sure, but it fixes a previous actually broken test case (missing image), add support for a new one, without breaking anything else. Also, this rules doesn't make much sense to me… removing elements with some weight but having some large link density doesn't really make sense to me. Don't forget we've already extracted our top candidate node at this point, so things like menus and navigation entries has been stripped out already…

Do you have a specific real-world use case in mind?

gijsk: I don't have a counterexample to hand, no. I am just really uncomfortable making sweeping changes like this *especially* when we don't have a good understanding of why the code is there, and considering that if we then later find out this breaks something, we won't necessarily be in a position to readjust.

I think the top candidate extraction isn't really that helpful here - if we select a container that is too high up the DOM tree, or if there are "related" links or something that come in, this will previously have stripped them and will now have kept them.

gijsk:

PR #223, opened 2015-05-13
Fixes #220 - Use article content when a single one exists in the page.

### LINKED ISSUE #220: France24.com missing content
Affected builds:
- Firefox 38.0.5 beta 1

Affected OS`s:
 - Windows 7 64-bit
 - Mac OS X 10.9.5
 - Ubuntu 14.04 32-bit

STR:
1. Visit http://www.france24.com/en/20150512-bangladesh-third-blogger-hacked-death
2. Enter reader view

Expected results:
Content is shown in reader view.

Actual results:
Improper content shown in reader view.

n1k0:
If there's a single `<article>` tag in a page, it's most likely where main contents reside.

r=? @leibovic @gijsk

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -622,6 +622,14 @@ Readability.prototype = {
       return null;
     }
 
+    var articleTags = this._doc.getElementsByTagName("article");
+    if (articleTags.length === 1) {
+      this._forEachNode(page.children, function(pageChild) {
+        page.removeChild(pageChild);
+      });
+      page.appendChild(articleTags[0]);
+    }
+
     var pageCacheHtml = page.innerHTML;
 
     // Check if any "dir" is set on the toplevel document element
@@ -926,7 +934,7 @@ Readability.prototype = {
       } else {
         return articleContent;
       }
-    }
+    } // end while(true)
   },
 
   /**
--- a/test/test-pages/remove-extra-brs/source.html
+++ b/test/test-pages/remove-extra-brs/source.html
@@ -11,9 +11,9 @@
       <br>
       <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
         tempor incididunt ut labore et dolore magna aliqua.</p>
-      <br><br><p>Ut enim ad minim veniam,
+      <br><br>Ut enim ad minim veniam,
       quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-      consequat.</p><br>
+      consequat.<br>
       <p>Duis aute irure dolor in reprehenderit in voluptate velit esse
       cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
       proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
--- a/test/test-pages/wapo-2/expected.html
+++ b/test/test-pages/wapo-2/expected.html
@@ -28,8 +28,4 @@
         <p>“Now it’s hard to see what could persuade the Palestinians” to hold up on their ICC plans, Indyk said. “That has nothing to do with negotiations, but if both sides can’t be persuaded to back down, then they will be on a trajectory that could lead to the collapse of the Palestinian Authority because it can’t pay wages anymore.</p>
         <p>“That could be an issue forced onto the agenda about the same time as a potential nuclear deal.”</p>
     </article>
-    <div class="post-body-sig-line">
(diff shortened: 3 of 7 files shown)


Readability.js
```diff
@@ -624,6 +624,11 @@ Readability.prototype = {
 
     var pageCacheHtml = page.innerHTML;
 
+    var articleTags = this._doc.getElementsByTagName("article");
+    if (articleTags.length === 1) {
+      this._doc.body.innerHTML = articleTags[0].parentNode.innerHTML;
```
gijsk: There are some issues here.

1. This will break if the page goes through the `while` loop below a second time, as we'll re-set the `innerHTML` to `pageCacheHtml` which is set directly above this block.
2. This leaves a whole lot of work inside the code here, like looking for candidates and whatnot, whereas the summary seems to indicate we just want to use the entire contents of the `article` tag. Why do it this way, then?
3. Why the `parentNode` of the `article`, which will include siblings?
4. This has a performance penalty: we'll be reparsing the HTML. I would suggest:
`this._doc.body.parentNode.replaceChild(articleTags[0].parentNode, this._doc.body)` which is almost O(1).

test/test-pages/remove-extra-brs/source.html
```diff
@@ -11,9 +11,9 @@
       <br>
       <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
         tempor incididunt ut labore et dolore magna aliqua.</p>
-      <br><br><p>Ut enim ad minim veniam,
+      <br><br>Ut enim ad minim veniam,
       quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-      consequat.</p><br>
+      consequat.<br>
```
gijsk: Why this change?

n1k0: Because jsdom was breaking on it.

Readability.js
n1k0: > 1. This will break if the page goes through the `while` loop below a second time, as we'll re-set the `innerHTML` to `pageCacheHtml` which is set directly above this block.

Good catch.

> 2. This leaves a whole lot of work inside the code here, like looking for candidates and whatnot, whereas the summary seems to indicate we just want to use the entire contents of the `article` tag. Why do it this way, then?

Because it was simple :) I can try skipping the candidate selection part though.

> 3. Why the `parentNode` of the `article`, which will include siblings?

Fair point.

> 4. This has a performance penalty: we'll be reparsing the HTML. I would suggest:
`this._doc.body.parentNode.replaceChild(articleTags[0].parentNode, this._doc.body)` which is almost O(1).

Replacing the `<body>` tag with an `<article>` one? Anyway, that breaks everything.

gijsk: >> This has a performance penalty: we'll be reparsing the HTML. I would suggest:
`this._doc.body.parentNode.replaceChild(articleTags[0].parentNode, this._doc.body)` which is almost O(1).

> Replacing the `<body>` tag with an `<article>` one? Anyway, that breaks everything.

Ah, d'oh. It's clear I didn't think this through. In that case, I guess:

`this._doc.body.innerHTML = ''; this._doc.body.appendChild(articleTags[0]);`

That or using `_setNodeTag` to make it a `<body>` again.

n1k0: I went with

```js
var articleTags = this._doc.getElementsByTagName("article");
    if (articleTags.length === 1) {
      this._forEachNode(page.children, function(pageChild) {
        page.removeChild(pageChild);
      });
      page.appendChild(articleTags[0]);
    }
```

It seems to work well, though your solution is definitely shorter :)

n1k0:
So I tried skipping the top candidate selection, though that breaks a lot of cases; that part of the code would need some heavy refactoring, but I suggest we play the safest we can here.

test/test-pages/wapo-2/expected.html
```diff
@@ -28,8 +28,4 @@
         <p>“Now it’s hard to see what could persuade the Palestinians” to hold up on their ICC plans, Indyk said. “That has nothing to do with negotiations, but if both sides can’t be persuaded to back down, then they will be on a trajectory that could lead to the collapse of the Palestinian Authority because it can’t pay wages anymore.</p>
         <p>“That could be an issue forced onto the agenda about the same time as a potential nuclear deal.”</p>
     </article>
-    <div class="post-body-sig-line">
-        <a href="http://www.washingtonpost.com/people/steven-mufson"><img class="post-body-headshot-left" src="http://img.washingtonpost.com/wp-apps/imrs.php?src=http://www.washingtonpost.com/blogs/wonkblog/files/2014/07/mufson_steve.jpg&amp;h=180&amp;w=180"></a>
-        <p class="post-body-bio has-photo">Steven Mufson covers the White House. Since joining The Post, he has covered economics, China, foreign policy and energy.</p>
-    </div>
```
n1k0: Yeah, we're now missing this, because it's outside the `<article/>` element :/

gijsk:

PR #238, opened 2015-07-19
jsdom dependency version update

ParallelMazen:
jsdom 4.0+ no longer supports nodejs (check https://github.com/tmpvar/jsdom)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "devDependencies": {
     "chai": "^2.1.*",
     "js-beautify": "^1.5.5",
-    "jsdom": "^3.1.2",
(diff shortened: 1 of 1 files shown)


gijsk:
Thanks for the patch!

https://github.com/npm/node-semver#caret-ranges-123-025-004 implies that the current version requirement should already exclude 4.0.0. Am I missing something?

pdehaan:
Yeah, you can see the supported semver ranges using the handy npm semver calculator at: http://semver.npmjs.com/

^3.1.2 will only match versions in 3.x and not the newer builds in 4.x or 5.x.

gijsk:
OK, then it sounds like this should be closed.

PR #239, opened 2015-07-24
Ignore script tags when calculating whether something is readable

wesj:
BBC stories on the iPad have lots of scripts mixed in the body element. Some of these are huge and have lots of commas which gives them huge readability scores. Readability should probably just ignore all script tags (or anything with display: none?)

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -109,6 +109,7 @@ Readability.prototype = {
   REGEXPS: {
     unlikelyCandidates: /banner|combx|comment|community|disqus|extra|foot|header|menu|related|remark|rss|share|shoutbox|sidebar|skyscraper|sponsor|ad-break|agegate|pagination|pager|popup/i,
     okMaybeItsACandidate: /and|article|body|column|main|shadow/i,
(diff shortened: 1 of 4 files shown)


wesj:
r=? @gijsk @leibovic

leibovic:
I thought that we already stripped scripts before parsing the document... 
https://github.com/mozilla/readability/blob/master/Readability.js#L1767

Is that logic not working properly? /cc @n1k0

gijsk:
@wesj I think Margaret is right, so I'm not sure what's going on here. Do you know why we're scoring these scripts at all?

n1k0:
> BBC stories on the iPad have lots of scripts mixed in the body element.

Could we have a sample url where this happens, with the generated readable html?

gijsk:
There's a sample in the commit in this pull-request, right? I haven't had time to check if/how that's broken with readability on the commandline as-is (I mean, it must be, considering the test presumably only passes after the code changes?). I probably won't have time soon, either, so if you can look at that, @n1k0 , that would be great.

gijsk:
I just verified that this testcase essentially passes with current tip of Readability.js, so I'm closing this in favour of #262 .

Issue #247, opened 2015-09-22
isProbablyReadable false positive when page uses <br> instead of <p>


salty-horse:
The websites Livejournal and its fork, Dreamwidth, seem to publish blog posts using `<br>` tags for paragraphs instead of `<p>`.

[Example page on dreamwidth.org](https://ladybusiness.dreamwidth.org/2015/08/02/station-eleven-emily-st-john-mandel.html)

As a consequence, pages that Readability can parse() easily, fail on the isProbablyReaderable() test, which means that Firefox doesn't show the Reader View button.

I have raised the issue to @gijsk on IRC. Here is the chat log, posted by request.

```
<Gijs> salty-horse: if instead of all the <br> the text in the content used <p> tags, it'd work. :)
<Gijs> salty-horse: and yeah, sadly github didn't let me have "gijs"
<Gijs> (I was late to the github party, what can I say...)
<Gijs> salty-horse: isProbablyReaderable is very stupid
<Gijs> somewhat intentionally
<Gijs> it runs on every page load
<Gijs> so for hopefully obvious reasons we want it to be very very fast :)
<Gijs> making it smarter will mean making it slower
<salty-horse> so, what's the recommended course of action? :)
<Gijs> if you control the markup of the page, fixing the markup seems like the easiest solution
<salty-horse> thinking outside of Readability, why not allow FF to "force" reader view?
<Gijs> if you just want to use readability more often, it would be pretty trivial to create an add-on that forced that button to always be visible
<salty-horse> Gijs, I have a feeling it's an issue with the Livejournal authoring tools, not the specific page
<Gijs> yes, that's plausible
<Gijs> that or the markup was copied from msword
<Gijs> or... well, there are a number of other options
<Gijs> either way, if you don't control the markup, you could try to force the button to be visible more often
<Gijs> as for why we don't do that
<Gijs> it's an 80/20 type thing
<Gijs> the failure behaviour for the button not working is pretty shitty UX
<Gijs> it used to be "reload the original page"
<Gijs> IIRC we changed it now to just show you a message along the lines of "uh oh, it seems that didn't work"
<Gijs> because the lack of feedback was concerning
<Gijs> ie, there are obviously pages where readability just goes "uhm, I can't make this readerable"
<Gijs> either correctly or incorrectly
<salty-horse> and there's no will to change isProbablyReaderable to allow <br>'s because it's either too slow or creates too many false positives?
<Gijs> hm
<Gijs> it would depend on how slow, exactly
[...]

PR #248, opened 2015-09-22
Fix issue 247: detect brs in text to create readerable content

gijsk:
@leibovic How does this look for #247 ? I've not really done a serious performance test, and the messing with the array is a bit icky... but otherwise I think this should do? I picked 5 somewhat arbitrary (see the issue IRC log), but that seems like an OK start to me, and we can tweak if necessary...

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1703,6 +1703,17 @@ Readability.prototype = {
    */
   isProbablyReaderable: function(helperIsVisible) {
     var nodes = this._getAllNodesWithTag(this._doc, ["p", "pre"]);
+    if (this._doc.querySelectorAll && Array.from) {
+      nodes = Array.from(nodes);
+      var brs = this._doc.querySelectorAll("br:nth-of-type(5)");
+      for (var i = 0; i < brs.length; i++) {
+        var parentNodeName = brs[i].parentNode.nodeName.toLowerCase();
+        if (parentNodeName != "p" && parentNodeName != "pre" &&
+            nodes.indexOf(brs[i].parentNode) == -1) {
+          nodes.push(brs[i].parentNode);
+        }
+      }
+    }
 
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -1703,6 +1703,17 @@ Readability.prototype = {
    */
   isProbablyReaderable: function(helperIsVisible) {
     var nodes = this._getAllNodesWithTag(this._doc, ["p", "pre"]);
+    if (this._doc.querySelectorAll && Array.from) {
+      nodes = Array.from(nodes);
+      var brs = this._doc.querySelectorAll("br:nth-of-type(5)");
```
leibovic: I would at least leave a comment explaining that 5 is a bit of an arbitrary value here.

Readability.js
```diff
@@ -1703,6 +1703,17 @@ Readability.prototype = {
...
+      for (var i = 0; i < brs.length; i++) {
```
leibovic: Should we limit this to more than just brs.length? What if we have a page that has a ton of brs?

leibovic:
This approach seems sane, but I do worry a bit about performance. Do we have any performance tests in place? I know we have telemetry histogram probes for parse times, but I wonder if it would be worth landing one for the `isProbablyReaderable` time as well.

gijsk:
Fair points. The code is simpler this way, but it would probably be more performant if we avoided getting the nodelist length (as well as converting the one for the paragraphs into an array). I can look at that.

We do have performance tests for the commandline stuff but that won't work with this selector, I'm pretty sure, so I'll have to do some manual testing to get a better idea of what the impact is.

salty-horse:
Bug #247 appears fixed in Firefox 52.0.1 - should this be closed?

gijsk:
Yeah, closing makes sense at this point. Thanks!

Issue #251, opened 2015-10-16
iab.com article has Reader View button but displays "Failed to load article from page" error


cpeterso:
STR:
0. I am using Firefox Nightly 44 on OS X 10.11.
1. Load http://www.iab.com/news/lean/ ("Getting LEAN with Digital Ad UX" article)
2. Click the Reader View button.

RESULT:
The Reader View page is blank except for a "Failed to load article from page" error

gijsk:
This works when the document is requested via XHR (so if you close the "failed" tab, and then reopen the tab, it works). I expect the ads are somehow breaking things. I don't have time to investigate why right now.

gijsk:
This is working on fx-team tip which has the changes from #252, so I'm closing this. :-)

PR #252, opened 2015-10-19
Fix issue #251 by making JSDOMParser deal with non-self-closed-things

gijsk:
This has been a thing for a while now, and is the cause of issue #251 and likely other cases where we fail to parse the XML/HTML we get from gecko/webkit/whatever.

This worked with everything we had in our test-suite already so I'm reasonably confident this doesn't break anything else in particular. The other thing I can think of is that we could get rid of the self-closing tag malarkey entirely - I don't think our XML serializer will ever send us self-closed things without a `/>` to end with, because it's an XML serializer... unless it treats SVG separately? In any case, it also didn't really seem worth the risk to rely on that rather than fixing it (though the fix ended up being a bit more involved than I'd hoped).

I added the page from #251 as a testcase, and wrote a minimal testcase for the original issue. I also improved our error reporting a little bit, so that hopefully in future it's easier to debug stuff like this.


gijsk:
@leibovic Margaret, can you have a look at this? :-)

leibovic:
I looked this over, but I'm not sure if I'm the best reviewer here... I believe @thebnich originally wrote this, maybe he can take a look? Otherwise I'll have to spend a bit more time here to re-familiarize myself with what's going on.

gijsk:
> I could use a bit more background on why this fix is needed; I thought that we already did assume that all void tags were self-closed.

We do, but that assumption is broken. If I have a tree:
```
<input><p></p></input>
```
We currently assume that because `input` is in the `voidElems` array, there are never children. JSDOMParser checks if the next character is `/` and then checks for `>` -- but if the first character isn't `/` it just assumes that the element is closed anyway. That is wrong.

When we imported JSOMParser into this tree, that was already the case:

https://github.com/mozilla/readability/commit/cd89e186473876377f1bb1ee85545d931a738245#diff-34fa83b6e031ec8ff17b65b3483f5367R752

We return the tag as closed whether or not there is a closing `/>` present. Looking at blame in hg.mozilla.org, it seems this approach has been there since the dawn of time (ie since you fixed https://bugzilla.mozilla.org/show_bug.cgi?id=779796 ).

We could start assuming now that we never need to set closed to true for `voidElems` and just let the detection of the closing tag take care of this, but I don't know that that is a safe assumption. In particular, I don't know what iOS and other consumers of this code do in terms of XML-based sanitization. The patch in this PR is more conservative from that perspective, but we could equally take the simpler approach here and see what breaks, if you prefer.

Thoughts? :-)

thebnich:
Thanks for the detailed explanation!

[iOS also uses `XMLSerializer`](https://github.com/mozilla/firefox-ios/blob/b5a9b27276ec86ff1f6ba238b8c3d84cad71d2f6/Client/Frontend/Reader/ReaderMode.js#L47), and I imagine all other consumers do too (since `XMLSerializer` fixes all sorts of syntax errors that we don't handle as part of `JSDOMParser`). Given the current state of the repository, I'd vote for the auto-self-closing approach for a couple of reasons:
 1. `JSDOMParser` is pretty complicated, so the simpler option (assuming it works just as well) should reduce the potential for bugs and lower the long-term maintenance burden, and
 2.  `XMLSerializer` does a ton of syntax fixup as mentioned above, and I'm pretty sure we'd be quite broken without it (unless a bunch of robustness handling was added to the parser that I'm unaware of). Having this "proper" fix is fruitless, then, since we need to go through the serializer step either way.

That said, if we have hopes to eventually decouple this from the `XMLSerializer`, I'm fine with this PR, too. I'm barely involved with this code nowadays, so I'll leave that decision to you :)

gijsk:
Hmm, so I just looked at this again (finally!) and it seems like our test suite really doesn't like this, because `jsdom` generates HTML-like `<br>` and `<meta>` and so forth, and `JSDOMParser` doesn't.

That and all the test input uses HTML rather than XHTML so that needs updating, which is boring but doable... I'll have a look if I can just make the tests more robust here.

gijsk:
Right. This should do the trick, I believe. @thebnich , could you take a look at `test-readability.js` and `JSDOMParser.js` (I mean, you're welcome to look at the test changes, but they're a bit redundant...)

thebnich:
Looks fine to me!

PR #254, opened 2015-10-21
Readme change to fix scrambling of DOM on parse(). Fixes #250

### LINKED ISSUE #250: The parse function scrambles the document
When I run 
`var article = new Readability(uri, document).parse();`

the parse() function scrambles the document by removing various content in the page, for instance the main article, images. 

Screenshots:
Before:
before readability
After:
after readability

From what I understand, readability works by removing these elements from the DOM.
Is there a way to avoid scrambling the page apart from simply reloading ?

hsemarap:


gijsk:
I'd rather add a separate subsection to the readme inside the "Usage" section that explains this, rather than commented out code in the example code.

It could explain that the DOM of the document you pass into Readability will be modified (which is a clearer word than "scrambled", I think) by Readability, and that if you want to avoid this, you should pass in a clone of the original document. It could then list the two lines you propose here, not commented out.

How does that sound?

gijsk:
Woops, sorry for the delay - why did github not send me email for the updated PR? Oh well - going to merge this, thanks!

PR #255, opened 2015-11-10
Create a "NOT-Parse Code" <=> Duplicate Content (Copy) - Big Issue for Webmasters...

lancomega:
Dear community,

I am the owner of the site http://mr-ginseng.com , with thousands of visitors every day. I spend a lots of time doing articles (more than days for only one like http://mr-ginseng.com/ortie/ , so I try to protect my content as much as I can. 

However, since the release of the "Reader Mode", it's a huge issue, because users can easily copy/paste the content. As a result, I found plenty of my articles duplicate again, sometimes even before my website in search results. When you spend hours, doing your website, you can imagine my reaction... 

Speaking with other webmasters and writers, we are a lot's to have this problem, and nobody can't find any solution... So I decided to Pull a request here, hoping you will find a solution to fixe this.

=> What we would like is a "specific code" that can be adding on our pages and articles (wordpress and others CMS..), avoiding the parse of our pages in Firefox and other search engines. So at the end, the page will be considere as NOT parse and readable at all. So no Read mode at the end.

Really hope you will fixe this, as this is a really big issue for webmasters!
Best regards


gijsk:
Hi,

> However, since the release of the "Reader Mode", it's a huge issue, because users can easily
> copy/paste the content. As a result, I found plenty of my articles duplicate again, sometimes even
> before my website in search results. When you spend hours, doing your website, you can imagine
> my reaction...

I very much doubt that this is the result of reader mode. This can be achieved through a variety of means, including but not limited to:
* Google/Bing/Yahoo's cache of the page
* Removing all page styles (in Firefox: View > Page Style > No Style)
* Depending on how you're trying to prevent copy/paste, by using an add-on like NoScript or otherwise disabling scripts
* Saving the page as plain HTML without the CSS/JS applied (File > Save As...)
* Viewing the webpage source and copying from there (Tools > Web Developer > Page Source)
* Fetching the content with something like `wget` or `curl` and converting the HTML output to text with one of the many tools available to do that.

It is a given feature of the web that textual content online can easily be copied and re-posted elsewhere (with or without modifications). There is fundamentally no real way around this.

> Speaking with other webmasters and writers, we are a lot's to have this problem, and nobody can't
> find any solution... So I decided to Pull a request here, hoping you will find a solution to fixe this.

Pull requests are for adding specific code to a project. You meant to open an issue. In the end it does not really matter because there is nothing we can really do about the effect of reader mode, and even if we added some kind of `<meta>` tag or HTTP header that would let pages "opt out" of reader mode, that would not stop any of the myriad other ways listed above.

> => What we would like is a "specific code" that can be adding on our pages and articles (wordpress
> and others CMS..), avoiding the parse of our pages in Firefox and other search engines.

Firefox is not a search engine, it is a web browser. We cannot make Google or other search engines do what you want, nor can we make Safari or other browsers with a "reader mode" do what you want.

We do not wish to add the feature you're requesting, because it will be ineffective at achieving your aim, and will hinder the usability of the web. For instance, the pages you link would be hard to read by users who struggle with the colours, and reader mode is one of the tools they can use to make such pages readable. It would be wrong to forego the benefits of the feature in order to attempt to protect your pages, an attempt that is sure to fail.

> So at the end, the page will be considere as NOT parse and readable at all. So no Read mode at the end.

We will not do this.

I would suggest that if the pages that you see copied are under your copyright, and you feel that you suffer financial or personal loss as a result, that the appropriate method to seek redress would be to file suit in the appropriate court. That approach is as open to you now as it would be if they manually re-wrote your articles by looking at the screen and writing down what they saw - something which neither browsers nor search engines can prevent them from doing.

lancomega:
Thanks for your reply Gijsk. (ps: I meant not search engines but browsers).
I am working in my websites since 10 years, and I can confim you that most of my duplicate come from the release of READ MODE in Firefox, because everything else has been secure by some wordpress plugins. Moreover duplicate contents are often paste with the same CSS and lines break of the Read Mode from firefox. I saw that the homepage is not parse. What I was hoping is just a keyword or an advice, that can help me not to have my page in read mode?!

Best regards.

PR #258, opened 2015-12-08
set direction to rtl in RTL articles

djug:
if the article direction is not specified, set it to rtl if the title of the page is written in an RTL language.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1696,6 +1696,15 @@ Readability.prototype = {
     this._flags = this._flags & ~flag;
   },
 
+  _isRtl: function(str){           
+      var ltrChars    = 'A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF'+'\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF',
+          rtlChars    = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC',
+          rtlDirCheck = new RegExp('^[^'+ltrChars+']*['+rtlChars+']');
+
+      return rtlDirCheck.test(str);
+  },
+  
+
   /**
    * Decides whether or not the document is reader-able without parsing the whole thing.
    *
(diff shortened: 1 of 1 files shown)


djug:
Currently some Arabic articles (that do not specify the direction) do not show up correctly


here is an example:

http://aljazeera.net/knowledgegate/opinions/2015/12/8/-%D8%A7%D9%84%D8%AC%D9%87%D8%A7%D8%AF%D9%8A%D9%88%D9%86-%D9%81%D9%8A-%D8%A7%D9%84%D8%B5%D8%AD%D8%B1%D8%A7%D8%A1-%D9%88%D8%B1%D8%B3%D8%A7%D8%A6%D9%84-%D8%A7%D9%84%D8%B9%D9%88%D8%AF%D8%A9
ltr-arabic-text

Readability.js
```diff
@@ -1696,6 +1696,15 @@ Readability.prototype = {
     this._flags = this._flags & ~flag;
   },
 
+  _isRtl: function(str){           
+      var ltrChars    = 'A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF'+'\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF',
+          rtlChars    = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC',
+          rtlDirCheck = new RegExp('^[^'+ltrChars+']*['+rtlChars+']');
```
gijsk: Can you add a comment about where this logic comes from and why it's necessary?

gijsk: In fact, this regexp sees odd - the string should start with any number of non-LTR characters and then have 1 RTL character?

Readability.js
```diff
@@ -1806,6 +1815,9 @@ Readability.prototype = {
         metadata.excerpt = paragraphs[0].textContent.trim();
       }
     }
+    if (this._isRtl(articleTitle)){
```
gijsk: This doesn't do what you said in the description of the PR - it always overrides the article direction, even if the direction *was* specified. Please check whether `_articleDir` is specified.

gijsk:
It's been a year and this PR would need substantial changes in order to be merged. Separately, we have made improvements in how we gather RTL-ness in other ways, so I am going to close this PR. If there are other improvements necessary, they should be made in separate PRs.

PR #260, opened 2016-01-12
Fix bug 1230050 by checking for the 'hid' class specifically, r?MattN

gijsk:
Matt ( @mnoorenberghe ), does this look right to you? I realize I could use nested `(?: |$)` but that seemed likely to be slower and harder to grok than just this...

The reason that we don't detect this is that we don't use the visibility helper for `parse`, it seems... only for `isProbablyReaderable`. We can fix that separately.


mnoorenberghe:
r+. WFM.

Issue #267, opened 2016-03-05
Readability does not treat same-page anchor links correctly


akashagrahari:
For webpages which have anchor points to subheading, readability alters the link which shows up in the final view.

For example - for the [link](http://jnb.ociweb.com/jnb/jnbJun2009.html)
[Thrift Architecture](http://jnb.ociweb.com/jnb/jnbJun2009.html#arch) becomes [Thrift Architecture](http://jnb.ociweb.com/jnb/#arch)

Screenshots
<img width="945" alt="screen shot 2016-03-05 at 6 31 09 pm" src="https://cloud.githubusercontent.com/assets/2344230/13547634/8e5902b4-e300-11e5-9217-3edfc07b2198.png">
<img width="815" alt="screen shot 2016-03-05 at 6 31 21 pm" src="https://cloud.githubusercontent.com/assets/2344230/13547635/8e5ae1ec-e300-11e5-8bcb-ea09a297198f.png">

gijsk:
Hm, odd. I can reproduce in the browser, but not with the base readability code in here. I tried modifying a builtin testcase, and it treats simple `href="#foo"` links correctly. I wonder what's broken...

gijsk:
Oh, after my test refactoring to use DOM methods we're not checking attributes, so that's why my testcase modification isn't catching that. That's silly of me. Anyway, now I have a failing test, just need to fix the issue now...

gijsk:
Woops, forgot to close this - but it's generating the correct markup now. Of course, now it seems we're still opening `#hash` links in Firefox against the original page in a new tab, which is only marginally better. But not a bug in readability.

gijsk:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1280888 about the issue in Firefox.

PR #268, opened 2016-03-07
Fix issue #267 by ignoring hash URIs when making URIs absolute

gijsk:
@leibovic can you take a look? Should be a pretty trivial review - turns out there were some existing testcases where we mangled these URIs, but the guts of the change are visible through just looking at Readability.js, the base-url testcase, and the changes to test-readability.js.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -192,7 +192,8 @@ Readability.prototype = {
   },
 
   /**
-   * Converts each <a> and <img> uri in the given element to an absolute URI.
+   * Converts each <a> and <img> uri in the given element to an absolute URI,
+   * ignoring #ref URIs.
    *
    * @param Element
    * @return void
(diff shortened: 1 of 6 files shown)


leibovic:
Good catch, this looks reasonable to me.

leibovic:
@gijsk Will you make a patch to update Readability.js in m-c? What's our process been for updating that recently?

gijsk:
> @gijsk Will you make a patch to update Readability.js in m-c? What's our process been for updating that recently?
Yeah, I'll just uplift to m-c when I find a moment. I don't think we have a particularly solid process - there haven't been that many m-c-related changes in the last year or thereabouts. m-c already has everything before this change, I believe?

gijsk:
Heh, and now it turns out some folks changed readability in m-c because of eslint (bug 1245649). We might need more process. :-\

rnewman:
Hurgh.

PR #270, opened 2016-03-08
Fix eslint issues identified in m-c

gijsk:
@leibovic these are some eslint-related compat changes. I've added an eslintrc file (copied from toolkit, removed es6 and a few other things) and added  so that hopefully makes sure we won't have "broken" es6 code put in these files when people are doing eslint stuff in all of toolkit/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,194 @@
+{
+  "rules": {
+    // Braces only needed for multi-line arrow function blocks
+    // "arrow-body-style": [2, "as-needed"],
+
+    // Require spacing around =>
+    // "arrow-spacing": 2,
+
+    // Always require spacing around a single line block
+    // "block-spacing": 1,
+
+    // No newline before open brace for a block
+    // "brace-style": 2,
+
+    // No space before always a space after a comma
+    // "comma-spacing": [2, {"before": false, "after": true}],
+
+    // Commas at the end of the line not the start
+    // "comma-style": 2,
+
+    // Don't require spaces around computed properties
+    // "computed-property-spacing": [2, "never"],
+
+    // Functions must always return something or nothing
+    "consistent-return": 2,
+
+    // Require braces around blocks that start a new line
+    // Note that this rule is likely to be overridden on a per-directory basis
+    // very frequently.
+    // "curly": [2, "multi-line"],
+
+    // Always require a trailing EOL
+    "eol-last": 2,
+
+    // Require function* name()
+    // "generator-star-spacing": [2, {"before": false, "after": true}],
+
+    // Two space indent
+    // "indent": [2, 2, { "SwitchCase": 1 }],
+
+    // Space after colon not before in property declarations
+    // "key-spacing": [2, { "beforeColon": false, "afterColon": true, "mode": "minimum" }],
+
+    // Unix linebreaks
+    "linebreak-style": [2, "unix"],
+
+    // Always require parenthesis for new calls
+    // "new-parens": 2,
+
+    // Use [] instead of Array()
+    // "no-array-constructor": 2,
+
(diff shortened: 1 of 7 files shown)


leibovic:
Can we add a rule for this in our .travis.yml file, such that the build will break if someone violates these rules?

leibovic:
But otherwise, this looks fine to me, thanks for resolving this m-c issue to keep these things in sync.

gijsk:
> Can we add a rule for this in our .travis.yml file, such that the build will break if someone violates these rules?

I would love to, but I have no idea how to do that...

leibovic:
> > Can we add a rule for this in our .travis.yml file, such that the build will break if someone violates these rules?

> I would love to, but I have no idea how to do that...

We're like the blind leading the blind! Doing some quick searching online, I think we may be able to do this by adding an npm script for eslint in our `package.json`, then add somethign like `script: npm run eslint` to `.travis.yml`. Or we may need to add a test that runs eslint itself and fails if eslint fails.

I would have thought there would be a more obvious way to do this, given this seems like a common developer workflow.

pdehaan:
As long as you're installing eslint in package.json, the easiest way would be to make a new npm `script` in package.json saying `"lint": "eslint ."`, and then in .travis.yml  we can add a script to run `npm run test` and `npm run lint`. You can see a rough idea over at https://github.com/mozilla/activity-streams/blob/master/.travis.yml

We may also need to add an .eslintignore if there are files we DONT want linted (for example, anything in benchmarks/ or test/). We can also make a nested .eslintrc file if you want to override ESLint rules for specific directories.

pdehaan:
Or, you can merge this as-is, and I can do a follow-on PR to add ESLint to the .travis.yml, if you prefer.

test/generate-testcase.js
```diff
@@ -17,7 +17,7 @@ var FFX_UA = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:38.0) Gecko/20100
 if (process.argv.length < 3) {
   console.error("Need at least a destination slug and potentially a URL (if the slug doesn't have source).");
   process.exit(0);
-  return;
+  throw "Abort";
```
pdehaan: Do we want to throw a String here, or just throw a standard `throw new Error("Abort")`?

.eslintrc
```diff
@@ -0,0 +1,194 @@
...
+    // ++ and -- should not need spacing
+    // "space-unary-ops": [2, { "words": true, "nonwords": false }],
+
+    // No comparisons to NaN
+    "use-isnan": 2,
+
+    // Only check typeof against valid results
+    "valid-typeof": 2,
+  },
+}
```
pdehaan: Not sure if it'd be easier to just delete all the unused/commented rules so we don't have to find the ~25 rules that ARE enabled.

gijsk: I'd prefer to be able to easily incorporate changes from m-c once they land there. As it is this is just a copy of toolkit's eslint.

test/generate-testcase.js
gijsk: I mean, it doesn't really matter considering we're calling `process.exit` on the previous line and this is just a utility script, but return was technically illegal, so I literally wrote the shortest thing I could think of.

gijsk:
> We may also need to add an .eslintignore if there are files we DONT want linted (for example, anything in benchmarks/ or test/). We can also make a nested .eslintrc file if you want to override ESLint rules for specific directories.

Why wouldn't we want to lint the other files?

pdehaan:
> Why wouldn't we want to lint the other files?

No clue, I wasn't sure what the benchmarks/ stuff was. I was just saying it was an option (and maybe not a good option).

PR #274, opened 2016-03-15
Convert nested function declaration to function expression

### LINKED ISSUE #273: SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function
I just casually noticed this while trolling through some verbose logs on ActivityStream...

> console.log: activity-streams: [JavaScript Warning: "SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function" {file: "resource://gre/modules/ReaderMode.jsm -> resource://gre/modules/reader/Readability.js" line: 69 column: 13 source: "    function logEl(e) {
> "}]

It looks like this is coming from the following code, [./Readability.js:63](https://github.com/mozilla/readability/blob/edfe703a418baa2d03e4200b8efb891f087f3df3/Readability.js#L63) which seems to be trying to define a function within an `if` block. I'm guessing that strict mode prefers to not do that, as the function would probably be hoisted and possibly lead to some confusing behavior.

See:
1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Paving_the_way_for_future_ECMAScript_versions
2. http://whereswalden.com/2011/01/24/new-es5-strict-mode-requirement-function-statements-not-at-top-level-of-a-program-or-function-are-prohibited/

pdehaan:
Fixes #273

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -58,9 +58,11 @@ var Readability = function(uri, doc, options) {
   // Make an AJAX request for each page and append it to the document.
   this._curPageNum = 1;
 
+  var logEl;
+
   // Control whether log messages are sent to the console
   if (this._debug) {
-    function logEl(e) {
+    logEl = function(e) {
       var rv = e.nodeName + " ";
       if (e.nodeType == e.TEXT_NODE) {
         return rv + '("' + e.textContent + '")';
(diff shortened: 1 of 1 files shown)


gijsk:
Hm, my email client is lagging, or I would have gotten to this sooner - anyway, thanks!

Issue #271, opened 2016-03-10
Reader View not displaying BBC news articles as expected (only on iOS)


krudnitski:
Filed a [bug in bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1255357), but also posting the issue here. 

Running Firefox for iOS Beta (3.0). 

Go to: bbc.co.uk/news/uk-politics-35762443

The header image is very large, whilst the header text (title text) and article text are much smaller as a result. This actually looks like a regression. It behaves as expected on version 2.1 (14). 

Firefox for Android behaves better and Safari behaves really nicely.

![img_4136](https://cloud.githubusercontent.com/assets/5297048/13667740/dfdbf738-e6af-11e5-9d40-18fa68b7f0ef.PNG)

gijsk:
If this works on android + desktop then I suspect it's an iOS CSS-for-reader-mode, or potentially a webkit-CSS-implementation-related issue. I don't have an iDevice and don't really know much about how iOS uses Readability right now... @st3fan or @rnewman, can you help?

gijsk:
Oh, that or the BBC sends us 'special' markup for iOS that causes this issue... which would be another 'fun' possibility, but it'd need someone to look at what's actually going wrong on their iDevice.

st3fan:
http://www.bbc.com/news/world-europe-35772206 looks pretty much the same as in Safari for me. 

Same with http://www.bbc.com/news/world-europe-35771292

krudnitski:
With bbc.co.uk/news/businses-35770754, same problem. On Fx iOS Beta 3.0, it formats like my previous example. But on Safari, it looks great (see attached)
![img_4137_fxiosbb](https://cloud.githubusercontent.com/assets/5297048/13671247/b2ac83e2-e6c6-11e5-8200-af44460bbcdd.PNG)
![img_4138_safari](https://cloud.githubusercontent.com/assets/5297048/13671248/b2ae625c-e6c6-11e5-812f-f80eae19429f.PNG)

st3fan:
I don’t see this when I enable Reader Mode for that specific article. For me it looks pretty much similar to Safari.

Can you try the following: change your phone to landscape and back and see if that helps.

Try to change the Reader Mode settings (font size) and see if that makes a difference.

Does it happen for other articles on BBC News too? Or just this one?

 S.

st3fan:
I see it too now …

Karen, do you run a Content Blocker? Like Focus?

[...]

PR #282, opened 2016-03-31
Fix #280 and #271 by being more careful about how we iterate over node lists in case they're live

### LINKED ISSUE #280: calls that iterate over nodes with HTMLCollection should not mutate the collection
We (@chiua and I) noticed that only half of the scripts in our test page were being removed.

The issue is that getElementsByTagName returns an HTMLCollection, and when you iterate over it, and the loop mutates it, the iterator gets confused.  In our case there were 38 scripts and only 19 (every other) were removed.

Prime example:
https://github.com/mozilla/readability/blob/master/Readability.js#L1036-L1047
```
_removeScripts: function(doc) {
    this._forEachNode(doc.getElementsByTagName('script'), function(scriptNode) {
      scriptNode.nodeValue = "";
      scriptNode.removeAttribute('src');

      if (scriptNode.parentNode)
        scriptNode.parentNode.removeChild(scriptNode);
    });
```

We saw this on Chrome but tested on Firefox and saw the same thing.

It seems like _lots_ of code iterates over elements this way, and it might cause problems in other places.

We fixed it by updating `_forEachNode` to copy the HTMLCollection into a regular array.  Glad to put up a PR for this, but we figured someone might have another solution in-progress.

gijsk:
This is a plain JS fix without copying arrays. The performance aspects of this in the JSDOM/JSDOMParser case are fine, at least.

However... for those of us on iOS or other consumers that use a "real" DOM implementation, I actually suspect that creating a 'real' Array copy of the list might be better - there's good evidence that iterating over live NodeLists is very slow, because they have to be updated every time someone touches the DOM.

So we can take this, and it will be fine for iOS and Android... but I wonder if we should instead just check whether:

```
Array.isArray(nodeList)
```
and if not, use `Array.from` to create one. Then we always have a static array, and can just use `arrayInstance.forEach` directly.

It's not possible for me to benchmark this easily with representative examples on iOS / a "real" DOM implementation.

@leibovic, @st3fan, what do you think and/or can you help with the benchmark problem? :-)

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -155,7 +155,15 @@ Readability.prototype = {
    * @return void
    */
   _forEachNode: function(nodeList, fn) {
-    return Array.prototype.forEach.call(nodeList, fn, this);
+    var i = 0;
+    while (nodeList[i]) {
+      var el = nodeList[i];
+      fn.call(this, el, i, nodeList);
+      // Only increment i if the element is still in the list:
+      if (nodeList[i] == el) {
+        i++
+      }
+    }
   },
 
   /**


Readability.js
```diff
@@ -155,7 +155,15 @@ Readability.prototype = {
...
-    return Array.prototype.forEach.call(nodeList, fn, this);
+    var i = 0;
+    while (nodeList[i]) {
+      var el = nodeList[i];
+      fn.call(this, el, i, nodeList);
+      // Only increment i if the element is still in the list:
+      if (nodeList[i] == el) {
+        i++
+      }
+    }
```
persidskiy: The only thing that I don't like here is fact that if you delete any node, next node will have the same index as previous. It doesn't match to the original behaviour of `forEach` and may be unobvious for users in the future.

persidskiy: Ah, and one another thing: If nodeList shifts all the next nodes after deleting one, the complexity will be O(n<sup>2<sup>), which is not good.

gijsk: Fair. Let's see if we can hammer out something on #300 instead, then.

Issue #283, opened 2016-04-02
Shouldn't show display:none elements


loganrosen:
I opened the Reader View for [the Omer Adam Wikipedia article](https://en.wikipedia.org/w/index.php?title=Omer_Adam&oldid=712393315) in Firefox, and the first thing on the page below the title was:
> HIDDEN ERROR: Usage of "Img capt" is not recognized

This doesn't even show up in the article itself because it is a span with an inline style of display:none, as you can see here (in case the template in the article changes or something):

`<p><span class="error" style="display:none">HIDDEN ERROR: Usage of "Img capt" is not recognized</span></p>`

It probably shouldn't show up in Reader View either.

PR #285, opened 2016-04-11
Fix issue #283 and remove individually hidden nodes

gijsk:
@leibovic This not only fixes the issue, but also tidies up some of the iteration code now that `jsdom` ships `firstElementChild`, `nextElementSibling` etc., as well as a reasonable implementation of `document.style`, so this effectively supersedes the older #186 .


gijsk:
> Hi there,
> Why this is not merged yet?

It got lost. At this point it needs considerable rebasing to apply to today's codebase, and then it'd need a review. I don't know when I'll have time to do this myself, but I'll try to find some in the next week or two.

minas90:
Hi Gijs,
Great, thanks for reply.
I did it myself and found out that there is a pull request already.
This change improves parsing performance for some websites. 
Here is an example:
[https://www.segodnya.ua/world/wnews/v-kitae-passazhir-samoleta-nabrosilsya-na-bortprovodnika-s-sharikovoy-ruchkoy-1130846.html](url)

Search for `<div itemscope` in source.
This div contains some metadata and has inline style `"display: none;"`, but readability removes the inline style and reveals it.
Also I have written manual rules for parsing more than 100 websites and none of them has issue with content being hidden by `"display: none;"`, so I think it is safe thing to do for most of the websites. Some websites even have duplicate "mobile version" of the same content, which is being toggled to be visible on mobile and main content is set to be gone.

Regards,
Minas

bittlingmayer:
Maybe it will be easier to move this forward if we break this down, so first just do a small change like this:

https://github.com/FeedScaleInfra/readability/commit/b2d05f1425f1ba89c3aa3e4ccb43e304245b8ccf

And then do all the re-factoring in later changes?

gijsk:
FWIW, I think we solve the image size thing in Firefox on Android by using our own CSS with the markup that readability generates. This may also work for your project, depending on what you're doing? See e.g. https://dxr.mozilla.org/mozilla-central/source/mobile/android/themes/core/aboutReader.css#458-475

gijsk:
I've (finally! sorry...) updated the PR. Some existing tests needed adjusting because they included hidden content. I also added a unit-y test. @msujaws can you doublecheck this looks OK to you? Thanks!

PR #288, opened 2016-04-19
Fix #287: convert getElementsByTagName result to an array

### LINKED ISSUE #287: _getAllNodesWithTag() method fallback returns HTMLCollection array
It seems that _getAllNodesWithTag() uses node.querySelectorAll() for browsers that support it, which I'd assume are all at this point.

The fallback use of node.getElementsByTagName() returns an array of HTMLCollection items, vs nodes.

If moz used that fallback, the main consumer of the method isProbablyReaderable() would have issues, failing in the visibilityHelper:
Error: "NS_ERROR_XPC_BAD_CONVERT_JS: Could not convert JavaScript argument arg 0 [nsIDOMWindowUtils.getBoundsWithoutFlushing]" {file: "resource://gre/modules/ReaderMode.jsm" line: 140}]
ReaderMode.isNodeVisible() calls |... utils.getBoundsWithoutFlushing(node);|

If we need to keep the fallback, it seems the proper way to retrieve and flatten the node array is:

return [].concat.apply([], tagNames.map(tag => {
      return [].slice.call(node.getElementsByTagName(tag));
}));

gijsk:
@mcapella, can you check this works?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -193,7 +193,8 @@ Readability.prototype = {
       return node.querySelectorAll(tagNames.join(','));
     }
     return [].concat.apply([], tagNames.map(function(tag) {
-      return node.getElementsByTagName(tag);
+      var collection = node.getElementsByTagName(tag);
+      return Array.isArray(collection) ? collection : Array.from(collection);
     }));
   },
 


mcapella:
Yes, I see this work correctly as a fallback :)

PR #289, opened 2016-05-04
exposed textContent to be returned

usergit:
this returns the text content only, this is useful as it allows the content to be easily accessible


gijsk:
Thanks! Merged with an update to make us only compute `textContent` once.

PR #298, opened 2016-07-12
Support boolean attributes (closes #55)

### LINKED ISSUE #55: JSDOMParser doesn't handle boolean attributes
In the example below, `allowfullscreen` is what the html spec calls a [boolean attribute](http://www.w3.org/TR/REC-html40/intro/sgmltut.html#h-3.3.4.2):

``` html
<iframe width="560" height="315" 
  src="https://www.youtube-nocookie.com/embed/zOduYvDupXQ" 
  frameborder="0" 
  allowfullscreen></iframe>
```

Trying to parse this brings:

```
JSDOMParser error: expected '</iframe>'
```

Switching to

``` html
<iframe width="560" height="315" 
  src="https://www.youtube-nocookie.com/embed/zOduYvDupXQ" 
  frameborder="0" 
  allowfullscreen="allowfullscreen"></iframe>
```

makes it pass. We should probably support boolean attributes as they're widely used.

persidskiy:
Hello, the purpose is clear. Now parsing crashes if HTML has any boolean attribute. And it's very bad because you use JSDOMParser in `generate-testcase` script.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -888,28 +888,29 @@
      */
     readAttribute: function (node) {
       var name = "";
-
-      var n = this.html.indexOf("=", this.currentChar);
-      if (n === -1) {
-        this.currentChar = this.html.length;
-      } else {
-        // Read until a '=' character is hit; this will be the attribute key
-        name = this.html.substring(this.currentChar, n);
-        this.currentChar = n + 1;
+      var c;
+      while((c = this.nextChar()) && /[\w-]/i.test(c)) {
+        name += c;
       }
 
       if (!name)
         return;
 
-      // After a '=', we should see a '"' for the attribute value
-      var c = this.nextChar();
-      if (c !== '"' && c !== "'") {
-        this.error("Error reading attribute " + name + ", expecting '\"'");
-        return;
-      }
+      var value;
+      if(c === "=") {
+        // After a '=', we should see a '"' for the attribute value
+        c = this.nextChar();
+        if (c !== '"' && c !== "'") {
+          this.error("Error reading attribute " + name + ", expecting '\"'");
+          return;
+        }
 
-      // Read the attribute value (and consume the matching quote)
-      var value = this.readString(c);
+        // Read the attribute value (and consume the matching quote)
+        value = this.readString(c);
+      } else {
+        value = "";
+        this.currentChar--;
+      }
 
       node.attributes.push(new Attribute(name, value));
 
--- a/test/test-jsdomparser.js
+++ b/test/test-jsdomparser.js
@@ -10,7 +10,7 @@ var JSDOMParser = readability.JSDOMParser;
 
 var BASETESTCASE = '<html><body><p>Some text and <a class="someclass" href="#">a link</a></p>' +
                    '<div id="foo">With a <script>With < fancy " characters in it because' +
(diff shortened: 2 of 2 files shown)


gijsk:
JSDOMParser assumes XML-valid (XHTML) input. Boolean attributes with no equals sign aren't, so I don't think this patch makes sense.

What are you actually trying to do?

persidskiy:
I support some my own test cases in private fork. The problem is when I make the test case of the page which has any boolean attributes and try to run tests against it, the tests via JSDOMParser fail, because JSDOMParser couldn't understand boolean attributes. It forces me go to the `source.html` and remove boolean attributes by hand. 

At the end, support of boolean attributes is very cheap and I don't see any significant reasons why it shouldn't be implemented.

gijsk:
> At the end, support of boolean attributes is very cheap and I don't see any significant reasons why it shouldn't be implemented.

In node, running `npm run-script perf-reference` indicates approximately 35% perf regression with this patch (previously ~170 ops/s on my machine, now 125, which means runtime went from ~5.9ms per run to ~8ms per run, so 35% slower).

It seems like an easier fix for the issue with your private fork would be to make it an option to not run tests against JSDOMParser if you don't care about JSDOMParser. I would take a patch to do that.

gijsk:
(I'll also add that the attempt to somewhat/kind of deal with HTML but not really has bitten us in the past, and so I'm not convinced the added complexity is really worth it *anyway* - see e.g. https://github.com/mozilla/readability/pull/252 )

persidskiy:
> In node, running npm run-script perf-reference indicates approximately 35% perf regression with this patch (previously ~170 ops/s on my machine, now 125, which means runtime went from ~5.9ms per run to ~8ms per run, so 35% slower).

Hm, on my machine (MBP i7, node v4.4.7) the results are 244 ops/s (this patch) vs 224 ops/s (master). I've run `npm run-script perf-reference` 10 times per branch and took arithmetic mean of `» JSDOMParser test page perf` results (all results are [here](https://docs.google.com/spreadsheets/d/1Y5Y8L_hhOcamn8pzfUy41zN7QFip8As_iIpFZQAighY/edit?usp=sharing)). What do I wrong?

> It seems like an easier fix for the issue with your private fork would be to make it an option to not run tests against JSDOMParser if you don't care about JSDOMParser. I would take a patch to do that.

I'm experimenting with JSDOMParser in production, so it would be great for me to make it better :)

gijsk:
> Hm, on my machine (MBP i7, node v4.4.7) the results are 244 ops/s (this patch) vs 224 ops/s (master)

It does not make any sense that a `while()` loop involving a regexp check and manual string concatenation would ever be *faster* than an `indexOf` call. I tested with node v5.3.0 on an MBP i7.

> I'm experimenting with JSDOMParser in production, so it would be great for me to make it better :)

I don't think that's a good idea. There are lots of other HTML-without-XML cases (self-closing elements like `` that don't need/want a closing tag, automatically closing block-level elements like `<p>` on the occurrence of the next element that can't be a child, unquoted attribute values which are legal in HTML5, ...) that it does not deal with correctly. I'd strongly recommend you either use jsdom or find a way to convert input to be XML-compliant via pre-processing.

Mozilla uses readability with XML-compliant input in all our usecases (Firefox on desktop, android and iOS) and so there's no real reason for us to invest in making it better with non-XML-compliant input, and there are both performance and correctness risks to trying to implement this. We don't have the resources to justify taking those risks.

If you want to maintain this in your fork, that's obviously up to you, but I don't think we should take this fix.

PR #300, opened 2016-07-15
Use backward iteration for removing nodes

persidskiy:
Hello guys,
The script often iterates over nodes found by `getElementsByTagName()` and deletes some or all of them. This produces an interesting bug in webkit-based browsers (not in firefox desktop, but in firefox for ios) when not all nodes iterates and therefore not all nodes deletes. For example, when the script invokes `this._clean(doc, "iframe")` on [this page](http://www.aif.ru/sport/olymp/reshenie_ob_uchastii_rossiyskih_legkoatletov_v_oi_cas_obyavit_do_21_iyulya), not all the frames  are deleted. In my experiments 65 of 126 iframes weren't deleted. Actually there is a lot of bugs related to this issue (some `<br>`s, `<script>`s, even extra paragraphs are not cut).

It happens because `getElementsByTagName()` on webkit returns NodeList which mutates during iteration, on the other hand Firefox Desktop returns [HTMLCollection](https://bugzilla.mozilla.org/show_bug.cgi?id=14869) which behaves not like `NodeList` in this situation. Thus the bug could not be reproduced on Firefox Desktop, `jsdom`, and `JSDOMParser`.

The solution here is iteration in backward direction over the NodeList. Of course, we can copy `NodeList` before iteration and iterate in forward direction, but it happens too often and I don't want O(n<sup>2</sup>) complexity instead of O(n). But in some cases where direction is important I offer to use `_getAllNodesWithTag()` which copies the list.

If problem is not clear, I can provide you some screenshots.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -144,6 +144,28 @@ Readability.prototype = {
   },
 
   /**
+   * Iterates over a NodeList, calls `filterFn` for each node and removes node
+   * if function returned `true`.
+   *
+   * If function is not passed, removes all the nodes in node list.
+   *
+   * @param NodeList nodeList The no
+   * @param Function filterFn
+   * @return void
+   */
+  _removeNodes: function(nodeList, filterFn) {
+    for (var i = nodeList.length - 1; i >= 0; i--) {
+      var node = nodeList[i];
+      var parentNode = node.parentNode;
+      if (parentNode) {
+        if(!filterFn || filterFn.call(this, node, i, nodeList)) {
+          parentNode.removeChild(node);
+        }
+      }
+    }
+  },
+
+  /**
    * Iterate over a NodeList, which doesn't natively fully implement the Array
    * interface.
    *
@@ -152,10 +174,11 @@ Readability.prototype = {
    *
    * @param  NodeList nodeList The NodeList.
    * @param  Function fn       The iterate function.
+   * @param  Boolean  backward Whether to use backward iteration.
    * @return void
    */
-  _forEachNode: function(nodeList, fn) {
-    return Array.prototype.forEach.call(nodeList, fn, this);
+  _forEachNode: function(nodeList, fn, backward) {
+      Array.prototype.forEach.call(nodeList, fn, this);
   },
 
   /**
@@ -327,9 +350,7 @@ Readability.prototype = {
     var doc = this._doc;
 
     // Remove all style tags in head
-    this._forEachNode(doc.getElementsByTagName("style"), function(styleNode) {
-      styleNode.parentNode.removeChild(styleNode);
-    });
+    this._removeNodes(doc.getElementsByTagName("style"));
 
     if (doc.body) {
       this._replaceBrs(doc.body);
@@ -363,7 +384,7 @@ Readability.prototype = {
    *   <div>foo<br>bar<p>abc</p></div>
    */
   _replaceBrs: function (elem) {
-    this._forEachNode(elem.getElementsByTagName("br"), function(br) {
+    this._forEachNode(this._getAllNodesWithTag(elem, ["br"]), function(br) {
       var next = br.nextSibling;
 
       // Whether 2 or more <br> elements have been found and replaced with a
@@ -459,7 +480,7 @@ Readability.prototype = {
     this._cleanConditionally(articleContent, "div");
 
     // Remove extra paragraphs
-    this._forEachNode(articleContent.getElementsByTagName('p'), function(paragraph) {
+    this._removeNodes(articleContent.getElementsByTagName('p'), function (paragraph) {
       var imgCount = paragraph.getElementsByTagName('img').length;
       var embedCount = paragraph.getElementsByTagName('embed').length;
       var objectCount = paragraph.getElementsByTagName('object').length;
@@ -467,11 +488,10 @@ Readability.prototype = {
       var iframeCount = paragraph.getElementsByTagName('iframe').length;
       var totalCount = imgCount + embedCount + objectCount + iframeCount;
 
-      if (totalCount === 0 && !this._getInnerText(paragraph, false))
-        paragraph.parentNode.removeChild(paragraph);
+      return totalCount === 0 && !this._getInnerText(paragraph, false);
     });
 
-    this._forEachNode(articleContent.getElementsByTagName("br"), function(br) {
+    this._forEachNode(this._getAllNodesWithTag(articleContent, ["br"]), function(br) {
       var next = this._nextElement(br.nextSibling);
       if (next && next.tagName == "P")
         br.parentNode.removeChild(br);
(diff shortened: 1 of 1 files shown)


gijsk:
Hi again! I'm aware of this issue - I submitted an alternative fix in #282 which is a lot smaller - it simply doesn't increment the iteration counter if the list changes after the deletion. Can you verify if that fix addresses all the issues you're seeing? This change seems more complex, because consumers now have to think about what to do, and as you noted, because of `JSDOMParser` we won't notice a difference on Firefox for desktop if we pick the wrong thing.

I've been waiting for a review of #282 for a while now - if it works for you I will just merge that in. Thanks!

gijsk:
Oh, I just realized there are still parts in the codebase that don't use that helper to iterate, so they won't be fixed by my changes. We should still take the parts of this PR that are addressing that. Thanks for pointing that out!

Readability.js
```diff
@@ -469,9 +476,9 @@ Readability.prototype = {
 
       if (totalCount === 0 && !this._getInnerText(paragraph, false))
         paragraph.parentNode.removeChild(paragraph);
-    });
+    }, true);
 
-    this._forEachNode(articleContent.getElementsByTagName("br"), function(br) {
+    this._forEachNode(this._getAllNodesWithTag(articleContent, ["br"]), function(br) {
```
gijsk: This removes items as well. Shouldn't this just call `_forEachNode` with the 'backward' param set to `true` and the direct result of getElementsByTagName? That seems more straightforward...

Readability.js
```diff
@@ -1627,43 +1634,43 @@ Readability.prototype = {
...
-    var tagsList = e.getElementsByTagName(tag);
-    var curTagsLength = tagsList.length;
     var isList = tag === "ul" || tag === "ol";
 
     // Gather counts for other typical elements embedded within.
     // Traverse backwards so we can remove nodes at the same time
     // without effecting the traversal.
     //
     // TODO: Consider taking into account original contentScore here.
-    for (var i = curTagsLength-1; i >= 0; i -= 1) {
```
gijsk: Does this (and the entire rest of the loop) actually need changing? I'm not sure I see why - it seems to already manually iterate backwards and do the same thing. Might make more sense to preserve blame?

Readability.js
```diff
@@ -152,10 +152,17 @@ Readability.prototype = {
    *
    * @param  NodeList nodeList The NodeList.
    * @param  Function fn       The iterate function.
+   * @param  Boolean  backward Whether to use backward iteration.
    * @return void
    */
-  _forEachNode: function(nodeList, fn) {
-    return Array.prototype.forEach.call(nodeList, fn, this);
+  _forEachNode: function(nodeList, fn, backward) {
```
gijsk: If every caller of this that uses the `backward` param is doing so because it removes items, maybe it makes more sense to make a separate function for doing this? Something like `_removeNodes(tagName, filterFn)` that always iterates backwards and removes the item if `filterFn` returns true/false? That makes it more obvious when we need to use one function and one the other. Could also rename this function something like `iterateStaticDOM` or whatever.

Readability.js
persidskiy: Yes, but algorithm of deleting `<br>`s depends on order,  because it uses `nextSiblings` and do some magic with them, so I decided to live forward direction here.

Readability.js
persidskiy: As I can see in the [first commit](https://github.com/mozilla/readability/blob/55587d91ac5b333ca6bf968e5504f3083c59cf6c/Readability.js)  all the deleting iterations were using  for loop in backward direction, but then some cycles [were changed](https://github.com/mozilla/readability/commit/4f9615cb9af8f3b27bc1448c445bb1c7c5aabd23) to use `_forEachNode` in wrong direction. The motivation for this is clear, beacause everybody love helpers) So I think we should just  introduce the "right" helper for delete nodes and I don't see any reasons why we should not do this.

Readability.js
persidskiy: Great idea! I'll make `_removeNodes`.

persidskiy:
@gijsk take a look please

Readability.js
```diff
@@ -143,6 +143,22 @@ Readability.prototype = {
     this._fixRelativeUris(articleContent);
   },
 
+  _removeNodes: function(nodeList, filterFn) {
+    for(var i = nodeList.length - 1; i >= 0; i--) {
+      var node = nodeList[i];
+      var parentNode = node.parentNode;
+      if(parentNode) {
+        if(filterFn) {
+          if(filterFn.call(this, node, i, nodeList)) {
```
gijsk: Can simplify this to just:

```
if (!filterFn || filterFn.call(this, node, i, nodeList)) {
  parentNode.removeChild(node);
}
```

Readability.js
```diff
@@ -143,6 +143,22 @@ Readability.prototype = {
     this._fixRelativeUris(articleContent);
   },
 
+  _removeNodes: function(nodeList, filterFn) {
+    for(var i = nodeList.length - 1; i >= 0; i--) {
```
gijsk: Nit: prevailing style is spaces after `if`, `for` etc. (before opening brace)

Readability.js
```diff
@@ -1641,14 +1643,14 @@ Readability.prototype = {
...
+    this._removeNodes(e.getElementsByTagName(tag), function(node) {
       var weight = this._getClassWeight(node);
       var contentScore = 0;
 
       this.log("Cleaning Conditionally", node);
 
       if (weight + contentScore < 0) {
-        node.parentNode.removeChild(node);
+        return true;
       } else if (this._getCharCount(node,',') < 10) {
```
gijsk: Nit: no else after return.

Readability.js
```diff
@@ -1667,30 +1669,26 @@ Readability.prototype = {
 
         var linkDensity = this._getLinkDensity(node);
         var contentLength = this._getInnerText(node).length;
-        var toRemove = false;
 
         // Make an exception for elements with no p's and exactly 1 img.
         if (img > p && !this._hasAncestorTag(node, "figure")) {
-          toRemove = true;
+          return true;
```
gijsk: Alternatively, could leave this the same and just `return toRemove` at the end of the outer block.

gijsk:
Sorry for the delay - I'm not usually around on weekends. I left some notes. I can make the changes myself later today and merge, if that sounds OK?

persidskiy:
don't bother, i'll do

Readability.js
persidskiy: ohmy, seems it should be added to eslint rules

Readability.js
```diff
@@ -1627,63 +1642,55 @@ Readability.prototype = {
...
+
+        var haveToRemove =
+          // Make an exception for elements with no p's and exactly 1 img.
+          (img > p && !this._hasAncestorTag(node, "figure")) ||
+          (!isList && li > p) ||
+          (input > Math.floor(p/3)) ||
+          (!isList && contentLength < 25 && (img === 0 || img > 2)) ||
+          (!isList && weight < 25 && linkDensity > 0.2) ||
+          (weight >= 25 && linkDensity > 0.5) ||
+          ((embedCount === 1 && contentLength < 75) || embedCount > 1);
```
persidskiy: I've decided to get rid of if clause. What do you think?

gijsk: LGTM!

PR #301, opened 2016-07-18
Fix code style, tighten up eslint rules

gijsk:
@msujaws , can you doublecheck these changes make sense? Just trying to ensure eslint helps us get a consistent style from contributors (see e.g. comments on #300), and readability itself is relatively easy to get into shape. Thanks!


msujaws:
I don't see a way to mark this as reviewed, but the changes look good to me. r=me

PR #303, opened 2016-07-28
Add microformats2 class names to heuristics

tigt:
Microformats updated their old `hentry` to [a newer
`h-entry`](http://microformats.org/wiki/h-entry).

With the [number of IndieWeb sites breaking into the
ten-thousands](http://tantek.com/2016/190/b1/state-of-indieweb-summit),
this seems like a fair idea.


gijsk:
Sorry for the extreme delay, I was on vacation for 2 weeks and this got lost in the noise since my return (most of what I do is in bugzilla, not github...). I promise I'm normally more prompt with giving feedback on PRs. :-\

This looks sane to me, though it would be nice if you had an example page where this is useful that we could use to augment the test set. Can do that in a separate PR / issue though.

PR #309, opened 2016-10-07
Bug 1167568 - Find a better parent if there are other nodes scores are high enough

evanxd:


gijsk:
Looks like there are failing tests. :-(

evanxd:
Let me investigate the failures.

At the same time, what do you think we score `article` element by default in the patch? Do you think it's a good approach for the issue? @gijsk

gijsk:
I'm not sure, I haven't checked the ehow thing in a while. It would be good if you could add a testcase for the article from bug 1167568 (including the ad HTML) and show that this change makes it pass. That would also help me understand why you're making this change and what the impact would be.

I think unfortunately a lot of websites use `<article>` tags for comments or for a list of linked articles, instead of for one "main" article on the main page. Which might make this approach tricky to make work.

evanxd:
Sure, let me add tests.

Scoring `article` element by is due to current algorithm might not find out a `<article>` element in a web page.

Just like you said, it's a really tricky issue.

evanxd:
@gijsk 

Added tests for the patch.

I haven't cleaned up the source code of web page[2](just copy the source code from the web page), and it causes some JSDOMParser errors. If we try to clean up the web page[1] to fix the JSDOMParser errors, it will not be the original web page and we don't test the readability lib in syntax (or other) error situations.

Is that what we want, just test the readability lib with the web pages without syntax errors? But in real world, there might be some syntax errors in web pages.

[1]: https://github.com/evanxd/readability/blob/bug-1167568/test/test-pages/ehow-2/source.html

gijsk:
In all our clients, we serialize documents as XML and feed that into JSDOMParser, or use an actual browser DOM, to run readability. So these parse issues are occurring because you're using HTML that has been serialized as HTML rather than XML. The `generate-testcase` script doesn't do any re-serializing, and when I last looked I couldn't figure out how to do this with `jsdom`. It might be possible by writing our own custom serializer for the DOM `jsdom` generates.

Anyway, the short story is that yes, you'll need to manually fix the issues with the testcase, or update the tools so they generate XML and you don't have to manually fix this problem.

gijsk:
Note also that quite a number of existing testcases are also failing with this change, so I think that while the failure for your new testcase might be due to this serialization issue, there are still other issues with your patch that will need to be addressed.

evanxd:
Hi @gijsk ,

I've updated the patch with a new idea, find a better parent if there are other (at least three) nodes'(in `topCandidates` array) scores which quite close to current `topCandidate` node.

Let me explain the root cause here, currently in code base, we find out a wrong `topCandidate` in the [web page][web-page]. It find out `<div class="content lead">` as `topCandidate` in the below DOM tree, but the correct one is `<div class="col-main">` because other `<div class="mod step">` nodes contain useful content we want.

Here is the thing, we need to find a good way to deal with the above situation(find a wrong `topCandidates`). So I would like to propose a strategy to find a new better parent(`topCandidates`) if there are other (at least 3) nodes' score in `topCandidates` array which are quite closed(bigger or smaller than 22%) with current `topCandidate` node.
(For sure, we can discuss the number, why is `3` and `22%`?)

I've wrote code with the above approach, and that cause (only) [two tests failed][ci], not too many. The failures seems to be caused by choosing a node which is an ancestor of previous `topCandidate` as new`topCandidate`. 

If the approach is on the good way, we can continue to discuss the two failures case by case and other details. What do you think?

```html
<div class="col-main">
    <header class="page-head bordered"></header>
    <article data-type="article">
        <div class="mod step">
            <div class="stepContent mod">
                <div class="content lead"></div>
            </div>
        </div>
        <span>
            <span>
                <div class="mod step"></div>
            </span>
        </span>
        <span>
            <span>
                <div class="mod step"></div>
            </span>
        </span>
        <span>
            <span>
                <div class="mod step"></div>
            </span>
        </span>
        <span></span>
        ...
</div>
```

[web-page]: http://www.ehow.com/how_4851888_throw-graduation-party-budget.html
[ci]: https://travis-ci.org/mozilla/readability/builds/180422858

gijsk:
I haven't found time to look at this in more detail so far - and quite likely won't before Hawaii - but here are some thoughts off-hand:
1) it sounds like we could achieve a similar goal by increasing the rate at which we pass on scores from children to parents/grandparents. What happens if we do that instead?
2) there's another pull request open from a contributor which fixes some of the sibling/parent scoring stuff a bit. It needs code style fixes, but I wonder if it could help us here.
3) in the HTML you pasted, it looks to me like we want `<article>` not `<div class=col-main>`. What am I missing?

evanxd:
Reply the [above comment][above-comment].

For Point 1:
Do you mean that we make  `scoreDivider` smaller to let ancestors' `contentScore` higher in below [code][code-1]? Then we could get a better candidate?
```js
if (level === 0)
  var scoreDivider = 1;
else if (level === 1)
  scoreDivider = 2;
else
  scoreDivider = level * 3;
ancestor.readability.contentScore += contentScore / scoreDivider;
```

If so, this idea might work but it might also cause many tests failed. Just want to ensure we're in same page for the idea, then we can try it. We can compare the two approaches(this idea and the implementation in the pull request) and choice better one we think.

For Point 2:
Could you tell me which pull request? I could not find out the pull request.

For Point 3:
Yes, you're right. The `<article>` node is what we want.

[code-1]: https://github.com/mozilla/readability/blob/master/Readability.js#L769-L779
[above-comment]: https://github.com/mozilla/readability/pull/309#issuecomment-264451181
[DEFAULT_TAGS_TO_SCORE]: https://github.com/mozilla/readability/blob/master/Readability.js#L110

gijsk:
For the `scoreDivider`: maybe - changing that or changing how many ancestors we score (it's limited to 3 right now). I don't know if we need only one of those 2 things or both, off-hand.

The other pull request is https://github.com/mozilla/readability/pull/313 .

evanxd:
```
For the scoreDivider: maybe - changing that or changing how many ancestors we score (it's limited to 3 right now). I don't know if we need only one of those 2 things or both, off-hand.
```

I've tried the above idea. Honestly, I didn't be successful to find the correct parent node with the idea, changing `scoreDivider` or changing the number of ancestors. I always find out the `<body>` as the parent node with the below code. And changing `scoreDivider` make regression issues easily.
```
var ancestors = this._getNodeAncestors(elementToScore, 5);

if (level === 0)
  var scoreDivider = 1;
else if (level === 1)
  scoreDivider = 1.025;
else
  scoreDivider = 1 + (level * 0.025);
ancestor.readability.contentScore += contentScore / scoreDivider;
```

#313  pull request doesn't help here

So I would like to suggest we use the method in the current pull request(#309)  to find a better parent node.

What do you think?

gijsk:
OK, so let's move forward with this idea. The implementation needs some work, though:

1. let's use a score similarity of 25% rather than 22% just because it's a round number. :-)
2. our goal is to find an ancestor that contains at least 3 items with a score at least 75% of the top candidate, one of which is of course the top candidate itself. The top candidate array is sorted by scores, so we should do this by:
   1. selecting the candidates with a high-enough score
   2. get an ancestor array for all of those except the top candidate
   3. for the top candidate (which must be included), iterate over its ancestors one by one (in a special loop) and check that it occurs in the ancestor array at least 2 other items besides the top candidate.
This minimizes the number of nested loops and the repetitive iterating over (the entire) ancestor chain. It also avoids using `Node.contains` which means we can use it with both JSDOMParser and jsdom.

Readability.js
```diff
@@ -112,7 +112,7 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
-    unlikelyCandidates: /banner|combx|comment|community|cover-wrap|disqus|extra|foot|header|legends|menu|modal|related|remark|rss|shoutbox|sidebar|skyscraper|sponsor|ad-break|agegate|pagination|pager|popup|yom-remote/i,
+    unlikelyCandidates: /banner|combx|comment|community|cover-wrap|disqus|extra|foot|header|legends|menu|modal|related|remark|replies|rss|shoutbox|sidebar|skyscraper|sponsor|ad-break|agegate|pagination|pager|popup|yom-remote/i,
```
evanxd: Add `replies` to avoid getting wrong top candidate the and fix the failure of [lifehacker-post-comment-load][lifehacker-post-comment-load] test.

[lifehacker-post-comment-load]: https://github.com/mozilla/readability/blob/master/test/test-pages/lifehacker-post-comment-load/source.html#L1202

evanxd:
@gijsk ,

Sure, I already updated patch for the [review comments][review-comments]. Could you take a look? Thanks.

[review-comments]: https://github.com/mozilla/readability/pull/309#issuecomment-268979152

Readability.js
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
           lastScore = parentOfTopCandidate.readability.contentScore;
           parentOfTopCandidate = parentOfTopCandidate.parentNode;
         }
+
+        // Find a better top candidate node if it contains (at least three) nodes which belong to `topCandidates` array
+        // and whose scores are quite closed with current `topCandidate` node.
+        var topCandidateAncestors = [];
+        topCandidates.some(function(_topCandidate, i) {
```
gijsk: Nit: rename the argument to `alternativeCandidate`

gijsk: In fact, it seems simpler to implement this as a raw for loop instead of using `some` with a function callback:

```js
for (let i = 1; i < topCandidates.length; i++) {
  if (/* score too low */) {
   break;
  }
  // add ancestors
}
```

using `some` (which is meant to answer a yes/no question about the array) in order to stop the iteration isn't the most obvious way of doing this.

Readability.js
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
...
+          if (i === 0) {
+            return false;
+          }
+          if (Math.abs((_topCandidate.readability.contentScore -
+                        topCandidate.readability.contentScore) /
+                        topCandidate.readability.contentScore) < 0.25) {
```
gijsk: The subtraction and Math.abs call here is not necessary. The array is sorted, and so the `alternativeCandidate` will always have a lower score. So you can do:

```js
if (alternativeCandidate.readability.contentScore / topCandidate.readability.contentScore >= 0.75) {
   // add ancestors to a set. 
}

File: Readability.js [hunk 4]
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
           lastScore = parentOfTopCandidate.readability.contentScore;
           parentOfTopCandidate = parentOfTopCandidate.parentNode;
         }
```
gijsk: Can we just replace this method of determining a better candidate by using an ancestor node with the method you're adding? If not, why not?

Readability.js
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
...
+          if (Math.abs((_topCandidate.readability.contentScore -
+                        topCandidate.readability.contentScore) /
+                        topCandidate.readability.contentScore) < 0.25) {
+            topCandidateAncestors.push(this._getNodeAncestors(_topCandidate));
+          }
+          return false;
+        }.bind(this));
+
+        topCandidateAncestors.some(function(ancestors, i) {
+          var that = this;
```
gijsk: You're using `.bind`, so this is not necessary.

Readability.js
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
...
+          return that._someNode(ancestors, function(topCandidateAncestor) {
+            var numberOfHighScoreNode = 0;
+            return topCandidateAncestors.some(function(_ancestors, j) {
```
gijsk: You're using 3 nested loops and another `indexOf` check. You shouldn't need that many loops. Instead, first check that we found at least 3 nodes, then loop over the ancestors of the top candidate and check if the list you built up of the ancestors of the other candidates has at least 3 sub-lists that contain the ancestor:

```js
var MINIMUM_TOPCANDIDATES = 3;
if (alternativeCandidateAncestors.length >= MINIMUM_TOPCANDIDATES) {
  var topCandidateAncestor = topCandidate.parentNode;
  while (topCandidateAncestor) {
    var listsContainingThisAncestor = 0;
    for (var ancestorIndex = 0; ancestorIndex < alternativeCandidateAncestors.length && listsContainingThisAncestor < MINIMUM_TOPCANDIDATES; ancestorIndex++) {
       listsContainingThisAncestor += Number(ancestorsForOneNode.includes(topCandidateAncestor));
    }
    if (listsContainingThisAncestor == MINIMUM_TOPCANDIDATES) {
      topCandidate = topCandidateAncestor;
      break;
    }
    topCandidateAncestor = topCandidateAncestor.parentNode;
  }
}
```

Readability.js
```diff
@@ -867,6 +878,41 @@ Readability.prototype = {
...
+        // and whose scores are quite closed with current `topCandidate` node.
+        var topCandidateAncestors = [];
+        topCandidates.some(function(_topCandidate, i) {
+          if (i === 0) {
+            return false;
+          }
+          if (Math.abs((_topCandidate.readability.contentScore -
+                        topCandidate.readability.contentScore) /
+                        topCandidate.readability.contentScore) < 0.25) {
+            topCandidateAncestors.push(this._getNodeAncestors(_topCandidate));
```
gijsk: Likewise, the difference between this array and e.g. the result of getting all the ancestors of the "main" current `topCandidate` would probably be more clear if you called the list `alternativeCandidateAncestors`.

Readability.js
gijsk: Eh, this should use `var` of course, not `let`, and it probably needs a unique iterator variable.

Maybe for clarity we should factor it out into its own function? `lookForSharedAncestorAsTopCandidate` or something?

Readability.js
evanxd: Good point.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - '5.3'
+  - '6.0'
```
evanxd: For supporting `Array.includes` method.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
         var scoreThreshold = lastScore / 3;
-        while (parentOfTopCandidate && parentOfTopCandidate.readability) {
+        while (parentOfTopCandidate && parentOfTopCandidate.tagName !== "BODY") {
+          if (!parentOfTopCandidate.readability) {
+            parentOfTopCandidate = parentOfTopCandidate.parentNode;
+            continue;
+          }
           var parentScore = parentOfTopCandidate.readability.contentScore;
           if (parentScore < scoreThreshold)
             break;
```
evanxd: We don't remove the below method because these two methods are for different situations(one is the below, another one is add just added in the PR). For example, if we remove the below method the [`herald-sun-1`][herald-sun-1] test cannot be passed(some other tests cannot be passed neither) because there only one item(`<div class="story-intro">`) in `topCandidates` array is part of main content and that its score is not high enough. We need at least 3 node and theirs scores need to be high enough(75%) for the new method I just added.

```
var lastScore = topCandidate.readability.contentScore;
// The scores shouldn't get too low.
var scoreThreshold = lastScore / 3;
while (parentOfTopCandidate && parentOfTopCandidate.tagName !== "BODY") {
  if (!parentOfTopCandidate.readability) {
    parentOfTopCandidate = parentOfTopCandidate.parentNode;
    continue;
  }
  var parentScore = parentOfTopCandidate.readability.contentScore;
  if (parentScore < scoreThreshold)
    break;
  if (parentScore > lastScore) {
    // Alright! We found a better parent to use.
    topCandidate = parentOfTopCandidate;
    break;
  }
  lastScore = parentOfTopCandidate.readability.contentScore;
  parentOfTopCandidate = parentOfTopCandidate.parentNode;
}
```

[herald-sun-1]: https://github.com/mozilla/readability/blob/master/test/test-pages/herald-sun-1/source.html#L642

Readability.js
evanxd: I replied this [here][here].

[here]: https://github.com/mozilla/readability/pull/309/files?diff=unified#r95936960

Readability.js
evanxd: Much better. Learned. Thanks a lot.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
 
         this._initializeNode(topCandidate);
       } else if (topCandidate) {
+        parentOfTopCandidate = topCandidate.parentNode;
+
+        // Find a better top candidate node if it contains (at least three) nodes which belong to `topCandidates` array
+        // and whose scores are quite closed with current `topCandidate` node.
+        var alternativeCandidateAncestors = [];
+        for (var i = 1; i < topCandidates.length; i++) {
```
evanxd: I would say this for loop is clear enough. I think we don't need to add new method `lookForSharedAncestorAsTopCandidate` for clarity.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+          if (topCandidates[i].readability.contentScore / topCandidate.readability.contentScore >= 0.75) {
+            alternativeCandidateAncestors.push(this._getNodeAncestors(topCandidates[i]));
+          }
+        }
```
evanxd: I would say this for loop is clear enough. I think we don't need to add new method lookForSharedAncestorAsTopCandidate for clarity.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
         // few steps up the tree, that's a decent sign that there might be more content
         // lurking in other places that we want to unify in. The sibling stuff
         // below does some of that - but only if we've looked high enough up the DOM
         // tree.
-        parentOfTopCandidate = topCandidate.parentNode;
         var lastScore = topCandidate.readability.contentScore;
         // The scores shouldn't get too low.
         var scoreThreshold = lastScore / 3;
-        while (parentOfTopCandidate && parentOfTopCandidate.readability) {
+        while (parentOfTopCandidate && parentOfTopCandidate.tagName !== "BODY") {
```
evanxd: I think we don't need the `<body>` node as new `topCandidate`.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+          if (!parentOfTopCandidate.readability) {
+            parentOfTopCandidate = parentOfTopCandidate.parentNode;
+            continue;
+          }
```
evanxd: Don't stop to find new `topCandidate` because one of `topCandidate`'s ancestors don't has score. Only stop when `if (parentScore < scoreThreshold)` or `if (parentScore > lastScore) `. Also like the method just added in this PR.

evanxd:
Hi @gijsk ,

Thanks for previous review. I've updated patch for your comments and left some comments. Could you take a look? Thanks.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
         // few steps up the tree, that's a decent sign that there might be more content
         // lurking in other places that we want to unify in. The sibling stuff
         // below does some of that - but only if we've looked high enough up the DOM
         // tree.
-        parentOfTopCandidate = topCandidate.parentNode;
         var lastScore = topCandidate.readability.contentScore;
         // The scores shouldn't get too low.
         var scoreThreshold = lastScore / 3;
-        while (parentOfTopCandidate && parentOfTopCandidate.readability) {
+        while (parentOfTopCandidate.tagName !== "BODY") {
```
evanxd: I think we don't need the `<body>` node as new topCandidate.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+          if (!parentOfTopCandidate.readability) {
+            parentOfTopCandidate = parentOfTopCandidate.parentNode;
+            continue;
+          }
```
evanxd: Don't stop to find new `topCandidate` because one of `topCandidate`'s ancestors don't has score. Only stop when `if (parentScore < scoreThreshold)` or `if (parentScore > lastScore) `. Also like the method just added in this PR.

Readability.js
gijsk: I think it's clear enough, but I'm aware that because we can't assume `let` exists (at least, historically we haven't been able to - maybe it's better now? Does `node` 6.0 support `let`, and does our iOS runtime environment? Esp. for the last question, I don't know the answer...

In any case, it works like this so we can move the refactoring and/or use of `let` to a separate issue.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+        var alternativeCandidateAncestors = [];
+        for (var i = 1; i < topCandidates.length; i++) {
+          if (topCandidates[i].readability.contentScore / topCandidate.readability.contentScore >= 0.75) {
+            alternativeCandidateAncestors.push(this._getNodeAncestors(topCandidates[i]));
+          }
+        }
+        var MINIMUM_TOPCANDIDATES = 3;
+        if (alternativeCandidateAncestors.length >= MINIMUM_TOPCANDIDATES) {
+          while (parentOfTopCandidate.tagName !== "BODY") {
+            if (!parentOfTopCandidate.readability) {
```
gijsk: Why do we need to skip ancestors without a `readability` property? Couldn't we generate it after the fact if we decide to choose it?

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
 
         this._initializeNode(topCandidate);
       } else if (topCandidate) {
+        parentOfTopCandidate = topCandidate.parentNode;
```
gijsk: This can go underneath the next loop, because we don't use it here.

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+        }
+        var MINIMUM_TOPCANDIDATES = 3;
+        if (alternativeCandidateAncestors.length >= MINIMUM_TOPCANDIDATES) {
+          while (parentOfTopCandidate.tagName !== "BODY") {
+            if (!parentOfTopCandidate.readability) {
+              parentOfTopCandidate = parentOfTopCandidate.parentNode;
+              continue;
+            }
+            var listsContainingThisAncestor = 0;
+            for (var ancestorIndex = 0; ancestorIndex < alternativeCandidateAncestors.length; ancestorIndex++) {
```
gijsk: How often is the `alternativeCandidateAncestors` array "long"? In my previous review I think I added the `listsContainingThisAncestor >= MINIMUM_TOPCANDIDATES` condition to the loop, which would avoid checking the remaining lists once you've seen the ancestor as relevant for enough candidates, which seems like it'd make sense? Though thinking about it, I guess the condition should be inverted in the loop (ie `listsContainingThisAncestor < MINIMUM_TOPCANDIDATES`)

Readability.js
```diff
@@ -844,18 +844,50 @@ Readability.prototype = {
...
+        }
+
         // Because of our bonus system, parents of candidates might have scores
         // themselves. They get half of the node. There won't be nodes with higher
         // scores than our topCandidate, but if we see the score going *up* in the first
         // few steps up the tree, that's a decent sign that there might be more content
         // lurking in other places that we want to unify in. The sibling stuff
         // below does some of that - but only if we've looked high enough up the DOM
         // tree.
-        parentOfTopCandidate = topCandidate.parentNode;
```
gijsk: I think this needs to stay here, because we might have just changed the `topCandidate`.

test/test-pages/ehow-2/expected.html
```diff
@@ -0,0 +1,227 @@
...
+                    this.src = failover;
+                }" /> </a>
+                    <p class="btn btnFollow" data-type="contributor" data-author-url="/contributor/gina_robertsgrey/" data-follow-data="{&quot;name&quot;:&quot;Gina Roberts-Grey&quot;,&quot;slug&quot;:&quot;\/contributor\/gina_robertsgrey\/&quot;,&quot;image_url&quot;:&quot;http:\/\/s3.amazonaws.com\/cme_public_images\/www_demandstudios_com\/sitelife.studiod.com\/ver1.0\/Content\/images\/store\/9\/2\/d9dd6f61-b183-4893-927f-5b540e45be91.Small.jpg&quot;,&quot;website&quot;:&quot;&quot;}">Follow</p>
+                </div>
+                <p class="article-meta">
+                    <time datetime="2016-09-14T07:07:00-04:00" itemprop="dateModified">Last updated September 14, 2016</time>
+                </p>
+            </div>
+            <div class="clearfix mg-3">
+                <ul class="social-icons js-social-icons inline " data-url="http://www.ehow.com/how_4851888_throw-graduation-party-budget.html" share-buttons-uninitialized="">
```
gijsk: Can we make these go away?

test/test-pages/ehow-2/expected.html
```diff
@@ -0,0 +1,227 @@
+<div id="readability-page-1" class="page">
+    <section id="Body" class="InlineTemplate FLC" data-page-id="inlinetemplate" data-section="body">
+        <ol class="breadcrumbs unstyled inline" itemscope="" itemtype="http://data-vocabulary.org/Breadcrumb" data-module="breadcrumb">
```
gijsk: Can we avoid including the breadcrumbs?

test/test-pages/ehow-2/expected.html
```diff
@@ -0,0 +1,227 @@
...
+</figure>
+<figcaption class="small caption">
+    jethuynh/iStock/Getty Images </figcaption>
+</div>
+</div>
+</span>
+</span>
+
+
+<h2 class="RsTitle head mg-2 title">
```
gijsk: Ideally we shouldn't include this either, but I don't know off-hand how best to do this, so let's leave it for now.

Readability.js
gijsk: To clarify: we'll need to set this both here *and* above.

Readability.js
evanxd: Sure, we can do that. I just tried to follow the logic at [1].
```
while (parentOfTopCandidate && parentOfTopCandidate.readability) {
  // Do something.
}
```

[1]: https://github.com/mozilla/readability/blob/master/Readability.js#L858

Readability.js
evanxd: Sure.

Readability.js
evanxd: Sure.

Readability.js
evanxd: Got it. Let's add it(`listsContainingThisAncestor < MINIMUM_TOPCANDIDATES`) back.

test/test-pages/ehow-2/expected.html
evanxd: Sure, I think we can add `social` into the `unlikelyCandidates` REGEXP to remove it.

test/test-pages/ehow-2/expected.html
evanxd: Sure, Sure, I think we can add `breadcrumbs` into the `unlikelyCandidates` REGEXP to remove it.

Readability.js
evanxd: Node supports `let` after [version 4.3.2][node-4-3-2]. iOS might support it after [version 10.1][ios-10-1].

I think we could start to use `let` after over 95% (even 100%) of iOS users upgrade to version 10.1+. For now, only [78.1% of iOS users][upgrade-ios-10-1] have upgraded to version 10.1+. What do you think?

[node-4-3-2]: http://node.green/#let
[ios-10-1]: http://caniuse.com/#feat=let
[upgrade-ios-10-1]: https://david-smith.org/iosversionstats

test/test-pages/ehow-2/expected.html
evanxd: Agree.

Readability.js
evanxd: I filed a [bug][bug] to address usage of `let`.

[bug]: https://bugzilla.mozilla.org/show_bug.cgi?id=1331336

Readability.js
```diff
@@ -844,6 +844,31 @@ Readability.prototype = {
...
+              listsContainingThisAncestor += Number(alternativeCandidateAncestors[ancestorIndex].includes(parentOfTopCandidate));
+            }
+            if (listsContainingThisAncestor >= MINIMUM_TOPCANDIDATES) {
+              topCandidate = parentOfTopCandidate;
+              break;
+            }
+            parentOfTopCandidate = parentOfTopCandidate.parentNode;
+          }
+        }
+        !topCandidate.readability && this._initializeNode(topCandidate);
```
gijsk: Please make this a normal if statement.

gijsk: I think this is the only thing - with that change, let's merge this in. Thanks for all your work on this!

evanxd: Sure. And thanks a lot for your review. :)

PR #310, opened 2016-10-11
Bug 1177619 - Score div nodes which have br nodes. r=Gijs

evanxd:
The solution here is score `<br>` tags because blogger use `<br>` tags to typeset their articles not `<p>` tags.

Their DOM structure looks like

```
<div>
  Dare to predict and suggest and improve your prediction based on time. Always try to add some more value to the user.<br />
  Its a continuous process, reiterate the above point &nbsp;time to time.<br />
  Let me know your thoughts how to get a wow ....<br />
  <br />
  Dare to predict and suggest and improve your prediction based on time. Always try to add some more value to the user.<br />
  Its a continuous process, reiterate the above point &nbsp;time to time.<br />
  Let me know your thoughts how to get a wow ....<br />
</div>
```

I also added tests. And the test web page is from http://www.myexperimentswithjava.com/2016/06/magic-with-computers.html.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1722,6 +1722,22 @@ Readability.prototype = {
   isProbablyReaderable: function(helperIsVisible) {
     var nodes = this._getAllNodesWithTag(this._doc, ["p", "pre"]);
 
+    // Get <div> nodes which have <br> node(s) and append them into the `nodes` variable.
+    // Some articles' DOM structures might look like
+    // <div>
+    //   Sentences<br>
+    //   <br>
+    //   Sentences<br>
+    // </div>
+    var brNodes = this._getAllNodesWithTag(this._doc, ["div > br"]);
+    if (brNodes.length) {
+      var set = new Set();
+      [].forEach.call(brNodes, function(node) {
+        set.add(node.parentNode);
+      });
+      nodes = [].concat.apply(Array.from(set), nodes);
+    }
+
     // FIXME we should have a fallback for helperIsVisible, but this is
     // problematic because of jsdom's elem.style handling - see
     // https://github.com/mozilla/readability/pull/186 for context.
@@ -1826,13 +1842,15 @@ Readability.prototype = {
     }
 
     var textContent = articleContent.textContent;
-    return { uri: this._uri,
-             title: articleTitle,
-             byline: metadata.byline || this._articleByline,
-             dir: this._articleDir,
-             content: articleContent.innerHTML,
-             textContent: textContent,
-             length: textContent.length,
-             excerpt: metadata.excerpt };
+    return {
+      uri: this._uri,
+      title: articleTitle,
+      byline: metadata.byline || this._articleByline,
+      dir: this._articleDir,
+      content: articleContent.innerHTML,
+      textContent: textContent,
+      length: textContent.length,
+      excerpt: metadata.excerpt,
+    };
   }
 };
--- a/test/test-pages/blogger/expected-metadata.json
+++ b/test/test-pages/blogger/expected-metadata.json
@@ -1,0 +1,6 @@
+{
+  "title": "Open Verilog flow for Silego GreenPak4 programmable logic devices",
+  "byline": null,
+  "excerpt": "I've written a couple of posts in the past few months but they were all for",
+  "readerable": true
+}
--- a/test/test-pages/blogger/expected.html
+++ b/test/test-pages/blogger/expected.html
@@ -1,0 +1,147 @@
+<div id="readability-page-1" class="page">
+    <div class="post-body entry-content" id="post-body-932306423056216142" itemprop="description articleBody">
+        <p style="display: inline;" class="readability-styled">
+            I've written a couple of posts in the past few months but they were all for </p><a href="http://blog.ioactive.com/search/label/Andrew%20Zonenberg">the blog at work</a>
+        <p style="display: inline;" class="readability-styled"> so I figured I'm long overdue for one on Silicon Exposed.</p>
+        <p>
+
+            <h2>
+                So what's a GreenPak?</h2>
+        </p>
+        <p style="display: inline;" class="readability-styled"> Silego Technology is a fabless semiconductor company located in the SF Bay area, which makes (among other things) a line of programmable logic devices known as GreenPak. Their </p><a href="http://www.silego.com/products/greenpak5.html">5th generation parts</a>
+        <p style="display: inline;" class="readability-styled"> were just announced, but I started this project before that happened so I'm still targeting the </p><a href="http://www.silego.com/products/greenpak4.html">4th generation</a>
+        <p style="display: inline;" class="readability-styled">.</p>
+        <p>
+            GreenPak devices are kind of like itty bitty <a href="http://www.cypress.com/products/32-bit-arm-cortex-m-psoc">PSoCs</a> - they have a mixed signal fabric with an ADC, DACs, comparators, voltage references, plus a digital LUT/FF fabric and some typical digital MCU peripherals like counters and oscillators (but no CPU).</p>
+        <p>
+            It's actually an interesting architecture - FPGAs (including some devices marketed as CPLDs) are a 2D array of LUTs connected via wires to adjacent cells, and true (product term) CPLDs are a star topology of AND-OR arrays connected by a crossbar. GreenPak, on the other hand, is a star topology of LUTs, flipflops, and analog/digital hard IP connected to a crossbar.</p>
+        <p>
+            Without further ado, here's a block diagram showing all the cool stuff you get in the SLG46620V:</p>
+        <p>
+
+
+            They're also tiny (the SLG46620V is a 20-pin 0.4mm pitch STQFN measuring 2x3 mm, and the lower gate count SLG46140V is a mere 1.6x2 mm) and probably the cheapest programmable logic device on the market - $0.50 in low volume and less than $0.40 in larger quantities.</p>
+        <p>
+            The Vdd range of GreenPak4 is huge, more like what you'd expect from an MCU than an FPGA! It can run on anything from 1.8 to 5V, although performance is only specified at 1.8, 3.3, and 5V nominal voltages. There's also a dual-rail version that trades one of the GPIO pins for a second power supply pin, allowing you to interface to logic at two different voltage levels.</p>
+        <p>
+            To support low-cost/space-constrained applications, they even have the configuration memory on die. It's one-time programmable and needs external Vpp to program (presumably Silego didn't want to waste die area on charge pumps that would only be used once) but has a SRAM programming mode for prototyping.</p>
+        <p>
+            The best part is that the development software (GreenPak Designer) is free of charge and provided for all major operating systems including Linux! Unfortunately, the only supported design entry method is schematic entry and there's no way to write your design in a HDL.</p>
+        <p>
+            While schematics may be fine for quick tinkering on really simple designs, they quickly get unwieldy. The nightmare of a circuit shown below is just a bunch of counters hooked up to LEDs that blink at various rates.</p>
+        <p>
+
+
+            As if this wasn't enough of a problem, the largest GreenPak4 device (the SLG46620V) is split into two halves with limited routing between them, and the GUI doesn't help the user manage this complexity at all - you have to draw your schematic in two halves and add "cross connections" between them.</p>
+        <p>
+            The icing on the cake is that schematics are a pain to diff and collaborate on. Although GreenPak schematics are XML based, which is a touch better than binary, who wants to read a giant XML diff and try to figure out what's going on in the circuit?</p>
+        <p>
+            This isn't going to be a post on the quirks of Silego's software, though - that would be boring. As it turns out, there's one more exciting feature of these chips that I didn't mention earlier: the configuration bitstream is 100% documented in the device datasheet! This is unheard of in the programmable logic world. As Nick of Arachnid Labs <a href="http://www.arachnidlabs.com/blog/2015/03/30/greenpak/">says</a>, the chip is "just dying for someone to write a VHDL or Verilog compiler for it". As you can probably guess by from the title of this post, I've been busy doing exactly that.</p>
+        <p>
+
+            <h2>
+                Great! How does it work?</h2>
+        </p>
+        <p style="display: inline;" class="readability-styled"> Rather than wasting time writing a synthesizer, I decided to write a GreenPak technology library for Clifford Wolf's excellent open source synthesis tool, </p><a href="http://www.clifford.at/yosys/">Yosys</a>
+        <p style="display: inline;" class="readability-styled">, and then make a place-and-route tool to turn that into a final netlist. The post-PAR netlist can then be loaded into GreenPak Designer in order to program the device.</p>
+        <p>
+            The first step of the process is to run the "synth_greenpak4" Yosys flow on the Verilog source. This runs a generic RTL synthesis pass, then some coarse-grained extraction passes to infer shift register and counter cells from behavioral logic, and finally maps the remaining logic to LUT/FF cells and outputs a JSON-formatted netlist.</p>
+        <p>
+            Once the design has been synthesized, my tool (named, surprisingly, gp4par) is then launched on the netlist. It begins by parsing the JSON and constructing a directed graph of cell objects in memory. A second graph, containing all of the primitives in the device and the legal connections between them, is then created based on the device specified on the command line. (As of now only the SLG46620V is supported; the SLG46621V can be added fairly easily but the SLG46140V has a slightly different microarchitecture which will require a bit more work to support.)</p>
+        <p>
+            After the graphs are generated, each node in the netlist graph is assigned a numeric label identifying the type of cell and each node in the device graph is assigned a list of legal labels: for example, an I/O buffer site is legal for an input buffer, output buffer, or bidirectional buffer.</p>
+        <p>
+
+
+            The labeled nodes now need to be placed. The initial placement uses a simple greedy algorithm to create a valid (although not necessarily optimal or even routable) placement:</p><br/>
+        <ol>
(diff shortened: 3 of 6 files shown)


evanxd:
The CI error is about JSDOMParser error. We can fix it by clean up the web page. Before that, let's make sure the approach of the patch is good.

Readability.js
```diff
@@ -1743,12 +1743,15 @@ Readability.prototype = {
         return false;
       }
 
-      var textContentLength = node.textContent.trim().length;
-      if (textContentLength < 140) {
-        return false;
+      // In some web pages, they use "<br>" to typeset articles.
+      if (node.tagName === "BR") {
+        score += 2;
```
gijsk: So there are 2 problems with this.

1. This will mean that any page that includes more than 10 `<br>` tags will be marked as 'readerable' (so they'll get a reader mode icon). That doesn't seem like a good idea.
1. Performance. This code runs when we load any webpage. It needs to be as light-weight as possible.

Readability.js
```diff
@@ -1743,12 +1743,15 @@ Readability.prototype = {
...
       }
 
-      score += Math.sqrt(textContentLength - 140);
+      var textContentLength = node.textContent.trim().length;
+      if (textContentLength > 140) {
```
gijsk: Why the change compared to how this length code worked previously? I don't think it really matters in practice (ie I think the outwardly visible behaviour will be the same), but we might as well leave it the same to avoid polluting blame. :-)

Readability.js
evanxd: Yes, there is no cheap way to deal with all kinds of web pages made by the site users. There is a lot of situations.

After investigated the blogger articles, I found out many articles'[1], [2], [3], [4], [5], [6], [7], [8], [9], [10](random pages from 10 bloggers) DOM structures look like
```
<div>
  Sentences<br>
  <br>
  Sentences<br>
  <br>
  Sentences<br>
  <br>
  Sentences<br>
  <br>
</div>
``` 

So how about we focus this pattern of DOM structure first? If OK, we can try to implement a light-weight algorithm to detect the pattern.

Or do we prefer to implement an algorithm to deal with almost all kinds of web page structure made by blogger users? If so, I think we should try to list all kinds of DOM structures look like articles first. Then we can implement an algorithm to detect them.

What do you think, @gijsk ? :)

[1]: http://siliconexposed.blogspot.tw/2016/05/open-verilog-flow-for-silego-greenpak4.html
[2]: http://www.stevenhoneyman.co.uk/2016/02/bga-on-budget.html
[3]: http://sequence15.blogspot.tw/2015/11/review-ssl-1250-quad-lfo.html
[4]: http://teslaprime.blogspot.tw/2016/04/the-behavior-of-capacitor.html
[5]: http://www.tarduino.cc/2016/07/fixing-stuck-touchpad-on-toshiba.html
[6]: http://itexpertinbd.blogspot.tw/2014/11/email-id-sell-email-marketing-bd-id.html
[7]: http://enginemonitor.blogspot.tw/2016/09/becoming-embedded-programmer.html
[8]: http://www.myexperimentswithjava.com/2016/06/magic-with-computers.html
[9]: http://decimater.blogspot.tw/2016/09/driver-memory-leak.html
[10]: http://www.bluetrainsoftware.com/2014/11/polymer-without-dependency-injection-is.html

Readability.js
```diff
@@ -1722,6 +1722,23 @@ Readability.prototype = {
...
+    var brNodes = this._getAllNodesWithTag(this._doc, ["div > br"]);
+    if (brNodes.length > 1) {
+      var set = new Set();
+      [].forEach.call(brNodes, function(node) {
+        var div = node.parentNode;
+        !set.has(div) && set.add(div);
+      });
+      nodes = [].concat.call(Array.from(set), nodes);
+    }
+
```
evanxd: Hi, @gijsk 

Right, it's really hard to implement a light-weight algorithm for detecting `<div>` nodes with the pattern. But let's try to figure out that.

How about this? We select all `<div>` nodes which have `<br>` nodes. Then try to score them in our original score algorithm.

What do you think? Maybe you have better ideas. :)

evanxd: Looks like the patch I provided here is not acceptable, and at same time we don't have better ideas. Maybe we need to re-design the `isProbablyReaderable` method[1] here.

As I know, Safari's reader mode is quite good, so maybe we could try to use some ideas from Safari algorithm[2] to design our new `isProbablyReaderable` method(even porting the `adoptableArticle` method from Safari algorithm?). I think it can help us to resolve the problem.

What do you think, @gijsk ?

[1]: https://github.com/mozilla/readability/blob/master/Readability.js#L1722-L1758
[2]: https://github.com/adadgio/safari-reader-js/blob/master/safari-reader.js#L1021-L1045

evanxd: One more thing,  I've updated the patch[1] when I left the comment[2] few hours ago. I'm afraid maybe you haven't seen it.

Just want to ensure one thing here`var brNodes = this._getAllNodesWithTag(this._doc, ["div > br"]);`[1] is really too expensive for `isProbablyReaderable ` method?

Thanks for the responses. (go to bed first)

[1]: https://github.com/mozilla/readability/pull/310/commits/c9f287ac8ff4bdd782085231151455e34cdbd4d3#diff-06d8d22df421dacde90a2268083424abR1725
[2]: https://github.com/mozilla/readability/pull/310#discussion_r83186465

gijsk: Oh, right, I did miss that. Sorry.

I don't know what the perf impact of this suggestion would be. You'd have to try... it might be difficult because this won't select anything in nodejs, because there's no `querySelectorAll` implementation in JSDOMParser (and I don't think we should write one). It will only work against the real browser and maybe `jsdom`. But we need to keep testing JSDOMParser for Readability because we will use the DOM from there for the actual conversion to reader mode. So this is a bit tricky.

Otherwise, it looks at least like we could try it. You would need to modify the test framework to always use a `jsdom`-generated DOM for the `isProbablyReaderable` call. Then there are perf benchmarks already defined in the package.json that I think it would be possible to use to check what the impact is. If it's 5-10% slower that's probably OK. If it's 3 times as slow as it is now, that's probably less OK. Note that IIRC they might be quite noisy so you might need to run them multiple times.

Finally, note that the 'point' of a Set is that you can just call `.add` without checking `.has` - there will not be duplicates. :-)

Readability.js
```diff
@@ -1722,6 +1722,22 @@ Readability.prototype = {
...
     var nodes = this._getAllNodesWithTag(this._doc, ["p", "pre"]);
 
+    // Get <div> nodes which have <br> node(s) and append them into the `nodes` variable.
+    // Some articles' DOM structures might look like
+    // <div>
+    //   Sentences<br>
+    //   <br>
+    //   Sentences<br>
+    // </div>
+    var brNodes = this._getAllNodesWithTag(this._doc, ["br"]);
```
gijsk: I thought we were using `div > br` here? Why the change?

evanxd:
Hi @gijsk ,

I've fixed the test failures and done the benchmark experiments. 

The result(with jsdom)[1] is 38.46% slower after patched the patch[3]. Additionally, the result with JSDOMParser[2] is 24.38% slower. You can check the detail on the reports[1][2].

I Run `npm run perf-reference` for 10 times for each experiment.  And the slower rate = `(Patched Result - Original Result)/Original Result`.

What do you think of the benchmark result? Could it be acceptable?

[1]: https://docs.google.com/spreadsheets/d/1s8CXR8QCdSZuAyrVvuYM8Zh-4aq6H9ZRP5g6wpKHIi0
[2]: https://docs.google.com/spreadsheets/d/1I4S0yQZ4-I9p3YI3NIP4jiuSasL3qDtp4QGEcKuqt7Y
[3]: https://github.com/mozilla/readability/pull/310/commits/b5602a4b715efb9211200af3b7660d0929005ed6

Readability.js
evanxd: The `querySelectorAll` method cannot be used on `JSDOMParser` so we cannot use `div > br` here. Do we have any other way to use the`div > br` selector?

evanxd: I think we could have better performance if we use `div > br` here. But somehow we need the `querySelectorAll` method when using `JSDOMParser `.

gijsk: In practice isProbablyReaderable is only ever invoked on a "real" DOM, not on a JSDOMParser one. So I would go with the `div > br` option and update the test runner to only test `isProbablyReaderable` against a `jsdom` DOM (so `if` it out in the `JSDOMParser` case). Would that make sense?

evanxd:
Hi @gijsk ,

I've updated patch[1] for your comments. And the benchmark(with `jsdom`) result[2] is `36.87%` slower. You can check the detail of benchmark experiment on the report[2]. 

I Run `npm run perf-reference` for 10 times for each experiment. And the slower rate = `(Patched Result - Original Result)/Original Result`.

What do you think of the benchmark result of `isProbablyReaderable ` method(36.87% slower)? Could it be acceptable?

[1]: https://github.com/mozilla/readability/pull/310/commits/61af7898401869d49a0ffbb1e0c769b6b8893bd6
[2]: https://docs.google.com/spreadsheets/d/140SfoJ3YTkExhrsBgFSuL8SWTinPNY_kVl7TS7EfQ6U

gijsk:
37% is a lot. Especially because this is a decrease in ops/s, and so really what it means is that the operation itself is now 60% slower (ie the same operation is now 60% slower, so if it used to take e.g. 100ms to run it now takes 160, which means that on a sample time of 1000ms we run ~6.4 ops instead of 10, which looks like a 36% decrease in ops/s but is actually bigger in terms of absolute speed in ms.).

I don't know how `jsdom`'s querySelectorAll performance compares with that of the native DOMs we use. Do you think it would be possible to make the modification to Readability.js from the patch to the one in the tree, and then use the browser console to run some rough experiments on a few pages? I recognize this is more work than running the commandline benchmark, but it would be a more accurate estimate of what the perf impact will be in practice.

evanxd:
Sure, let's do the benchmark experiment with Gecko.

I added two performance checkpoints[1] to evaluate the performance of patched `isProbablyReaderable` method. We could check how much time took by `isProbablyReaderable` method in the performance experiment. For sure, lower is better.
```
var t0 = content.performance.now();
var isProbablyReaderable = ReaderMode.isProbablyReaderable(content.document);
var t1 = content.performance.now();
dump("Call to isProbablyReaderable took " + (t1 - t0) + " milliseconds. \n");
```

Then randomly chose 4 articles (listed in the below) from yahoo, wikipedia, and blogger to test. Two articles have 0 `<br>` node, and two articles some `<br>` node (selected by `div > br`).
* Article 1: https://www.yahoo.com/tech/virtual-reality-systems-compared-144526498.html (with 0 `<br>` node)
* Article 2: https://en.wikipedia.org/wiki/Mozilla (with 0 `<br>` node)
* Article 3: http://www.stevenhoneyman.co.uk/2016/02/bga-on-budget.html (with 11 `<br>` nodes)
* Article 4: http://siliconexposed.blogspot.tw/2016/05/open-verilog-flow-for-silego-greenpak4.html (with 75 `<br>` nodes)

Did performance test for 10 times for each article and averaged them. The experiment result is
|                    | Article 1 | Article 2 | Article 3 | Article 4 |
|--------------------|-----------|-----------|-----------|-----------|
| Original           | 4.121 ms  | 4.009 ms  | 4.139 ms  | 3.924 ms  |
| Patched            | 4.201 ms  | 4.199 ms  | 4.397 ms  | 5.077 ms  |
| Time Increase Rate | 1.9%      | 4.7%      | 6.2%      | 29.38     |
P.S. We could say `1.9%` slower here if Time Increase Rate is `1.9%`. You can check the experiment details here[2].

Then we could find out
* If there is no `<br>` node (can be selected by `div > br`), the performance could be almost same after patched.
* If there are too many(`> 75`) `<br>` node (can be selected by `div > br`), the performance is not good.

Performance issues might be occurred if there are too many `<br>` node (can be selected by `div > br`). But maybe we should accept it because
1. Seems there might be no other way to fix this issue.
2. Only webpages with many `<br>` nodes have the performance issue. The performance of others looks good.

What do you think of it, @gijsk ?

[1]: https://gist.github.com/evanxd/fc5eb3e33732567957aa0c02baaffc00#file-tab-content-js-L9-L12
[2]: https://docs.google.com/spreadsheets/d/1vk7bU_Z0KG4F7OEGlVWX_SA8IrH94lwrNVcw6zYBR6E

gijsk:
I guess we might have to take the hit, then, yeah. :-(

It would be useful if we had a way of avoiding the hit. Perhaps we can file a followup bug to optimize this method? I bet that either running these checks sequentially or unifying the different selectors might help.

gijsk:
Though to be fair, a 30% hit on a br-heavy page is a lot better than what the jsdom tests suggested.

Readability.js
```diff
@@ -1722,6 +1722,22 @@ Readability.prototype = {
...
+    //   <br>
+    //   Sentences<br>
+    // </div>
+    var brNodes = this._getAllNodesWithTag(this._doc, ["div > br"]);
+    if (brNodes.length) {
+      var set = new Set();
+      [].forEach.call(brNodes, function(node) {
+        set.add(node.parentNode);
+      });
+      nodes = [].concat.apply(Array.from(set), nodes);
```
evanxd: Looks like the new idea might not be faster. I would say they are almost same.

I run 10 times performance test for each one on Firefox. The result is `5.952 ms` vs `6.12 `(origin vs new idea).

The below is the full test report.

|     | Origin | New Idea |
|-----|----------|----------|
| 1.  | 5.46 ms  | 6.09 ms  |
| 2.  | 6.00 ms  | 6.23 ms  |
| 3.  | 6.01 ms  | 6.13 ms  |
| 4.  | 6.01 ms  | 5.91 ms  |
| 5.  | 5.76 ms  | 6.53 ms  |
| 6.  | 5.63 ms  | 6.04 ms  |
| 7.  | 5.75 ms  | 6.15 ms  |
| 8.  | 6.22 ms  | 5.51 ms  |
| 9.  | 6.55 ms  | 6.41 ms  |
| 10. | 6.13 ms  | 6.20 ms  |
| AVG | 5.952 ms | 6.12 ms  |

evanxd:
Sure, I've filed a followup bug[1] to optimize the method.

So what's our next step for the patch? Just start to write tests with a blogger page(maybe use the page[2] with many `<br>` tags)?

Looks like we're almost done here. :-)

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1313301
[2]: http://siliconexposed.blogspot.tw/2016/05/open-verilog-flow-for-silego-greenpak4.html

gijsk:
Yes, add a test, make sure tests pass, and update any styling nits (haven't looked at the patch again), then we should be done.

evanxd:
Got it. Let me add the tests.

One more question, update what styling nits? It seems that we haven't discussed any (JS code) styling nits before for the patch.

gijsk:
If there aren't any, that's fine then. :-)

evanxd:
Weird, the eslint failures seems not be caused by the patch. Investigate...

```
/home/travis/build/mozilla/readability/Readability.js
  1846:14  error  Expected indentation of 6 spaces but found 13  indent
  1847:14  error  Expected indentation of 6 spaces but found 13  indent
  1848:14  error  Expected indentation of 6 spaces but found 13  indent
  1849:14  error  Expected indentation of 6 spaces but found 13  indent
  1850:14  error  Expected indentation of 6 spaces but found 13  indent
  1851:14  error  Expected indentation of 6 spaces but found 13  indent
  1852:14  error  Expected indentation of 6 spaces but found 13  indent
```
https://travis-ci.org/mozilla/readability/builds/172231876#L397-L404

Readability.js
```diff
@@ -1826,13 +1842,15 @@ Readability.prototype = {
...
+    return {
+      uri: this._uri,
+      title: articleTitle,
+      byline: metadata.byline || this._articleByline,
+      dir: this._articleDir,
+      content: articleContent.innerHTML,
+      textContent: textContent,
+      length: textContent.length,
+      excerpt: metadata.excerpt
+    };
```
evanxd: I updated here because of the [eslint failures][eslint-failures]. The root cause is mentioned [here].
```
/home/travis/build/mozilla/readability/Readability.js
  1846:14  error  Expected indentation of 6 spaces but found 13  indent
  1847:14  error  Expected indentation of 6 spaces but found 13  indent
  1848:14  error  Expected indentation of 6 spaces but found 13  indent
  1849:14  error  Expected indentation of 6 spaces but found 13  indent
  1850:14  error  Expected indentation of 6 spaces but found 13  indent
  1851:14  error  Expected indentation of 6 spaces but found 13  indent
  1852:14  error  Expected indentation of 6 spaces but found 13  indent
```

[eslint-failures]: https://travis-ci.org/mozilla/readability/builds/172231876#L397-L404
[here]: https://github.com/mozilla/readability/pull/310#issuecomment-257528601

Readability.js
```diff
@@ -1826,13 +1842,15 @@ Readability.prototype = {
...
-             excerpt: metadata.excerpt };
+    return {
+      uri: this._uri,
+      title: articleTitle,
+      byline: metadata.byline || this._articleByline,
+      dir: this._articleDir,
+      content: articleContent.innerHTML,
+      textContent: textContent,
+      length: textContent.length,
+      excerpt: metadata.excerpt
```
gijsk: Nit: if we're doing this anyway, please add a trailing comma to this line.

gijsk:
I'm pretty reluctant to fix the eslint version because IME this ends up meaning you get stuck on outdated versions if you're not paying attention (which we tend not to :-) ). So I'd just leave it as-is and fix the indentation while we're here.

Readability.js
evanxd: OK, let's do it.

evanxd:
Sure, got it and agreed.

Added the trailing comma for your comment. Thanks for reviewing. :-)

PR #312, opened 2016-10-14
Bug 1310074 - Tests for yahoo.com. r=Gijs

evanxd:
Hi @gijsk ,

I've added tests for yahoo.com, and they are all passed on Travis CI.

Thanks.


gijsk:
This seems to have the same problem as the wikipedia patch (spacing gone missing).

gijsk:
@evanxd , do you want to update this patch now that the wikipedia tests have landed?

evanxd:
@gijsk ,

Sure, I've updated the patch to fix the space issues. And this is the [`source.html` page][source-page] from Yahoo.

Thanks for reviewing.

[source-page]: http://finance.yahoo.com/news/best-psvr-games-170003443.html

gijsk:
Nice, thanks!

PR #313, opened 2016-11-02
Bug fix: _prepDocument removes every other <font> element

andrei-ch:
Since doc.getElementsByTagName("font") returns live array, _forEachNode
must be called with backward=true or it will skip elements.

Also, _forEachNode with backward=true was NYI.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -719,12 +719,14 @@ Readability.prototype = {
           } else {
             // EXPERIMENTAL
             this._forEachNode(node.childNodes, function(childNode) {
-              if (childNode.nodeType === Node.TEXT_NODE && childNode.textContent.trim() != "") {
+              if (childNode.nodeType === Node.TEXT_NODE && childNode.textContent.trim().length > 0) {
                 var p = doc.createElement('p');
                 p.textContent = childNode.textContent;
                 p.style.display = 'inline';
                 p.className = 'readability-styled';
                 node.replaceChild(p, childNode);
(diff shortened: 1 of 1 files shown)


gijsk:
Thanks for your PR.

Over in #300, we decided instead of having the 'backward' param, we'd just have a helper to remove items instead. I think it would be clearer if we did the same for the `_setNodeTag` usage - we could have a helper called `_rewriteAllNodes` that takes the original tag name and new tag name as an argument, for instance. We should remove the confusing unused `backward` parameter - I'm a bit surprised eslint didn't pick those up.

eslint noted some styling issues with the indentation, could you fix those?

Otherwise, could you explain how the extra cleanup calls for input/textarea/button/select relate to the bugfix for `<font>` ?

gijsk:
I separated out the `font` tag mistakes and cleaning up the form elements, fixed some tests and code style issues and merged those to master. Thank you!

It seems to me some of the reparenting logic would be interesting and probably fixes real bugs, but at the moment it's causing a lot of test failures. Have you had a chance to look at those at all? You should be able to run the tests locally with something like `mocha test/test-readability.js`, optionally passing something like `-g clean-links` to select a specific test to run. It's possible that in some (maybe even most?) cases we simply end up updating the testcases, if the result is not any worse (ie the failures do not indicate actual issues). You can do this automatically using `node generate-testcase.js <testcase-slug>`.

gijsk:
All of these changes have now been merged separately / after fixing test failures. Thank you!

PR #315, opened 2016-11-04
Bug 1173823 - Use base tag's href, r=Gijs

evanxd:
@gijsk ,

Thanks.


Readability.js
```diff
@@ -232,6 +232,25 @@ Readability.prototype = {
...
     var pathBase = this._uri.pathBase;
 
+    // Use `<base>` tag's href if it exists.
+    var bases = this._doc.getElementsByTagName("base");
+    var baseHref = null;
+    if (bases.length) {
+      baseHref = bases[0].getAttribute("href");
+      if (baseHref) {
+        if (baseHref.slice(-1) === "/") {
+          prePath = baseHref.slice(0, -1);
```
gijsk: I think this is wrong. `prePath` is the bit of the URL before the path, including domain but obviously not including the path itself.

The URL in a base tag can be both absolute and relative. If this is an absolute URL with a path, you'll set `prePath` to something including a path. If it's relative and includes a path, you'll still include that path as well.

Generally, doing this resolution in here seems pretty difficult.

Why don't we use document.baseURI when determining the URI to pass to reader mode in the consumer?

test/test-pages/base-url-2/source.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8"/>
+  <base href="https://base.url/"/>
```
gijsk: We should probably have tests with absolute and relative URLs with and without paths.

Readability.js
evanxd: Since we've already fixed the issue in the consumer[1], let's close the pull request.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1173823

PR #316, opened 2016-11-09
Bug 1310075 - Tests for qq.com. r=Gijs

evanxd:
@gijsk ,

I've added tests for [qq.com][qq]. (the `source.html` is from [1])

Thanks.

[qq]: http://www.alexa.com/siteinfo/qq.com
[1]: http://tech.qq.com/a/20161014/006270.htm

### DIFF
--- a/test/test-pages/qq/expected-metadata.json
+++ b/test/test-pages/qq/expected-metadata.json
@@ -1,0 +1,6 @@
+{
+  "title": "DeepMind新电脑已可利用记忆自学 人工智能迈上新台阶_科技_腾讯网",
+  "byline": null,
+  "excerpt": "DeepMind新电脑已可利用记忆自学 人工智能迈上新台阶",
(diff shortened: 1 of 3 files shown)


test/test-pages/qq/expected-metadata.json
```diff
@@ -0,0 +1,6 @@
+{
+  "title": "DeepMind新电脑已可利用记忆自学 人工智能迈上新台阶_科技_腾讯网",
+  "byline": null,
```
gijsk: Should we file a followup issue to detect a byline, or is there no author (clearly/semantically) specified? I'm afraid I can't speak/read Chinese, and Google Translate was not very helpful. Based on its "translation", it looks to me like there's an author at the end of the article（编译/林靖东）but I can't see any specific markup around that, and I don't know if the same information is in some of the metadata elsewhere.

PR #319, opened 2016-11-17
Bug 1255978 - Remove legends candidate, r=Gijs

evanxd:
Hi @gijsk ,

Thanks.


gijsk:
Adding 'text' to the `okMaybeItsACandidate` list seems like it's pretty likely to regress things. After all, 'commentText', 'bannertext', 'modaltext' or 'popuptext' all seem like pretty plausible things that we really don't want.

Are there classes on the ancestors that we could use, and if so would those prevent this node from being culled?

evanxd:
Yes, you're right. If we add `text` here, it means we allow the algorithm to select these nodes we don't want.

Here is the thing, we want to avoid removing the `<div class="text-wrapper" itemprop="articleBody" id="gigya-share-btns-2_gig_containerParent">` node we really want in the first round. It will be removed by the `share` keyword in `unlikelyCandidates` regexp[1] if we don't add a keyword in `okMaybeItsACandidate` regexp.

The `share` keyword lets algorithm removes some nodes we want, kind of like the `text` keyword `unlikelyCandidates` regexp lets it select some nodes we don't want.

Then I would say using more specific keyword here might helpful here. Let's add `text-wrapper` keyword into `okMaybeItsACandidate` regexp to avoid removing the node `<div class="text-wrapper" itemprop="articleBody" id="gigya-share-btns-2_gig_containerParent">` we want. I also updated patch for the the idea. What do you think?

gijsk:
My question was whether there are more meaningful classes/ids on the ancestors of this element that we can use. If we don't currently use those to make decisions about keeping elements despite matching `unlikelyCandidates`, perhaps we need to start doing that.

I am uncomfortable adding `text-wrapper` as an `okMaybeItsACandidate` because the class name does not semantically indicate "this is main article text" to me - it could be wrapping just about any text.

evanxd:
Yes, there is a ancestor meaningful, `<div class="main-content-column">`, extracted from below DOM structure.
```
<div class="main-content-column">
    <div class="text-wrapper" itemprop="articleBody" id="gigya-share-btns-2_gig_containerParent">
```html

I tried to do the checking ancestors idea. Then we got some regressions. Some of them are real regressions, some are not. Let's discuss real one first, e.g. the [social-buttons test][social-buttons], the algorithm will not remove the `<div class="sharedaddy sd-block sd-like" id="like-post-wrapper-70135762-15074-552441e572442">` node because its ancestor(`<article class="main-content">`) is meaningful. I'm not sure is it acceptable or not? We might need to continue discuss other fake/real regression if we want to keep using checking ancestors idea.

For adding `text-wrapper` thing, I would say your option `it could be wrapping just about any text.` is not wrong. But Currently, `article` in `okMaybeItsACandidate` regexp could also be any kinds of text, e.g. `<article>` for comments or others. So maybe adding `text-wrapper` into `okMaybeItsACandidate`(it's maybe :P) regexp might be wrong idea here?

The ideas we mentioned until now might be a little bit tricky. Maybe there might be no 100% good solution for this situation. Looks like we might need to do some compromises here. For me, I prefer adding `text-wrapper` idea because it seems not change the article result too much and it's a simple solution. But since you have a lot of experience on Reader Mode, I do believe your judgement could be better for this tricky situation.

What do you think?

[social-buttons]: https://github.com/mozilla/readability/tree/master/test/test-pages/social-buttons

File: Readability.js [hunk 1]
```diff
@@ -680,6 +688,7 @@ Readability.prototype = {
         if (stripUnlikelyCandidates) {
           if (this.REGEXPS.unlikelyCandidates.test(matchString) &&
               !this.REGEXPS.okMaybeItsACandidate.test(matchString) &&
+              !this._isMaybeAncestorACandidate(node, 3) &&
```
evanxd: I use 3 here because we use the 3 number in a other place, `var ancestors = this._getNodeAncestors(elementToScore, 3);`[1]. Seems 3 level depth means something for ancestors, and I'm also wondering why 3 level depth for ancestors?

[1]: https://github.com/mozilla/readability/blob/master/Readability.js#L741

gijsk: I think this is just set to 3 because it always was and attempts to score all the ancestors with incremental backoff produced regressions. Magical constants are sad, but I don't know that we can easily remove them here.

gijsk: That said, I don't know if 3 is the right depth here... but it can be a starting point, and we can always expand it if we think it's necessary.

evanxd:
Good idea, @gijsk . Let's do it and don't use ancestor logic.

I've updated patch for the idea and CI is good. Please review it again, thanks.

Readability.js
```diff
@@ -1689,6 +1695,26 @@ Readability.prototype = {
...
 
   /**
+   * Clean out elements whose id/class combinations match specific string.
+   *
+   * @param Element
+   * @param RegExp match id/class combination.
+   * @return void
+   **/
+  _cleanMatchedNodes: function(e, regex) {
+    var next = this._getNextNode(e);
```
gijsk: This is a bit confusing. Earlier you used `_forEachNode` and you already end up calling this for every child of `articleContent`. But `_getNextNode` is used to iterate over the entire document. So now you end up iterating over the entire document once for each child node - duplicating work.

The correct thing to do would likely be to have a `_cleanMatchedNodes` method that takes a regex and an element, and loops over that node and everything after it in the DOM tree until it moves out of the subtree. Note that the code you've written here also skips the element you're being passed, because you call `_getNextNode` immediately. How about this:

```js
_cleanMatchedNodes: function(root, regex) {
  // The second argument means _getNextNode will return the first node it'd iterate
  // over if it wasn't also iterating over itself and child nodes.
  var endOfSearchMarkerNode = this._getNextNode(root, true);
  // Ignore the articleContent node itself when iterating (ie only remove children)
  var n = this._getNextNode(root);
  while (n && n != endOfSearchMarkerNode) {
    if (regex.test(n.className + " " + n.id)) {
      n = this._removeAndGetNext(n);
    } else {
      n = this._getNextNode(n);
    }
  }
}
```

gijsk: (To be clear, this method was written so you can pass it `articleContent` without iterating over the children with `_forEachNode`. The end marker node is so that it doesn't look for anything after it's finished searching the `articleContent` node, though I'm not entirely sure if that's necessary... does `articleContent` have any parents or siblings at the point where this code runs?)

Readability.js
```diff
@@ -112,7 +112,7 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
-    unlikelyCandidates: /banner|combx|comment|community|disqus|extra|foot|header|menu|modal|related|remark|rss|share|shoutbox|sidebar|skyscraper|sponsor|ad-break|agegate|pagination|pager|popup/i,
+    unlikelyCandidates: /banner|combx|comment|community|disqus|extra|foot|header|menu|modal|related|remark|rss|shoutbox|sidebar|skyscraper|sponsor|ad-break|agegate|pagination|pager|popup|legends/i,
```
evanxd: Sure.

Readability.js
evanxd: Thanks, @gijsk . Pointed out we need `endOfSearchMarkerNode` here.

But even if we implement `cleanMatchedNodes` like above you mentioned[1], we still need pass all children of `articleContent` node to `cleanMatchedNodes` method to ensure we won't remove any `topCandidate` node, or in this case/bug we will still remove the `<div class="text-wrapper" itemprop="articleBody" id="gigya-share-btns-2_gig_containerParent">` node which is meaningful(we want to keep it here). Does it make sense?

I updated patch for updating `cleanMatchedNodes` method and resort the list. Could you take a look again? Thanks.

[1]: https://github.com/mozilla/readability/pull/319#discussion_r89467125

gijsk:
Ahh, right, this makes more sense now. Thanks for clarifying!

PR #320, opened 2016-11-18
Bug 1173548 - Find out text direction from ancestors of final candidate, r=Gijs

evanxd:
@gijsk ,

Thanks.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -655,9 +655,6 @@ Readability.prototype = {
 
     var pageCacheHtml = page.innerHTML;
 
-    // Check if any "dir" is set on the toplevel document element
-    this._articleDir = doc.documentElement.getAttribute("dir");
-
     while (true) {
       var stripUnlikelyCandidates = this._flagIsActive(this.FLAG_STRIP_UNLIKELYS);
 
@@ -805,6 +802,7 @@ Readability.prototype = {
 
       var topCandidate = topCandidates[0] || null;
       var neededToCreateTopCandidate = false;
+      var parentOfTopCandidate;
 
       // If we still have no top candidate, just use the body as a last resort.
       // We also have to copy the body node so it is something we can modify.
@@ -831,7 +829,7 @@ Readability.prototype = {
         // lurking in other places that we want to unify in. The sibling stuff
         // below does some of that - but only if we've looked high enough up the DOM
         // tree.
-        var parentOfTopCandidate = topCandidate.parentNode;
+        parentOfTopCandidate = topCandidate.parentNode;
         var lastScore = topCandidate.readability.contentScore;
         // The scores shouldn't get too low.
         var scoreThreshold = lastScore / 3;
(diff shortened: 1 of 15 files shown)


gijsk:
This doesn't look right. We should instead be replacing the existing setting of `_articleContent` in `_grabArticle` with this more elaborate scheme. If the location where that happens is not the right place because we don't have the final candidate yet, we should remove that line and do the selection later. Then you can also remove the specialcasing for `HTML`, and simplify the code.

Your modifications also miss out the block:

```
 if (topCandidate === null || topCandidate.tagName === "BODY") {
```

so we won't get the directionality off the `body` tag there.

Either way, this change should not result in changes to any pages under test which do specify the direction unless those are wrong - ie, why do we need to remove the direction from the metadata of the bbc article?

evanxd:
Sorry, I didn't understand what you said `We should instead be replacing the existing setting of _articleContent in _grabArticle with this more elaborate scheme. `. What existing setting of _articleContent?

But you're right, I didn't select final candidate in previous patch. So I updated patch to get final candidate to get the text direction. What do you think of it, @gijsk ?

gijsk:
We already set the article content dir here: https://github.com/mozilla/readability/blob/master/Readability.js#L658-L659

We should be replacing that. If we can't do it in that place, we should still be removing this code (which then means you can remove the "HTML" specialcase in your current checks). Does that make more sense? If not, ping me on IRC tomorrow? That will let us clarify more quickly, I expect.

evanxd:
Sure, I've updated patch to replace that and remove the "HTML" special case. Could you help review it again, thanks.

test/test-pages/bbc-1/expected-metadata.json
```diff
@@ -1,7 +1,6 @@
 {
   "title": "Obama admits US gun laws are his 'biggest frustration' - BBC News",
   "byline": null,
-  "dir": "ltr",
```
evanxd: Remove this `dir` for two reason.

1. This `dir` should not be existed before because we don't do text direction test before the patch.
2. If we want to keep this, the value of it would be `null` not `ltr`. Because there is no `dir` attribute in the `bbc-1` test page.

Readability.js
```diff
@@ -961,6 +961,18 @@ Readability.prototype = {
           return null;
         }
       } else {
+        // Find out text direction from ancestors of final top candidate.
+        var ancestors = [parentOfTopCandidate].concat(this._getNodeAncestors(parentOfTopCandidate));
```
evanxd: I would say we don't need to do that.

If `topCandidate` has `dir` attribute, we will just see the result generated by Reader Mode algorithm with correct text direction. For example, the result could be
```html
    <div dir="rtl"> <!-- The Top  Candidate-->
```

And also notice that at that time, all `topCandidate` nodes are already children of `articleContent` node[1]. If we want to check the `dir` of `topCandidate`, it means we check `articleContent` element for getting text direction. We might not need to do that.

What do you think?

[1]: https://github.com/mozilla/readability/blob/master/Readability.js#L906

gijsk: We only check the things explicitly in the array, right? So if we just concat `[topCandidate, parentOfTopCandidate]` with the ancestors of the `parentOfTopCandidate`, I don't think we check anything else?

While it's true that we would do the correct things in terms of display of the article itself, that wouldn't be the case for the article title and byline (which are likely also rtl in cases like this) because the metadata on our readability detection object would still be wrong.

evanxd: Got it. I've updated patch and added tests for it. Please review the patch, thanks.

gijsk:
LGTM!

PR #323, opened 2016-12-09
Bug 1322327 - Keep single image in div, only remove imgs who are not descendants of a figure if there are at least two imgs

Archaeopteryx:
1. Regenerated the tests which would change for a yet unchanged Readability.js to get more readable diffs for the test changes required by the Readability.js change later.
2. Testcase 'blogger' is not readable according to the generated metadata (even for Readability unchanged), but the test assumes it is, so I manually changed it to true.
3. As you can see, the Readability.js change required test changes for the intended scenario, so I didn't add an extra test.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1706,8 +1706,7 @@ Readability.prototype = {
         var contentLength = this._getInnerText(node).length;
 
         var haveToRemove =
-          // Make an exception for elements with no p's and exactly 1 img.
-          (img > p && !this._hasAncestorTag(node, "figure")) ||
+          (img > 1 && img > p && !this._hasAncestorTag(node, "figure")) ||
           (!isList && li > p) ||
           (input > Math.floor(p/3)) ||
           (!isList && contentLength < 25 && (img === 0 || img > 2)) ||
--- a/test/test-pages/blogger/expected.html
+++ b/test/test-pages/blogger/expected.html
@@ -1,65 +1,74 @@
 <div id="readability-page-1" class="page">
     <div class="post-body entry-content" id="post-body-932306423056216142" itemprop="description articleBody">
-        <p style="display: inline;" class="readability-styled">
-            I've written a couple of posts in the past few months but they were all for </p><a href="http://blog.ioactive.com/search/label/Andrew%20Zonenberg">the blog at work</a>
+        <p style="display: inline;" class="readability-styled"> I've written a couple of posts in the past few months but they were all for </p><a href="http://blog.ioactive.com/search/label/Andrew%20Zonenberg">the blog at work</a>
         <p style="display: inline;" class="readability-styled"> so I figured I'm long overdue for one on Silicon Exposed.</p>
         <p>
-
-            <h2>
-                So what's a GreenPak?</h2>
+            <h2> So what's a GreenPak?</h2>
         </p>
         <p style="display: inline;" class="readability-styled"> Silego Technology is a fabless semiconductor company located in the SF Bay area, which makes (among other things) a line of programmable logic devices known as GreenPak. Their </p><a href="http://www.silego.com/products/greenpak5.html">5th generation parts</a>
         <p style="display: inline;" class="readability-styled"> were just announced, but I started this project before that happened so I'm still targeting the </p><a href="http://www.silego.com/products/greenpak4.html">4th generation</a>
         <p style="display: inline;" class="readability-styled">.</p>
+        <p> GreenPak devices are kind of like itty bitty <a href="http://www.cypress.com/products/32-bit-arm-cortex-m-psoc">PSoCs</a> - they have a mixed signal fabric with an ADC, DACs, comparators, voltage references, plus a digital LUT/FF fabric and some typical digital MCU peripherals like counters and oscillators (but no CPU).</p>
+        <p> It's actually an interesting architecture - FPGAs (including some devices marketed as CPLDs) are a 2D array of LUTs connected via wires to adjacent cells, and true (product term) CPLDs are a star topology of AND-OR arrays connected by a crossbar. GreenPak, on the other hand, is a star topology of LUTs, flipflops, and analog/digital hard IP connected to a crossbar.</p>
+        <p> Without further ado, here's a block diagram showing all the cool stuff you get in the SLG46620V:</p>
         <p>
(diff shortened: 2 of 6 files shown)


gijsk:
Can you rebase/squash things so they're reviewable?

Why reformat the existing tests?

Why was the blogger testcase not failing if it was readerable but not marked as such?

Archaeopteryx:
> Can you rebase/squash things so they're reviewable?

Sure, but imho squashing them will make them less reviewable (see below).

>Why reformat the existing tests?
>Why was the blogger testcase not failing if it was readerable but not marked as such?

I get different output from the test generate script even with the code unaltered. For blogger, the expected-metadata.json must have been edited when it landed: It landed [on Nov 1st](https://github.com/mozilla/readability/commit/4fa0d1b207cd179d4a180f0feeaf6d69447f98a6) but already the [parent commit exports the textContent property](https://github.com/mozilla/readability/blob/4fa0d1b207cd179d4a180f0feeaf6d69447f98a6/Readability.js#L1851) which [didn't get deleted from the data exported to expected-metadata.json](https://github.com/mozilla/readability/blob/4fa0d1b207cd179d4a180f0feeaf6d69447f98a6/test/generate-testcase.js#L131) (which still doesn't happen on master) but the [checked in file doesn't contain textContent](https://github.com/mozilla/readability/blob/4fa0d1b207cd179d4a180f0feeaf6d69447f98a6/test/test-pages/blogger/expected-metadata.json).
readerable is false in the generated metadata and already returns false if re-run on the blogger commit because [the generate script checks myReader.isProbablyReaderable()](https://github.com/mozilla/readability/blob/master/test/generate-testcase.js#L114) but that points (at least here) to the iframe for commenting which doesn't contain much text and so [fails the checks for readerbility](https://github.com/mozilla/readability/blob/master/Readability.js#L1785-L1810).

File issue #324 for the textContent removal and issue #325 for readerable false if iframes without much text are involved.

gijsk:
The first two commits are a no-op when taken together. How does that help? They should just be removed...

I still don't understand why `readerable` is false, and you're saying it should be true, but the tests don't fail either before or after, and your comment hasn't answered that question.

Archaeopteryx:
> The first two commits are a no-op when taken together. How does that help? They should just be removed...

These will disappear once the commit gets stashed. There are just here for the review so it's obvious why exactly _these_ tests get reformatted and they get reformatted both the generated expected output contains less whitespace blocks than the current expected.html. Directly diffing the current file to the one updated for the logic change is harder to read.

> I still don't understand why readerable is false, and you're saying it should be true, but the tests don't fail either before or after, and your comment hasn't answered that question.

Is there any test coverage for the value of readerable? It's false for document with not much content in ps and pres, see #325 

What the log shows:

>     blogger
>       jsdom
>         √ should return a result object
>         √ should extract expected content (46ms)
>         √ should extract expected title
>         √ should extract expected byline
>         √ should extract expected excerpt
>         √ should extract expected direction
>         3) should probably be readerable
> 
>   3) Test pages blogger jsdom should probably be readerable:
> 
>       AssertionError: expected false to deeply equal true
>       + expected - actual
> 
>       -false
>       +true
>       
>       at Assertion.assertEql (node_modules\chai\lib\chai\core\assertions.js:473:10)
>       at Assertion.ctx.(anonymous function) [as eql] (node_modules\chai\lib\chai\utils\addMethod.js:40:25)
>       at Context.<anonymous> (test\test-readability.js:146:43)

gijsk:
> These will disappear once the commit gets stashed. There are just here for the review so it's obvious why exactly these tests get reformatted and they get reformatted both the generated expected output contains less whitespace blocks than the current expected.html. Directly diffing the current file to the one updated for the logic change is harder to read.

I don't understand. How does just having two of these commits make anything "obvious"? I can't see the test results for those commits, nor do they change test pages...

> Is there any test coverage for the value of readerable? It's false for document with not much content in ps and pres, see #325

I don't understand the question. Are you saying you just changed the readerable value for the blogger tests because you felt that was "right", and then it's failing tests? If so, just don't change the value for this bug?

We test isProbablyReaderable - it's done here: https://github.com/mozilla/readability/blob/master/test/test-readability.js#L145 . So all of the tests in the repo are "test coverage".

gijsk:
... and it seems the value on tip is `true`, so the tip code should be thinking the blogger article is readerable. Does that change with this patchset? Why?

gijsk:
Actually, looking at the patches, it almost looks like you re-ran `generate-testcase.js` on existing testcases. Is that right? That is generally not necessary...

Archaeopteryx:
> I don't understand. How does just having two of these commits make anything "obvious"? I can't see the test results for those commits, nor do they change test pages...

The first commit 044670f makes the change to Readability.js. After that, one runs the tests and sees which of the existing tests need updates. But running the generate-testcase.js with the new Readability.js will generate an expected.html with a different formatting (at least here with the dependencies installed): While the current file has e.g. linebreaks after each <p>, the newly generated one prints the text content on the same line like the <p>. So the plan is to regenerate the output for these tests with the old Readability.js. So I revert the change from 044670f in the second commit 9b25324, regenerate the test data for the tests which will change in the third commit 85efd42, finally make my change to Readability.js and update the affected tests in the fourth commit e778c93. Those img nodes new in the expected.html are now the only change to the tests and don't come with all the whitespace and linebreaks removals generate-testcase.js causes.

The fifth commit 2006d67 updates the expected-metadata.json because that file gets generated with readerable: false but it's current version has true and this is caused because isProbableReaderable() gets called after parse(). test-readability.js does this right and calculates readerable: true on the same revision.

> ... and it seems the value on tip is true, so the tip code should be thinking the blogger article is readerable. Does that change with this patchset? Why?

Regenerating it without changes generates it with false. Because the .textContent is also missing from that file, it seems this file got manually edit before it got committed. So using the generate-testcase.js, I get false before and false after my change.

> Actually, looking at the patches, it almost looks like you re-ran generate-testcase.js on existing testcases. Is that right? That is generally not necessary...

This is necessary because the output changes. Currently ignored images (usually between two abstracts now get shown in reader mode while it omitted them before.

gijsk:
Tentative r=me... The annoying thing here is that it's hard to see if this will have additional bad impact on ads, where it seems to me having a single `` without `<p>` and outside a `<figure>` might be relatively common. We'll see, I guess.

PR #328, opened 2016-12-14
Bug 1285543 - Only use "og:title" or "twitter:title" if _getArticleTitle does not return a valid title, r=Gijs

evanxd:
Hi @gijsk ,

Could you help review the patch? Thanks.


gijsk:
Looks good to me. Can you do a separate pull request to normalize spaces in the article title? We shouldn't have things like `\n` or `\t` in it (but that's a separate problem, so separate PR please!).

PR #329, opened 2016-12-14
Bug 1142312 - Add two more types of unlikely candidates: cover-wrap and yom-remote, r=Gijs

evanxd:
Hi @gijsk ,

The patch removes the useless items of current reader mode output of the page[1] caused the issue.

Could you help review it? Thanks.

[1]: https://gma.yahoo.com/veteran-wraps-baby-american-flag-photo-sparks-controversy-175203890--abc-news-topstories.html


gijsk:
This looks OK, but it has conflicts because this is the third yahoo article we're adding. Can you update the PR so it no longer conflicts and adds a third yahoo testcase instead?

evanxd:
Sure, I've updated the patch for it. Once CI is good, let's land the code. Thanks for reviewing.

PR #331, opened 2016-12-14
Clean <input>,<textarea>,<select>,<button> elements

gijsk:
This is also taken out of #313, but now includes two further changes:

1) ensure we don't remove divs inside `<figure>` even if they have only very little content, as this removes e.g. the text in the `<figcaption>` in one of the yahoo testcases.
2) fix tests

@evanxd, can you look at these changes and check they look OK to you?

CC @andrei-ch

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -470,6 +470,7 @@ Readability.prototype = {
 
     // Clean out junk from the article content
     this._cleanConditionally(articleContent, "form");
+    this._cleanConditionally(articleContent, "fieldset");
     this._clean(articleContent, "object");
     this._clean(articleContent, "embed");
     this._clean(articleContent, "h1");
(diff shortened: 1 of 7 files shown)


evanxd:
Looks like we want to clean `<form>` related tags. If so, why don't we also remove `<form>`, `<fieldset>`, and `<label>` tags here?

evanxd:
And we need to fix `yahoo-3` test for this change.

gijsk:
We already have code to remove `<form>` a bit higher up in the file. `<label>` seems like it's likely to be used outside forms (even though it probably shouldn't be). I agree that we can probably remove `<fieldset>`, so I can add support for that.

evanxd:
Sure, let's add `<fieldset>` and merge it.

PR #332, opened 2016-12-14
Use jsdom for parsing the document to determine readability (fixes #325)

### LINKED ISSUE #325: blogger testcase: test run returns readerable: true, but generate-testcase.js returns readerable: false
readerable can be false in the generated metadata because [the generate script checks myReader.isProbablyReaderable()](https://github.com/mozilla/readability/blob/master/test/generate-testcase.js#L114) but that points (at least here) to the iframe for commenting which doesn't contain much text and so [fails the checks for readerbility](https://github.com/mozilla/readability/blob/master/Readability.js#L1785-L1810).

gijsk:
This is a leftover from #310, where we changed `isProbablyReaderable` to check for `<br>` tags. That check only works with jsdom documents because JSDOMParser does not support `querySelectorAll`. This is OK in production because we use real DOMs to check `isProbablyReaderable`. Unfortunately we forgot to update `generate-testcase.js`, so now it produces the wrong output.

@evanxd, does this look OK?

CC @Archaeopteryx

### DIFF
--- a/test/generate-testcase.js
+++ b/test/generate-testcase.js
@@ -111,8 +111,21 @@ function runReadability(source, destPath, metadataDestPath) {
   var myReader, result, readerable;
   try {
     myReader = new Readability(uri, doc);
-    readerable = myReader.isProbablyReaderable();
     result = myReader.parse();
+  } catch (ex) {
+    console.error(ex);
+    ex.stack.forEach(console.log.bind(console));
+  }
+  // Use jsdom for isProbablyReaderable because it supports querySelectorAll
+  try {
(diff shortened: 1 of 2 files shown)


gijsk:
Note that we can't use the jsdom DOM to write out the result content, because it omits closing `/` for `<br>` tags, breaking JSDOMParser when we try to use it to parse stuff. Hence using two different DOM parsers :|

evanxd:
Looks good to me. Let's merge it.

evanxd:
Thanks for the fix.

PR #337, opened 2017-01-13
Fix Issue 333 And Bug 1300697: Some New York Times articles are missing first few paragraphs.

### LINKED ISSUE #333: New York Times article is missing the first few paragraphs.
http://www.nytimes.com/2016/12/18/books/review-debriefing-the-president-tears-into-the-cia.html

evanxd:
Fixes #333 and [Bug 1300697][bug-1300697].

cc @andrei-ch 

[bug-1300697]: http://bugzil.la/1300697


gijsk:
As I noted on bug 1300697, I think we should also investigate taking the next commit from @andrei-ch to remove completely empty `<div>` and `<header>` (and maybe other...) elements before doing sibling joining logic.

Readability.js
```diff
@@ -112,7 +112,7 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
-    unlikelyCandidates: /banner|breadcrumbs|combx|comment|community|cover-wrap|disqus|extra|foot|header|legends|menu|modal|related|remark|replies|rss|shoutbox|sidebar|skyscraper|social|sponsor|ad-break|agegate|pagination|pager|popup|yom-remote/i,
+    unlikelyCandidates: /banner|breadcrumbs|combx|comment|community|cover-wrap|disqus|extra|foot|header|legends|menu|modal|related|remark|replies|rss|shoutbox|sidebar|skyscraper|social|sponsor|supplemental|ad-break|agegate|pagination|pager|popup|yom-remote/i,
```
evanxd: Add `supplemental` to remove `<div id="supplemental-1">` and `<div id="supplemental-2">` nodes from below dom structure. Then the `while (parentOfTopCandidate.tagName != "BODY" && parentOfTopCandidate.children.length == 1)` [loop][loop] will work to fix the nytimes issue.

```html
<article id="story" class="story theme-main   ">

    <div id="TragedyAd" class="ad tragedy-ad nocontent robots-nocontent"></div>

    <div class="dealbook-branding"></div>

    <header id="story-header" class="story-header"></header>

    <div class="story-body-supplemental">
        <div class="story-body story-body-1">
			<p></p>
			<p></p>
			<p></p>
			<p></p>
        </div>
        <div class="supplemental first short" id="supplemental-1" data-between-flex-ads="true" data-pre-height="673" data-max-items="1" data-remaining="673" data-minimum="400" data-last-item-height="873" data-flex-ad-adjacency="true" data-post-height="673"></div>
    </div>

    <div class="story-interrupter" id="story-continues-1"></div>

    <div class="story-body-supplemental">
        <div class="story-body story-body-2">
			<p></p>
			<p></p>
			<p></p>
			<p></p>
        </div>
        <div class="supplemental " id="supplemental-2" data-pre-height="2132" data-max-items="2" data-remaining="242" data-minimum="400" data-last-item-height="945" data-flex-ad-adjacency="false" data-post-height="2132"></div>
    </div>

    <div class="reader-satisfaction-survey prompt feedback-prompt story-content hidden"></div>

    <div id="storage-drawer" class="hidden"></div>
</article>
```
[loop]: https://github.com/mozilla/readability/pull/313/commits/8ec087f0ef72585ae03a63249868eb3b2689a9d8

evanxd:
@gijsk 

I did `cherry-pick` the these two patch[1], [2] in PR 313 and added code to remove empty nodes. Then this pull request could fix the nytimes issue but it also cause other tests failed(I already fixed some in the PR).

The main reason of the test failures is that reader mode algorithm will choose parent/ancestor of previous `topCandidate` as new `topCandidate` if ["the top candidate is the only child"][code-reference-1]. For example, in [medium-2 test][medium-2-test], the result will change to
```html
  <div class="section-content">
    <div class="section-inner layoutSingleColumn">
      // content
    </div>
  </div>
</div>
```
from
```html
  <div class="section-inner layoutSingleColumn">
    // content
  </div>
</div>
```

Is this what we want? For some tests, it might choose parent's parent of original `topCandidate` as new `topCandidate`, like [daringfireball-1][daringfireball-1] test. If this is want we want, I'll continue to fix the [4 test failures][test-failures] caused by the patch.

[1]: https://github.com/mozilla/readability/pull/313/commits/8ec087f0ef72585ae03a63249868eb3b2689a9d8
[2]: https://github.com/mozilla/readability/pull/313/commits/310b636e79bea25e2a1f8583e8b13b740610f01b
[code-reference-1]: https://github.com/mozilla/readability/pull/337/files#diff-06d8d22df421dacde90a2268083424abR911
[medium-2-test]: https://github.com/mozilla/readability/pull/337/files#diff-db7156a5ed4aa2bb31453e270d7e4fb0
[daringfireball-1]: https://github.com/mozilla/readability/pull/337/files#diff-1708f54f6b55ee926468ece09e935bb0
[test-failures]: https://travis-ci.org/mozilla/readability/builds/192989282#L1283

gijsk:
I think that's fine, yeah. I looked at the ignore-whitespace version of the diff and the changes generally look sane, except for one thing - what's up with the rtl-4 change in the direction? Am I reading that right in the sense that it's no longer being detected as RTL? That would be an issue...

test/test-pages/rtl-4/expected-metadata.json
```diff
@@ -2,6 +2,6 @@
   "title": "RTL Test",
   "byline": null,
   "excerpt": "Lorem ipsum dolor sit amet.",
-  "dir": "rtl",
+  "dir": null,
```
gijsk: ie this change. :-)

evanxd: I think we maybe could accept it for below two reasons.

1. That is expected result for the [code][code] we added to fix text direction before. The dom structure of new result is:
```
<div>
    <article dir="rtl">
        <p>...</p>
        <p>...</p>
        <p>...</p>
    </article>
</div>
```

So the final `topCandidate` for `rtl-4` test is the `<div>` node. And we check text direction from `topCandidate` and its ancestors. If we want to fix that here, we might need to check the text direction of all child nodes.

2. Even we get `null` direction, the presentation of result is still correct. Because the `<article>` node is still with `dir="rtl"` attribute.

What do you think?

[code]: https://github.com/mozilla/readability/blob/master/Readability.js#L1017-L1028

test/test-pages/mozilla-1/expected.html
```diff
@@ -7,14 +7,7 @@
...
             </div>
         </section>
-        <section id="designed" class="ga-section" data-ga-label="Designed to be redesigned">
-            <div class="container">
-                <div id="designed-copy">
-                    <h2>Designed to <br class="wide-br"/>be redesigned</h2>
-                    <p>Get fast and easy access to the features you use most in the new menu. Open the “Customize” panel to add, move or remove any button you want. Keep your favorite features — add-ons, private browsing, Sync and more — one quick click away.</p><img class="js " src="" data-processed="false" data-src="//mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned.fbd3ee9402e6.png" data-high-res="true" data-high-res-src="//mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned-high-res.6efd60766484.png" alt="" id="designed-mobile"/></div>
-                <div class="animation-wrapper" id="flexible-bottom-animation"><img src="http://mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/animations/flexible-bottom-fallback.cafd48a3d0a4.png" class="fallback" alt=""/></div>
-            </div>
```
evanxd: Honestly, I cannot understand why adding [`childNode.textContent.trim().length > 0`][change] condition cause it remove above `<div class="container">` node? Do you have any idea, or it is acceptable we remove it?

[change]: https://github.com/mozilla/readability/pull/337/files#diff-06d8d22df421dacde90a2268083424abR739

gijsk: No, it's not really acceptable. I don't know off-hand why this is getting removed... You could add logging and tweak the debug flag to get the logging to show up and then only run this test (`mocha test/test-readability.js -g mozilla`) or the testcase generation script, to see why it does this thing.

Oh, maybe I do know.... before this change, was all the whitespace between nodes being converted to `<p>` tags which then added weighting to the container which led to it being kept, given that it also has images and we have some logic about removing image containers without too much text and so on?

evanxd:
```
Oh, maybe I do know.... before this change, was all the whitespace between nodes being converted to <p> tags which then added weighting to the container which led to it being kept, given that it also has images and we have some logic about removing image containers without too much text and so on?
```
Thanks a lot. You're right. I've updated patch to fix it. How about changing to [`(img > 1 && p / img < 0.5 && !this._hasAncestorTag(node, "figure"))`][code] from `(img > 1 && img > p && !this._hasAncestorTag(node, "figure"))`, which means number of `` node need to be higher enough than `p` node?

[code]: https://github.com/mozilla/readability/pull/337/files#diff-06d8d22df421dacde90a2268083424abR1766

Readability.js
```diff
@@ -1737,7 +1763,7 @@ Readability.prototype = {
         var contentLength = this._getInnerText(node).length;
 
         var haveToRemove =
-          (img > 1 && img > p && !this._hasAncestorTag(node, "figure")) ||
+          (img > 1 && p / img < 0.5 && !this._hasAncestorTag(node, "figure")) ||
```
evanxd:
```
Oh, maybe I do know.... before this change, was all the whitespace between nodes being converted to <p> tags which then added weighting to the container which led to it being kept, given that it also has images and we have some logic about removing image containers without too much text and so on?
```
Thanks a lot. You're right. I've updated patch to fix it. How about this, which means number of `` node need to be bigger enough than `p` node?

test/test-pages/rtl-4/expected-metadata.json
gijsk: We also use the detected direction (ie in the metadata we extract) for the markup for the title, subtitle etc., right? So it makes a difference there. I think we should probably post-process the final readability result and de-nest any singly-nested container nodes. That would also tidy up the markup, and we could preserve any directionality when doing this. Let's do this in a followup as well, though. I'll file a separate issue for this.

Readability.js
gijsk: I think it will do for now, though I also notice that some of the `` tags in the source don't have a `src` attribute. We should probably remove those images and not count them here, which would (I think?) also fix this with a bit less risk to regressing something else for which we might not have a testcase. But let's do that in a followup, this set of commits is confusing enough as it is.

PR #338, opened 2017-01-13
Bug 1330931 - Add tests for nytimes.com, r=Gijs

evanxd:
@gijsk 

I've add tests for nytimes.com. Could you take a look? Thanks.

The article only have one `<div class="story-body-supplemental'">` node. Generally, articles on nytimes.com seems to have two `<div class="story-body-supplemental'">` nodes. We can have at least two tests to test nytime articles. We will add second test on http://bugzil.la/1300697.


gijsk:
Sorry, this seems to have gotten lost. Do the tests still pass after landing #337 ? @evanxd

evanxd:
Yes, I think we still need this test here.

Because the the [`source.html`][source-html] in the test has different dom structure from #337. The dom structure here has only one `<div class="story-body-supplemental'">` node, but dom structure of #337 has two `<div class="story-body-supplemental'">` nodes.

We can handle one `<div class="story-body-supplemental'">` node situation before landing #337 but cannot with two `<div class="story-body-supplemental'">` nodes. After landing #337, we should need to ensure our algorithm can handle the two cases at same time.

[source-html]: https://github.com/mozilla/readability/pull/338/files#diff-62d992990cf0b3c2dba63bf01373ed92

gijsk:
Alright, let's take this too, then.

PR #339, opened 2017-01-16
Fix tiny mistake (typeof dump !== 'undefined')

Treora:


gijsk:
Thank you. Could you use double quotes, and also fix the same issue on line 83 (~5-6 lines below)?

Treora:
> Thank you. Could you use double quotes, and also fix the same issue on line 83 (~5-6 lines below)?

Sure, done.

PR #340, opened 2017-01-17
CommonJS support: set `module.exports = Readability`

Treora:
This seems all that is needed to get this library working with e.g. browserify (useful for browser extensions). Also the workaround in `index.js` would not be necessary if the object is exported this way.

Is there a reason why the object is not exported like so?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1904,3 +1904,7 @@ Readability.prototype = {
     };
   }
 };
+
+if (typeof module === 'object') {
+  module.exports = Readability;
+}


gijsk:
No good reason as long as we include the if statement, as otherwise it'll break being loaded as a simple JS file. Doesn't look like it does, so I'll just merge this.

I'm not sure about removing the hacks in index.js off-hand - I would imagine we'd need to convert the JSDOMParser as well, and we'd need to ensure Readability.js no longer assumes it has a `Node` global variable (and potentially others, I haven't checked in detail).

Treora:
Cool, thanks.

Removing workarounds from `index.js` would indeed require a similar change in JSDOMParser. I do not understand the internals well enough, but perhaps it would also make sense to turn JSDOMParser into a separate module. Readability itself does not seem to require it. A user desiring to use Readability on a NodeJS backend would simply have to include it separately (or use any other DOM 'emulator') to convert their html string to a Document object.

A quick check shows that `Readability.js` uses only `Node.TEXT_NODE` and `Node.ELEMENT_NODE` from the globals that `JSDOMParser.js` sets. It would be easy to just set those manually if they are undefined, so that a global polyfill would not be required.

gijsk:
I don't think we really have a strong interest in maintaining JSDOMParser separately, so splitting it off to a separate module seems like it won't happen. It is very barebones and only really deals with HTML serialized to valid XML (!) which means it breaks on most real-world input. Node consumers should bring their own DOM thing (likely to be jsdom, which is pretty reasonable).

I filed #346 to fix the remaining issues here though, because making this code easier to use/understand from node seems valuable, as does tidying up the hacks. I guess in addition to that we should probably use jsdom for some of the examples in the readme... I'll add a comment to that effect in #346.

Issue #281, opened 2016-03-28
Missing opening paragraph on CNN articles


leibovic:
Articles on CNN frequently have an opening paragraph that is styled differently than others. Reader mode does not include this lead paragraph in its view.

See, for example, http://money.cnn.com/2016/02/01/news/economy/poverty-inequality-united-states/index.html

gijsk:
This is because they use an `<h2>` for that text, with no classes or anything else, and Readability then does:

https://github.com/mozilla/readability/blob/master/Readability.js#L446-L449
```javascript
    // If there is only one h2, they are probably using it as a header
    // and not a subheader, so remove it since we already have a header.
    if (articleContent.getElementsByTagName('h2').length === 1)
      this._clean(articleContent, "h2");
```

I'm not really sure what to do about this... There doesn't seem to be any logic to how the h2 is being used - it really ought not to be a header in the first place. It's bad for accessibility and doesn't add anything to the structure of the page. They should just use a `<p>` and a class to style it how they want to. That said, I don't know that we'll be able to get them to fix that...

evanxd:
Fixed by #347 .

PR #347, opened 2017-01-25
Bug 1259763 - Reader mode omits opening paragraph on CNN articles, r=Gijs

evanxd:
Added test and found out somehow the `<h2>The U.S. has long been heralded as a land of opportunity -- a place where anyone can succeed regardless of the economic class they were born into.</h2>` node in the `source.html` is just removed by some kind of reason. Continue to investigate the root cause...

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -32,6 +32,7 @@ function Readability(uri, doc, options) {
   this._uri = uri;
   this._doc = doc;
   this._biggestFrame = false;
+  this._articleTitle = null;
   this._articleByline = null;
   this._articleDir = null;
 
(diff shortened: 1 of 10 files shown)


gijsk:
See https://github.com/mozilla/readability/issues/281#issuecomment-208893497 .

I expect we should avoid removing the `<h2>` if it is substantially different from the title of the page that we detect.

evanxd:
@gijsk 
I updated patch for the comments. Could you take a look again? Thanks.

test/test-pages/medium-1/expected.html
```diff
@@ -1,5 +1,6 @@
 <div id="readability-page-1" class="page">
     <div class="section-inner layoutSingleColumn">
+        <h2 name="3c62" id="3c62" data-align="center" class="graf--h2">Open Journalism Project:</h2>
         <h4 name="425a" id="425a" data-align="center" class="graf--h4"><em class="markup--em markup--h4-em">Better Student Journalism</em></h4>
```
gijsk: Funny, here we should really be taking out the `h4` *and* the `h2`. Anyway, we're not making things worse, so...

test/test-pages/tmz-1/expected.html
```diff
@@ -1,6 +1,7 @@
 <div id="readability-page-1" class="page">
     <div class="post single-post" id="post-2015_02_26_lupita-nyongo-pearl-dress-stolen-oscars">
         <p class="headline">
+            <h2 class="hf1">Lupita Nyong'o</h2>
```
gijsk: Maybe file a followup to remove things with class/id containing `headline`?

In fact, this `h2` is clearly a substring of the title in the metadata, so why isn't it being removed?

Readability.js
```diff
@@ -482,10 +483,18 @@ Readability.prototype = {
       this._cleanMatchedNodes(topCandidate, /share/);
     });
 
-    // If there is only one h2, they are probably using it as a header
-    // and not a subheader, so remove it since we already have a header.
-    if (articleContent.getElementsByTagName('h2').length === 1)
-      this._clean(articleContent, "h2");
+    // If there is only one h2 and its text content substantially equals article title,
+    // they are probably using it as a header and not a subheader,
+    // so remove it since we already have a header.
```
gijsk: Nit: ... since we already extract the title separately.

evanxd:
Thanks for reviewing. :)

gijsk:
Can you file the followup to remove the `.headline` things, @evanxd ? :-)

evanxd:
@gijsk , sure.
Sorry for replying lately. I was in vacation at that time.

evanxd:
The followup issue: https://github.com/mozilla/readability/issues/349

PR #348, opened 2017-01-30
Fix bug 1322674 by only conditionally removing content from tables when the tables are layout tables

gijsk:
@evanxd I wrote up a fix for [bug 1322674](https://bugzilla.mozilla.org/show_bug.cgi?id=1322674). I also added some extra debugging logic to the test runner script. Can you take a look?

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -468,6 +468,11 @@ Readability.prototype = {
   _prepArticle: function(articleContent) {
     this._cleanStyles(articleContent);
 
+    // Check for data tables before we continue, to avoid removing items in
+    // those tables, which will often be isolated even though they're
+    // visually linked to other content-ful elements (text, images, etc.).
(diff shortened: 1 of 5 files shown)


evanxd:
Looks good to me.
The logging code to easily find the error nodes is a good idea!

Readability.js
```diff
@@ -1723,7 +1825,7 @@ Readability.prototype = {
         // ominous signs, remove the element.
         var p = node.getElementsByTagName("p").length;
         var img = node.getElementsByTagName("img").length;
-        var li = node.getElementsByTagName("li").length-100;
+        var li = node.getElementsByTagName("li").length - 100;
```
evanxd: Let's add new eslint rules to avoid the bad code style like this. Already filed a follow up [issue][issue].

[issue]: https://github.com/mozilla/readability/issues/351

PR #350, opened 2017-02-06
Issue 349 - Remove elements with headline class

### LINKED ISSUE #349: Remove elements with headline class
Remove elements with `headline` class.

evanxd:
Fixes #349.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -113,7 +113,7 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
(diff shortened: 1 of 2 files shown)


evanxd:
After patched, there are three [test failures][test-failures]. For example, in [ehow-1][ehow-1] test, we will remove [`<span class="headline2 head mg-1 block">What You'll Need:</span>`][ehow-1] element we should keep in result. Same things in other two failures.

Maybe we should only remove elements whose classes include `headline`, not like current patch does, remove elements whose id/class combinations match `headline`.

What do you think? @gijsk 

[test-failures]: https://travis-ci.org/mozilla/readability/builds/198778856
[ehow-1]: https://github.com/mozilla/readability/blob/master/test/test-pages/ehow-1/expected.html#L13

gijsk:
OK, then I guess we can't do this. I don't see a good way to remove these elements and not break other cases (like the eHow ones). The web is awful sometimes. :-(

PR #352, opened 2017-02-10
Bug 1338468 - Tests for tumblr.com

evanxd:
Tests for tumblr.com. The test page is http://mcupdate.tumblr.com/post/96439224994/minecraft-18-the-bountiful-update.

@gijsk , could you help review the patch?

Thanks.


test/test-pages/tumblr/expected-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Minecraft 1.8 - The Bountiful Update - Minecraft 1.8 - The Bountiful Update",
```
gijsk: Hm, odd - does tumblr always repeat the title? It looks odd, but I can see this was already in the original page...

test/test-pages/tumblr/expected-metadata.json
```diff
@@ -0,0 +1,7 @@
...
+  "byline": null,
+  "dir": null,
+  "excerpt": "+ Added Granite, Andesite, and Diorite stone blocks, with smooth versions\n+ Added Slime Block\n+ Added Iron Trapdoor\n+ Added Prismarine and Sea Lantern blocks\n+ Added the Ocean Monument\n+ Added Red...",
```
gijsk: Can you file a followup issue so we use the unicode ellipsis character instead of 3 periods when chopping off text like this?

test/test-pages/tumblr/expected-metadata.json
evanxd:
```
does tumblr always repeat the title?
```
In general, no. But that is from the original page. I used the page because the blog is one of the top tumblr blogs.

test/test-pages/tumblr/expected-metadata.json
evanxd: Sure, let's do it.

PR #354, opened 2017-02-17
Bug 1217007 - Remove nodes without any content, r=Gijs

evanxd:
Fixes [bug 1217007][bug-1217007].

Hi @gijsk,

The patch removes nodes without any content(e.g. text, image, video, or iframe). It helps the [logic][logic] added for New York Times pages find the correct top candidate.

Could you help review it? Thanks.

[bug-1217007]: http://bugzil.la/1217007
[logic]: https://github.com/mozilla/readability/blob/master/Readability.js#L925-L931


Readability.js
```diff
@@ -721,11 +721,11 @@ Readability.prototype = {
           }
         }
 
-        // Remove empty DIV, SECTION, and HEADER nodes
+        // Remove DIV, SECTION, and HEADER nodes without any content(e.g. text, image, video, or iframe).
```
gijsk: "e.g." means "for example". I'm not clear on what this code is doing and why it's necessary to add a specialcase for elements that contain no images, videos or iframes and no text (but do contain other children). It feels like what we're really doing is hacking around the problem of nested empty things, so if I have:

```
<div>
<div>
</div>
</div>
```

we might be removing the inner div but not the outer one, or something. It feels like we should just explicitly check for that case (so when we remove a node, check if the parent then has no more children and no text using the existing `_isEmptyElement` function).

Alternatively, it seems we should use the same list the if condition here is using to check whether there are any child elements that *aren't* in this set of nodes about which we don't care.

As a counterexample where your code would make a different decision, it looks like this patch will remove containers that contain `<br>` or `<hr>` elements (which wouldn't affect their parent's textContent), which seems wrong.

evanxd: The root cause of this bug is choosing wrong top candidate(currently, the algorithm select `<div class="section-3">` in the below dom tree as the top candidate).

```html
<div class="postArticle-content">
  <div class="section-1">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-2">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-3">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-4">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-5">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
</div>
```

And the patch is trying to remove the `<div class="section-divider">` node, and it helps the [logic added for nytimes pages][nytimes-logic] select correct top candidate(that is `<div class="postArticle-content">`).

I would say removing containers that ONLY contain `<br>` or `<hr>` might not be a wrong idea. Because it will only remove the (useless?) nodes like below
```
<div><hr /></div>
<div><br /></div>
```
and it won't remove the (meaningful) nodes like below
```
<div>
  <h1>Lorem</h1>
  <hr />
  <p>Tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum</p>
  <br />
</div>
```

Removing the useless `<div class="section-divider">` nodes in the above dom tree might be the easiest approach to fix the issue and it also seems to make sense. If you also think we could take the solution, then we could change and update current implementation in the patch.

What do you think? Or @gijsk you have other good ideas to fix the issue. (I'm think of it)

[nytimes-logic]: https://github.com/mozilla/readability/blob/master/Readability.js#L925-L931

evanxd: The root cause of this bug is choosing wrong top candidate(currently, the algorithm select `<div class="section-content">` that belongs to `<div class="section-3">` in the below dom tree as the top candidate).

```html
<div class="postArticle-content">
  <div class="section-1">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-2">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-3">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-4">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
  <div class="section-5">
    <div class="section-divider"><hr class="section-divider" /></div>
    <div class="section-content"></div>
  </div>
</div>
```

And the patch is trying to remove the `<div class="section-divider">` node that belongs to `<div class="section-3">`, and it helps the [logic added for nytimes pages][nytimes-logic] select correct top candidate(that is `<div class="postArticle-content">`).

I would say removing containers that ONLY contain `<br>` or `<hr>` might not be a wrong idea. Because it will only remove the (useless?) nodes like below
```html
<div><hr /></div>
```
```html
<div><br /></div>
```
and it won't remove the (meaningful) nodes like below
```html
<div>
  <h1>Lorem</h1>
  <hr />
  <p>Tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum</p>
  <br />
</div>
```

Removing the useless `<div class="section-divider">` nodes in the above dom tree is a simple approach to fix the issue and it also seems to make sense. If you also think we could take the solution, then we could change and update current implementation in the patch.

What do you think? Or @gijsk you have other good ideas to fix the issue. (I'm think of it)

[nytimes-logic]: https://github.com/mozilla/readability/blob/master/Readability.js#L925-L931

gijsk: Hm. Without any CSS, an `hr` and `br` have an actual visual effect on the page, though, so we're removing content that might be important to keep the article readable in some cases. If we detect containers with only an hr/br in them, can we reparent the hr/br to the parent of the container and then remove the container, or does that still not help? I guess it might not because I'm not sure the sibling joining logic will kick in at that point...

This is annoying because really, the `hr` and/or `br` belong inside the section that they start/end (depending on the semantics), but that's not where they are in the markup. :-(

If we can't come up with something better I guess we should take this, even if it removes these elements.

Out of curiosity, in pages where this happens, are the hr/br styled to no longer affect the visual appearance of the page? I wonder why they're included at all...

evanxd:
```
 If we detect containers with only an hr/br in them, can we reparent the hr/br to the parent of the container and then remove the container, or does that still not help?
```
No, it doesn't help. The [logic] only works if the top candidate is the only child. I think we should take the solution since we could not come up other better ideas.

```
are the hr/br styled to no longer affect the visual appearance of the page?
```
You can see the `<hr>` node is represented as the three dots on the page in the below screenshot. I would say removing the `<hr>` node here is acceptable.

If it is acceptable, let's merge the patch and move on the next issue. :)

screen shot 2017-02-21 at 4 23 40

[logic]: https://github.com/mozilla/readability/blob/master/Readability.js#L925-L931

gijsk: I'm still wary that we have a list of elements that avoid removal, rather than a list of elements that don't prevent removal. So if we're forgetting about non-text elements that aren't images or video/audio/iframe elements, that *should* trigger keeping the container, this code is buggy. I would prefer it if it treated elements being either empty or containing only br/hr elements as superfluous, rather than the wider approach this patch is taking. Does that make sense?

evanxd: Sure, let's do it. If we find out more superfluous we want to remove in the future, we can list them at that time.

Already updated the patch for it, please take a look, thanks. :)

Readability.js
```diff
@@ -1184,10 +1184,11 @@ Readability.prototype = {
...
   },
 
-  _isEmptyElement: function(node) {
+  _isElementWithoutContent: function(node) {
     return node.nodeType === Node.ELEMENT_NODE &&
-      node.children.length == 0 &&
-      node.textContent.trim().length == 0;
+      node.textContent.trim().length == 0 &&
+      (node.children.length == 0 ||
+       node.children.length == node.getElementsByTagName("br").length + node.getElementsByTagName("hr").length);
```
gijsk: Hm, if this is sufficient I guess that works. In the future we may need to include other purely container elements - though maybe right now that already works? What happens for `<div><div><br></div></div>` ? Do we end up first removing the inner div and then remove the outer div afterwards, or something?

PR #357, opened 2017-02-22
Remove general part of title when viewing websites with hierarchical titles

Farbdose:
I mainly use readability to convert web blogs with hierarchical article titles:
e.g. "Story A > Chapter 1" -> "Chapter 1"
into continuous text (which I then feed into a Text2Speech engine)
Therefore it's kind of annoying if the domain/story name always repeats itself in the title.

In case anyone else considers this useful I thought I just create a pull request if not, that's ok too.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -314,11 +314,11 @@ Readability.prototype = {
         curTitle = origTitle = this._getInnerText(doc.getElementsByTagName('title')[0]);
     } catch (e) {/* ignore exceptions setting the title. */}
 
-    if (curTitle.match(/ [\|\-] /)) {
-      curTitle = origTitle.replace(/(.*)[\|\-] .*/gi, '$1');
+    if (curTitle.match(/ [\|\-\\\/>»] /)) {
+      curTitle = origTitle.replace(/(.*)[\|\-\\\/>»] .*/gi, '$1');
 
       if (curTitle.split(' ').length < 3)
-        curTitle = origTitle.replace(/[^\|\-]*[\|\-](.*)/gi, '$1');
+        curTitle = origTitle.replace(/[^\|\-\\\/>»]*[\|\-\\\/>»](.*)/gi, '$1');
     } else if (curTitle.indexOf(': ') !== -1) {
       // Check if we have an heading containing this exact string, so we
       // could assume it's the full title.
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -346,8 +346,9 @@ Readability.prototype = {
     }
 
     curTitle = curTitle.trim();
-
-    if (curTitle.split(' ').length <= 4)
+    var curTitleLen = curTitle.split(' ').length;
+    if (curTitleLen <=4 && ( !origTitle.match(/ [\\\/>»] /)
+        || curTitleLen != origTitle.replace(/[\|\-\\\/>» ]+/g, " ").split(' ').length -1))
```
gijsk: Thanks for the pull request! Certainly the first two additions to the regexps a bit higher up are fine. Here I'm less sure - can you explain what is going on with the second clause to the if statement?

Otherwise, two small style nits: we should probably use `.test()` rather than `.match()` for the first check here, and we normally have spaces around operators but not inside braces (`()`) (we need to put that in our eslint config if they aren't already, sorry if you ran that and this didn't come up. :-( ).

Farbdose: Certainly. So normally we check in line 350 if our new title got shorter than 4 words and if yes, use the origTitle instead but this only makes sense if the current title isn't a hierarchical one because here we don't have a problem with the title having less than 4 words because it probably means we correctly found the last layer of the hierarchy. 

Of course, this approach is way too broad and would fail some tests so I limited it to only removing the root of the hierarchy.

That's the second part of the if, it boils down to: hierarchical title implies: the new title should only be one word shorter than the original title.

Point of interes is now line 317-321:
```Javascript
if (curTitle.match(/ [\|\-\\\/>»] /)) {
  curTitle = origTitle.replace(/(.*)[\|\-\\\/>»] .*/gi, '$1');

  if (curTitle.split(' ').length < 3)
  curTitle = origTitle.replace(/[^\|\-\\\/>»]*[\|\-\\\/>»](.*)/gi, '$1');
}
```
because in combination it is possible that we now only shorten hierarchical titles with a maximum of two layers which would basically defeat the whole point of all this. But on the other side, this means less interference with the existing behavior. My test use case only contains titles with two layers and I am not sure if there are any with more layers in the wild that would qualify for this kind shortening.

@Your eslint config: no problem, but you could add a line to the test section of the readme to make first-time users aware of eslint (I only found out when my pull request failed the  continuous-integration check)

gijsk: > @ Your eslint config: no problem, but you could add a line to the test section of the readme file to make first-time users aware of eslint (I only found out when my pull request failed the  continuous-integration check)

An excellent point. I updated the "Contributing" and "Tests" sections of the ReadMe.

gijsk:
Thanks! I added a bunch of comments, tweaked two of the regexps and replaces (one in the context of your patch, one new one that you were adding) so hopefully this code is now easier to follow than before.

I'm sure I messed up the 'git' part of doing that, though, as github doesn't seem to be closing this issue. :-\

Anyway, I squashed your changes locally and merged them as:

https://github.com/mozilla/readability/commit/aed7f10594b76f0faaf707fed3931e7c91598b9f

and then added https://github.com/mozilla/readability/commit/a63e63c91ee060873e3db06babd217323ad3979b . Thanks again for helping make Readability nicer!

PR #362, opened 2017-03-09
Bug 1345797 - Tests for yahoo.co.jp, r=Gijs

evanxd:
Tests for yahoo.co.jp(Global Rank 27[1]). The test page is http://headlines.yahoo.co.jp/hl?a=20170309-35097838-cnetj-sci.

Could you take a look? @gijsk 
Thanks.

[1]: http://www.alexa.com/siteinfo/yahoo.co.jp


test/test-pages/yahoo-4/expected-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "トレンドマイクロ、公衆無線LANを安全に使うためのアプリ「フリーWi-Fiプロテクション」 （CNET Japan） - Yahoo!ニュース",
+  "byline": null,
```
gijsk: Can you file issues about the missing byline and about us not stripping the "Yahoo!" bit at the end from the title?

Other than that, this looks OK to me. I'm assuming you checked there aren't major issues with the content we read otherwise.

evanxd: Sure, filed the two bugs, #363 and #364.

PR #370, opened 2017-03-15
Bug 1347472 - Tests for WordPress blogs, r=Gijs

evanxd:
Tests for WordPress blogs(global rank 40). The test page is https://wptavern.com/stack-overflow-jobs-data-shows-reactjs-skills-in-high-demand-wordpress-market-oversaturated-with-developers.


test/test-pages/wordpress/expected-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Stack Overflow Jobs Data Shows ReactJS Skills in High Demand, WordPress Market Oversaturated with Developers – WordPress Tavern",
+  "byline": null,
```
evanxd: Filed a follow up issue #371 to get the right byline.

test/test-pages/wordpress/expected-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Stack Overflow Jobs Data Shows ReactJS Skills in High Demand, WordPress Market Oversaturated with Developers – WordPress Tavern",
```
gijsk: I thought we'd be trying to get rid of the "WordPress Tavern" bit, too. Why isn't that working? Maybe that wants to be another issue?

evanxd: Yes, you're right. Filed the follow-up issue #372.

PR #385, opened 2017-08-09
Remove presentational HTML attributes

### LINKED ISSUE #383: Strip bgcolor attributes
Example: https://www.jwz.org/doc/linuxvideo.html

I use Reader Mode to reformat old pages on my phone, and it usually works well. Some pages like the above, Scott McCloud's, etc. with heavy use of `<td bgcolor>` clash with the Dark theme, becoming illegible:

Screenshot of a page in Firefox's Reader Mode, with white text clashing on a yellow background.

This could be fixed by removing `bgcolor` attributes in Readability.js, but also in the applied stylesheet with `[bgcolor] { background-color: transparent }` or something. Which is preferable?

tigt:
Fixes #383

This patch loops through a list of known-presentational attributes in HTML, attempting to remove each from each cleaned element. (Checking for the attribute's existence first seems to just add needless overhead.)

The extra check for the HTML namespace is to avoid removing attributes that inline SVG needs.


Readability.js
```diff
@@ -1266,23 +1270,24 @@ Readability.prototype = {
...
+
+          if (this.DEPRECATED_SIZE_ATTRIBUTE_ELEMS.indexOf(e.tagName) !== -1) {
+            e.removeAttribute('width');
+            e.removeAttribute('height');
+          }
+        }
       }
 
-      cur = cur.nextSibling;
+      e = e.nextSibling;
```
gijsk: Now we no longer recurse into descendant elements, which doesn't seem right. This also explains why the test you added is now failing as well...

When I complained about the recursion, my issue was that we call `removeAttribute("style")` (and, with your patch, a few extra `removeAttribute` calls) inside the loop, and then we call `_cleanStyles`, which does the same thing again before entering the `while()` loop. You've fixed this, but now we never recurse.

Here's how I think we can fix this:

```js
_cleanStyles: function(e) {
  // remove stuff off `e` here if all the conditions are met
  var cur = e.firstElementChild;
  while (cur) {
    this._cleanStyles(cur);
    cur = cur.nextElementSibling;
  }
}
```

tigt: Ah, thank you very much. I noticed my error before you posted this, but was having a hard time figuring out the right order of operations.

tigt:
Hm. I might need to draw a diagram to really wrap my head around it, because I seem to be getting nowhere with my current mental model.

Readability.js
```diff
@@ -1266,23 +1270,25 @@ Readability.prototype = {
...
-    while (cur !== null) {
-      if (cur.nodeType === cur.ELEMENT_NODE) {
-        // Remove style attribute(s) :
-        if (cur.className !== "readability-styled")
-          cur.removeAttribute("style");
+    if (e.nodeType === e.ELEMENT_NODE &&
+        e.namespaceURI === 'http://www.w3.org/1999/xhtml' &&
+        e.className !== 'readability-styled') {
+      // Remove `style` and deprecated presentational attributes
+      for (var i = 0; i > this.PRESENTATIONAL_ATTRIBUTES.length; i++) {
```
gijsk: It took me a while, but this isn't yet working because this comparison is the wrong way around - it should be `<` not `>`. :-)

Readability.js
```diff
@@ -1266,23 +1270,25 @@ Readability.prototype = {
...
-      e.removeAttribute('style');
-
-    // Go until there are no more child nodes
-    while (cur !== null) {
-      if (cur.nodeType === cur.ELEMENT_NODE) {
-        // Remove style attribute(s) :
-        if (cur.className !== "readability-styled")
-          cur.removeAttribute("style");
+    if (e.nodeType === e.ELEMENT_NODE &&
+        e.namespaceURI === 'http://www.w3.org/1999/xhtml' &&
```
gijsk: Because we're guaranteed element nodes, I don't think you need to check `e.nodeType`.

The namespaceURI check doesn't seem to work in JSDOMParser. I don't think it supports namespaces... But the attributes don't have meaning in SVG ( https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute ) and I don't think we'll have other namespaces mixed in, so I think you can just leave it out - or maybe do:

```js
if ((!e.namespaceURI || e.namespaceURI === "http://www.w3.org/1999/xhtml") &&
    e.className != "readability-styled")
```

tigt: It seems to call it on the `Document`, so the first instance may not be an `Element` proper, maybe?

Inline SVG I'm mostly worried about `style`, since most of SVG's attributes can be set via CSS as well.

tigt: Oh, and certain SVG elements require `width` and `height`, or they don't render at all:
* `<rect>`
* `<image>`
* `<foreignObject>`

Other assume default dimensions if unspecified, which can break rendering:
* `<svg>` (yeah, you can nest them)
* `<filter>`
* `<pattern>`
* `<mask>`
* `<use>`

gijsk: Hm, there's a fallback to `this._doc`, but we pass in `articleContent`, which is never the document, and there's only 1 callsite, so perhaps we should just remove the fallback?

For SVG, I mean, the `width` and `height` would already not be removed because we're only doing that for the elements in a list that you already created. 
The `style` point is well-made, but is an existing bug, right? We could try to refine that later. I don't really have good ideas about that anyway, short of either creating more lists of elements or trying to teach JSDOMParser about namespaces (but really, don't try to do that... namespaces are seriously messed up). The simplest thing might just be to prevent the recursion entirely for `<svg>` elements, and leave it at that. That would miss out HTML-nested-in-SVG, but I think we can live with that... if people put table markup in their SVG, I don't really want to know!

gijsk: For the remaining errors from the test runner, I think we'll need to update the "expected.html" files for the tests that fail, to remove the attributes that we're now removing.

Sorry this is turning into a bigger adventure than you maybe bargained for!

tigt: Yeah, that sounds reasonable. Let's see how that goes.

tigt:
Is there a way to get more information about the match failures? The only test that won't pass is `svg-parsing`, and I can't figure out what the runner expects to be different.

tigt:
Here's my sticky wicket: when I remove the `style` attribute from the `<svg>`, the tests report:

`AssertionError: expected 6 to deeply equal 7`

But if I leave it on, instead:

`AssertionError: expected 7 to deeply equal 6`

I am plumb baffled.

gijsk:
Sorry, I'm fixing the logging for you in #386. I think the issue is likely that all tests run once with jsdom, once with JSDOMParser, and one of the checks in the patch probably succeeds with one and fails with the other. I'll look at the patch in a second.

Readability.js
```diff
@@ -1263,26 +1267,25 @@ Readability.prototype = {
    * @return void
   **/
   _cleanStyles: function(e) {
-    e = e || this._doc;
-    if (!e)
+    if (!e || e.tagName === 'svg')
```
gijsk: It looks like in jsdom, tagName is "svg" and in JSDOMParser it's "SVG".  Lowercase is correct (or at least, that's what gecko does too), but it seems we need to call `.toLowerCase()` to make it work in JSDOMParser. :-(

gijsk:
Wonderful, thank you!

tigt:
Thanks for your patience :)

PR #399, opened 2017-09-19
Allow the word threshold parameter to be configurable

iq-dot:
I have added a simple PR that allows a client to configure what is acceptable for them as a word threshold. Many times I have found myself having to adjust these figures based on the type of articles it is being used on.


gijsk:
Fantastic, thanks!

PR #405, opened 2017-11-13
Added npmignore for test and benchmarks resources

todvora:
Added a `.npmignore` file to this repository, excluding `test` and `benchmarks` from the distribution tarball, where no test and benchmark files are needed. 

The distribution tarball shrinks from 2.7MB down to 30,4KB by excluding 297 files, totalling 16.9MB before compression. This is also the space it saves in the `node_modules` directory of every project depending on this library.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+/benchmarks/
+/test/


gijsk:
Thank you for the PR.

Why does NPM pull in everything by default? That seems pretty dumb...

Also,

> The distribution tarball shrinks from 2.7MB down to 30,4KB by excluding 297 files, totalling 16.9MB before compression. This is also the space it saves in the node_modules directory of every project depending on this library.

How do they do that (ie depending on this "library")? #43 is still open. I'm not aware of the package being available through npm at all...

Separately, wouldn't removing the tests and benchmarks break actually running the tests? (cf. https://docs.npmjs.com/cli/test ) Is there a way to avoid doing that?

todvora:
NPM packages everything in the project minus files excluded in `.gitignore` and `.npmignore`. You can test that behaviour by calling `npm pack`, which creates the tarball.

I provide in my application dependency to this library by utilizing [Github URLs](https://docs.npmjs.com/files/package.json#github-urls), so the dependency name is `mozilla/readability` and is pulled directly from Github. Then I don't need any build published in the npm registry. The same `.npmignore` file is used and it still saves those 16.9MB in my `node_modules`. 

Finally, the tests are not affected anyhow, because they are running inside a cloned git repository, not on the distributed tarball. The tarball is used only by those, who are depending on this library in their software. This is a common pattern, for example in Java, you don't have any tests / test resources included in the distribution jar neither.  You can verify the functionality easily on a Travis-CI build of this PR: https://travis-ci.org/mozilla/readability/builds/301384397

gijsk:
> Finally, the tests are not affected anyhow, because they are running inside a cloned git repository, not on the distributed tarball. The tarball is used only by those, who are depending on this library in their software. This is a common pattern, for example in Java, you don't have any tests / test resources included in the distribution jar neither.  You can verify the functionality easily on a Travis-CI build of this PR: https://travis-ci.org/mozilla/readability/builds/301384397

I understand that the tests we run won't be affected, but my understanding is this change will make it impossible for people who "just" run the package to run the tests, per e.g. https://stackoverflow.com/a/25125870/713326 . Anyway, probably still the right trade-off considering the size of the tests dir at this point...

gijsk:
(but still sad that there's no way to tell npm "these dirs are my test things, pull them in when someone asks to run the tests")

gijsk:
(Thanks again for the PR and the explanations!)

todvora:
Thank you for the merge! 

From my experience - people are usually removing all non-production files, compress resources, compile sources to javascript (from coffeescript, ES6, typescript).... So the distributed tarball is very different from what you see in the git repository. 

Take for example the [Request](https://github.com/request/request) package, on of most depend-on packages. They are including only a handful of files in the tarball: https://github.com/request/request/blob/master/package.json#L24 ([files](https://docs.npmjs.com/files/package.json#files) is the opposite of `.npmignore`, it says which files to include). Everything else is excluded and available in the git repository only. 

I believe this helps the majority of users and does not add any troubles to others.

PR #406, opened 2017-11-16
Clean IDs and classes from output.

heycam:
I suspect I need to re-generate test expectations.  I did `apt-get install mocha` but when running tests I get:

```$ mocha test/test-*.js
module.js:328
    throw err;
    ^

Error: Cannot find module 'chai'
...
```

and I'm not sure what to do from there.

/cc @gijsk


Readability.js
```diff
@@ -226,6 +234,30 @@ Readability.prototype = {
...
+    }
+
+    if (node.id && this.IDS_TO_PRESERVE.indexOf(node.id) == -1) {
+      node.removeAttribute("id");
+    }
+    if (node.className && this.CLASSES_TO_PRESERVE.indexOf(node.className) == -1) {
+      node.removeAttribute("class");
+    }
+
+    for (node = node.firstChild; node; node = node.nextSibling) {
```
gijsk: Nit: it would be more efficient to use `firstElementChild` and `nextElementSibling` and then omit the `nodeType` check. :-)

gijsk:
I think you'd want `npm install -g mocha` and/or `npm install chai` (either with or without the `-g` switch) to get the test stuff setup locally. If that doesn't trivially make things work, I can have a look at regenerating the testcases, and/or you could try to remove the class/id attributes with a global search/replace over the files...

heycam:
I failed at getting the tests running locally, so I've removed the `class=""` and `id=""` attributes from the test expectations.  Hopefully it's right or close to right.

gijsk:
I'm so sorry for the slow here. This patch is basically exactly what I originally thought we wanted, but then I realized we actually do have some rules that involve content classes ( https://searchfox.org/mozilla-central/source/toolkit/themes/shared/aboutReaderContent.css#171-178 and some `.wp-caption` stuff in that file), and that we might want more (cf. https://bugzilla.mozilla.org/show_bug.cgi?id=1399616 ). So this needs some more work to effectively have a list of allowed classes and to filter out other stuff. I was planning on just merging this thing and then writing a separate PR to address that, but I haven't had time so far. Leaving a note for now.

If you're busy, no worries and I'll do my best to get back to this later this week. If this is blocking you, please ping me and I'll try to up-prio. Alternatively, you can have a look at the links and adjust the patch accordingly. I feel bad though, and figured I should fix the crap I technically own myself, because you've already jumped through too many hoops / exploratory stuff for the iframe sandboxing, and this stuff isn't going to be the easiest either, because the synthetic JS-based DOM we have in reader mode doesn't have `classList` and we need to filter classes. Bah, software is awful, let's go shopping...

heycam:
That's fine, I can make the Readability constructor take a list of class names to preserve in the options object.

gijsk:
<3

Thank you very much!

Issue #407, opened 2017-11-17
Telegraph articles failing


NinoSkopac:
Repro URL: http://www.telegraph.co.uk/news/2017/11/16/zimbabwes-robert-mugabe-wife-grace-insisting-finishes-term-priest/

Problem: missing paragraphs from the article

Repro code:
```js
var read = require('node-readability');

read('http://www.telegraph.co.uk/news/2017/11/16/zimbabwes-robert-mugabe-wife-grace-insisting-finishes-term-priest/', function(err, article, meta) {

    // Main Article
    console.log(article.textBody);

    // Close article to clean up jsdom and prevent leaks
    article.close();
});
```

andreskrey:
After some investigation, the missing paragraph is the following one:

```
<div class="component-content">
<p><span class="m_first-letter m_first-letter--flagged">Z</span>imbabwean intelligence reports seen by Reuters suggest that former security chief Emmerson Mnangagwa, who was ousted as vice-president this month, has been mapping out a post-Mugabe vision with the military and opposition for more than a year.</p>
</div>
```

The `<span>` surrounding the Z is skipping the [hasSinglePInsideElement](https://github.com/mozilla/readability/blob/master/Readability.js#L756) function and the [conversion to a P node](https://github.com/mozilla/readability/blob/master/Readability.js#L766). This leaves the node with 0 score and never gets added to the article when scanning siblings of the topCandidate.

Not sure what a possible fix would be.

NinoSkopac:
By the way, about half of the paragraphs are missing, not just one.

gijsk:
Going to close this per #408 . :-)

PR #408, opened 2017-11-22
Fix issue #407:  Telegraph articles failing

andreskrey:
Hi.

This change fixes issue #407.

In *The Telegraph* articles, some paragraphs are missing in the readability result. For example, [in this article](http://www.telegraph.co.uk/news/2017/11/16/zimbabwes-robert-mugabe-wife-grace-insisting-finishes-term-priest/), 3 paragraphs will be missing if you parse it with this library. An example of those missing paragraphs is this one:

```
<div class="component-content">
    <p>
        <span class="m_first-letter m_first-letter--flagged">Z</span>imbabwean intelligence reports seen by Reuters suggest that former security chief Emmerson Mnangagwa, who was ousted as vice-president this month, has been mapping out a post-Mugabe vision with the military and opposition for more than a year.
    </p>
</div>
```

Initially I thought that the `span` surrounding the Z was confusing the algorithm but the real problem was that the code was converting the `div` node to a `p` node and never rating it, which was later discarded when scanning the siblings of the top candidate.

This patch adds the node to the `elementsToScore` array after converting it to a `p` node. Makes sense since `p` elements are the default ones to score.

This does not break any test case. I also added a new one with this issue.

Feel free to contact me if you need more information.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -797,6 +797,7 @@ Readability.prototype = {
             var newNode = node.children[0];
             node.parentNode.replaceChild(newNode, node);
             node = newNode;
+            elementsToScore.push(node);
           } else if (!this._hasChildBlockElement(node)) {
             node = this._setNodeTag(node, "P");
             elementsToScore.push(node);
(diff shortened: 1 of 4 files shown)


gijsk:
This is awesome, thanks for the PR! :-)

Issue #400, opened 2017-09-26
Title is wrong


Praveena0989:
I am trying run readability on this [link](https://www.ndtv.com/india-news/surgical-strikes-a-message-to-pakistan-more-if-necessary-army-chief-general-bipin-rawat-1755007)  It works perfectly but the problem is with the title. I am running `readability` using `JSDOM` and `document.title` gives me title `Surgical Strikes A Message To Pakistan, More If Necessary: Army Chief General Bipin Rawat` but `Readability` giving me `Army Chief General Bipin Rawat` even in firefox browser reader view

gijsk:
This is now (via #409) fixed both in Readability the library and in Firefox Nightly (59).

PR #409, opened 2017-11-24
Fix issue #400: Fall back to the original title if there are too many words before the colon

andreskrey:
Hi,

This fixes issue #400.

In [the example provided](https://www.ndtv.com/india-news/surgical-strikes-a-message-to-pakistan-more-if-necessary-army-chief-general-bipin-rawat-1755007), the `<title>` tag says "Surgical Strikes A Message To Pakistan, More If Necessary: **Army Chief General Bipin Rawat**" and the `<h1>` says "Surgical Strikes A Message To Pakistan, More If Necessary: **Army Chief**". Since Readability considers the colon as a hierarchical separator (and no H node matches exactly the title), the resulting title is "Army Chief General Bipin Rawat"

This patch checks the length of the discarded text before the colon. If there are more than 5 words, it assumes something weird is happening with the title and falls back to the original title that comes from the `<title>`  tag.

I also added a test case for this.

Let me know if you have any questions.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -375,8 +375,13 @@ Readability.prototype = {
         curTitle = origTitle.substring(origTitle.lastIndexOf(':') + 1);
 
         // If the title is now too short, try the first colon instead:
-        if (wordCount(curTitle) < 3)
+        if (wordCount(curTitle) < 3) {
           curTitle = origTitle.substring(origTitle.indexOf(':') + 1);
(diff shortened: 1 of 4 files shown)


gijsk:
Apologies for the delay. This looks fine to me, though it also feels like maybe this section of code needs some refactoring TLC... Oh well, this fixes the problem at hand!

PR #411, opened 2017-11-29
Keep ids but not classes

gijsk:
@heycam can you sanity-check these changes? I'm basically part-reverting #406 so we keep the IDs, per the last comments in https://bugzilla.mozilla.org/show_bug.cgi?id=1417837 .


test/generate-testcase.js
```diff
@@ -110,7 +110,7 @@ function runReadability(source, destPath, metadataDestPath) {
   };
   var myReader, result, readerable;
   try {
-    myReader = new Readability(uri, doc);
+    myReader = new Readability(uri, doc, { classesToPreserve: ["caption"] });
```
heycam: Maybe add a comment here about why we're preserving this class?

heycam:
Looks good, thanks for fixing this up.

PR #413, opened 2017-12-02
Fix className usage so it deals correctly with SVG nodes (fixes #412).

### LINKED ISSUE #412: "node.className.split is not a function"
Some URLs (eg https://www.cnet.com/es/noticias/twitter-lite-america-latina/#ftag=ESPdf5230a) yield the following error:

```
TypeError: node.className.split is not a function
    at Object._cleanClasses (/user_code/node_modules/readability/Readability.js:246:8)
    at Object._cleanClasses (/user_code/node_modules/readability/Readability.js:259:12)
    at Object._cleanClasses (/user_code/node_modules/readability/Readability.js:259:12)
    at Object._cleanClasses (/user_code/node_modules/readability/Readability.js:259:12)
    at Object._postProcessContent (/user_code/node_modules/readability/Readability.js:141:10)
    at Object.parse (/user_code/node_modules/readability/Readability.js:1723:10)
    ...
```

I believe a potential fix would be to check if `node.className` is actually a string.

gijsk:
@bittlingmayer - could you confirm this fixes things for you? I've also created an automated testcase for this article.


gijsk:
@heycam would you mind reviewing this? See #412 for more context. :-)

gijsk:
Thanks!

bittlingmayer:
I'm assuming it does, will update when I get back to that code.

Issue #277, opened 2016-03-17
Nearly identical markup on two pages. One fails to render, wonder why?


un33k:
There are two pages that are rendered via the same template code.  The are nearly identically as far as the markup goes.

#1 - https://simplyfound.com/article/6d733be2aef4/apple-pay-slow-and-steady-wins-the-race

The above works very well when clicking on the READER icon.

#2 - https://simplyfound.com/article/eb9a5e137034/raspberry-pi-3-the-credit-card-sized-pc-that-cost-only-35-all-time-bestselling-computer-in-uk

The above fails with the "**Failed to load article from page**" message.

Same logic produces both pages, but one works and the other does not.

It would be cool to have a errorno of some sort displayed alongside the error message for debugging purposes.

gijsk:
Fixed by #278

PR #278, opened 2016-03-17
Address issue #277 by marking 'modal' unlikely+negative

gijsk:
@leibovic this seems pretty trivial, fairly safe (can't see why someone would use 'modal' for a story or main content) and doesn't require changing any of our existing tests, so should be OK, I think? I'll add debug output in a second for what happens when you try to reader-ize the story before this change, but basically it loses out the score contest with the hidden modal registration box on the page, which then doesn't yield reader output because there isn't enough content in that (modal) box.

PR #415, opened 2017-12-09
Remove 'modal' from the unlikelyCandidates and negative regexps

andreskrey:
Hi,

While parsing newspapers I noticed that the 'modal' string present in the [`unlikelyCandidates`](https://github.com/mozilla/readability/blob/master/Readability.js#L104) and [`negative`](https://github.com/mozilla/readability/blob/master/Readability.js#L107) regexps was removing the images on the lanacion.com.ar website. This website uses the `conModal` class for their images which matched this regexp and removed the whole node.

Example
(The alt-text for the image is there but the actual image is missing)

I removed the `modal` string from both regexps and all test cases pass, so I think is safe to merge this in. I also created a new test case for this change.

Let me know what you think


gijsk:
This is interesting, because I added this in #278 to fix #277 - see https://github.com/mozilla/readability/pull/278/files . A testcase was added as part of that PR. Do we know why we now no longer need to exclude 'modal' in order for that testcase to pass?

andreskrey:
As far as I can tell, the modal nodes get discarded during the "sibling appending" right after the creation of the articleContent var. Those divs belong to a bigger div that has 0 content score so it never reaches to the final HTML.

I rolled back to fcc645019f90800179bca0c12f40622e913bb696 (one commit before you merged those changes) and the algorithm was selecting a more broad div that contained these offending divs. My best guess is that we made the algorithm smarter to select an specific div instead of a more general one. I can investigate it further if you want to but I think we are safe to merge this in.

gijsk:
Cool, thanks for checking! I was traveling, so apologies for the delay here. Your explanation makes sense to me, let's do it.

Issue #414, opened 2017-12-04
Node is not defined


lazharichir:
Hello, just trying the package out but things aren't working as expected. I run it on `Node.js`.

	const jsdom = require("jsdom");
	const { JSDOM } = jsdom;
	const Readability = require('readability');

	process.on('unhandledRejection', r => console.log(r));

	//const url = 'https://engineering.linkedin.com/blog/2017/11/improving-resiliency-and-stability-of-a-large-scale-api';
	const url = "https://engineering.quora.com/Machine-Learning-Workshop-Combatting-Abuse-Online";

	JSDOM.fromURL(url).then(function( theDom ) {

		console.log(theDom.window.document.querySelector("p").textContent);
		console.log(theDom.window.document.querySelector("strong").textContent);

		var article = new Readability(url,theDom.window.document).parse();

		console.log(dom);
		console.log(article.title);
		
	});

This does give me a ReferenceError when executing it.

    My-MacBook-Pro:test-readability Lazhar$ node article.js --trace-warnings

    ReferenceError: Node is not defined
        at Readability._nextElement (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:430:30)
        at Readability.<anonymous> (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:455:27)
        at Proxy.forEach (<anonymous>)
        at Readability._forEachNode (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:192:29)
        at Readability._replaceBrs (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:445:10)
        at Readability._prepDocument (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:416:12)
        at Readability.parse (/Users/My/Tests/gcf/test-readability/node_modules/readability/Readability.js:1712:10)
        at /Users/My/Tests/gcf/test-readability/article.js:22:60
        at <anonymous>
        at process._tickCallback (internal/process/next_tick.js:188:7)

I am pretty much following the examples both for JSDOM and Mozilla Readability so I am somewhat confused. Help would be appreciated!

gijsk:
If using `jsdom` you need to add a `Node` global, like e.g. in this sample: https://gist.github.com/bittlingmayer/15dfeb44e8b5108d1a3f5107d4adc50d#file-test-js-L17


This is effectively a dupe of #346 .

gijsk:
[...]

PR #416, opened 2017-12-31
[Docs] Fixed JSDOM usage note

backflip:
As described in https://github.com/mozilla/readability/issues/414#issue-279172404, we need to pass JSDOM's `.window.document` property to Readability. Otherwise, it will fail with `TypeError: Cannot read property 'firstElementChild' of undefined`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Until [#346](https://github.com/mozilla/readability/issues/346) is fixed, if you
 ...
 var dom = new JSDOM(html, OPTIONS);
 Node = dom.window.Node;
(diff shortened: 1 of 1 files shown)


gijsk:
Huh, I wonder why the build is failing. Oh well. Thanks for fixing my oversight in the documentation! (And apologies for the slow response time, most of Mozilla has been on a break around Christmas and New Year's Eve. I'm normally slightly quicker to respond to PRs...)

backflip:
@gijsk, two days is blazingly fast! And thanks so much for this great library!

Issue #381, opened 2017-06-20
Fails to detect article content


aehlke:
This page contains an article title, image, and body which Readability fails to detect (`_grabArticle` returns `null`): http://hukumusume.com/douwa/pc/aesop/01/01.htm 

However, this similar (though longer) page is detected properly: http://hukumusume.com/douwa/pc/aesop/01/02.htm

andreskrey:
Resulting text is 406 words long, and by default Readability discards texts shorter than 500 words.
If you set the `wordThreshold` option to 300, the article gets parsed correctly. 

Try using something like this:

```
var uri = {
    spec: loc.href,
    host: loc.host,
    prePath: loc.protocol + "//" + loc.host,
    scheme: loc.protocol.substr(0, loc.protocol.indexOf(":")),
    pathBase: loc.protocol + "//" + loc.host + loc.pathname.substr(0, loc.pathname.lastIndexOf("/") + 1)
};
var article = new Readability(uri, document, {'wordThreshold': 303}).parse();
```

aehlke:
Thanks! I've forked readability for now to iteratively reduce the threshold if no body text is found, rather than bail completely.

gijsk:
Is the word count accurate here? I'm afraid I don't speak Japanese, but judging from the translated result, I expect our naive regex-based word parsing is also massively undercounting the number of words. From a quick look at Google, without a fully-fledged natural language parser, it's not all that feasible to "count words" in Hiragana, Katakana or Kanji (Chinese or Japanese).

Perhaps we should ignore the word count if we encounter a significant proportion of non-Latin text? That would be a bit error-prone, but it feels somewhat better than just giving up for the wrong reasons...

@aehlke does that sound sane, and/or is this something you would be interested in fixing? :-)

andreskrey:
Maybe we can count characters, if that is possible in other languages? Something like "if the majority of the characters are non-western, count X amount of characters to decide if the text was parsed correctly"

gijsk:
Yes, I suppose we could fall back to a character count. That's probably better than ignoring the word count completely - good idea!

JoanEspasa:
Would a customized "length" function for strings and a regexp that detects when a piece of text belongs to the CKJ codepoints be a good start?

andreskrey:
Performance might be an issue. I think counting characters for non-western languages is better.

aehlke:
[...]

PR #421, opened 2018-02-17
Return longest text after failing to detect text longer than the configured value

andreskrey:
Hi,

This solves issue #381.

I altered the loop that disables flags and now the algorithm will always return some text, which will be the longest it can find across the different attempts.

There's only one concern I have with this PR and it's [on line 1108](https://github.com/mozilla/readability/blob/0cdd9c61c4abba907f488da2768c4926d31e0939/Readability.js#L1108). I need to access the parent `this` scope to use the `_getInnerText` function and the only way I found to do so was to clone the `this` variable in the `that` variable. I'm sure there's something wrong because my IDE complains saying that the "Mutable variable is accessible from closure".

@gijsk can you apply some of your JS-fu and maybe suggest a better way to do this?

Thanks

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1083,12 +1083,15 @@ Readability.prototype = {
       if (this._debug)
         this.log("Article content after paging: " + articleContent.innerHTML);
 
+      var parseSuccessful = true;
+
       // Now that we've gone through the full algorithm, check to see if
       // we got any meaningful content. If we didn't, we may need to re-run
       // grabArticle with different flags set. This gives us a higher likelihood of
       // finding the content, and the sieve approach gives us a higher likelihood of
       // finding the -right- content.
       if (this._getInnerText(articleContent, true).length < this._wordThreshold) {
+        parseSuccessful = false;
         page.innerHTML = pageCacheHtml;
 
         if (this._flagIsActive(this.FLAG_STRIP_UNLIKELYS)) {
@@ -1102,19 +1105,25 @@ Readability.prototype = {
           this._attempts.push(articleContent);
         } else {
           // No luck after removing flags, just return the longest text we found during the different loops
-          that = this;
+          var that = this;
           var results = this._attempts.sort(function (a, b) {
             return that._getInnerText(a, true).length < that._getInnerText(b, true).length;
           });
+
+          // But first check if we actually have something
           if (this._getInnerText(results[0], true).length === 0) {
             return null;
           }
+
           articleContent = results[0];
+          parseSuccessful = true;
         }
-      } else {
+      }
+
+      if (parseSuccessful) {
         // Find out text direction from ancestors of final top candidate.
         var ancestors = [parentOfTopCandidate, topCandidate].concat(this._getNodeAncestors(parentOfTopCandidate));
-        this._someNode(ancestors, function(ancestor) {
+        this._someNode(ancestors, function (ancestor) {
           if (!ancestor.tagName)
             return false;
           var articleDir = ancestor.getAttribute("dir");
--- a/Readability.js
+++ b/Readability.js
@@ -101,10 +101,10 @@ Readability.prototype = {
   // All of the regular expressions in use within readability.
   // Defined up here so we don't instantiate them repeatedly in loops.
   REGEXPS: {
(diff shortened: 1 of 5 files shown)


Readability.js
```diff
@@ -1082,27 +1083,47 @@ Readability.prototype = {
...
 
+      var parseSuccessful = true;
+
       // Now that we've gone through the full algorithm, check to see if
       // we got any meaningful content. If we didn't, we may need to re-run
       // grabArticle with different flags set. This gives us a higher likelihood of
       // finding the content, and the sieve approach gives us a higher likelihood of
       // finding the -right- content.
       if (this._getInnerText(articleContent, true).length < this._wordThreshold) {
+        parseSuccessful = false;
```
gijsk: So what I would do is move the result of the `_getInnerText().length` text call to a variable before the `if` statement:

```js
var textLength = this._getInnerText(articleContent, true).length;
if (textLength < this._wordThreshold) {
```

Readability.js
```diff
@@ -1082,27 +1083,47 @@ Readability.prototype = {
...
         page.innerHTML = pageCacheHtml;
 
         if (this._flagIsActive(this.FLAG_STRIP_UNLIKELYS)) {
           this._removeFlag(this.FLAG_STRIP_UNLIKELYS);
+          this._attempts.push(articleContent);
```
gijsk: Then in all of these, I'd push an object:

```js
this._attempts.push({articleContent: articleContent, textLength: textLength});
```

Readability.js
```diff
@@ -1082,27 +1083,47 @@ Readability.prototype = {
...
         } else if (this._flagIsActive(this.FLAG_WEIGHT_CLASSES)) {
           this._removeFlag(this.FLAG_WEIGHT_CLASSES);
+          this._attempts.push(articleContent);
         } else if (this._flagIsActive(this.FLAG_CLEAN_CONDITIONALLY)) {
           this._removeFlag(this.FLAG_CLEAN_CONDITIONALLY);
+          this._attempts.push(articleContent);
         } else {
-          return null;
+          // No luck after removing flags, just return the longest text we found during the different loops
```
gijsk: Note that you're missing the last result here - we need to push `articleContent` in the array here, too!

Readability.js
```diff
@@ -1082,27 +1083,47 @@ Readability.prototype = {
...
+          var that = this;
+          var results = this._attempts.sort(function (a, b) {
```
gijsk: Then you don't need to rebind `this` here and can just sort like so:

```js
this._attempts.sort(function(a, b) {
  return a.textLength < b.textLength;
});
```

This has the advantage that we don't recompute the text content a lot of times, and we also avoid the `this` reference.

Note that in JavaScript, `Array.sort` modifies the array on which it's called, so you don't need to assign to `var results` here.

Readability.js
```diff
@@ -1082,27 +1083,47 @@ Readability.prototype = {
...
+            return that._getInnerText(a, true).length < that._getInnerText(b, true).length;
+          });
+
+          // But first check if we actually have something
+          if (this._getInnerText(results[0], true).length === 0) {
```
gijsk: Here you could then just check `if (!this._attempts[0].textLength)`

gijsk:
Should we add a test for this case, by the way? :-)

andreskrey:
Yeah, of course, I'll add the html of the referenced issue. I'll try to solve your reviews this week. Thanks!

gijsk:
Hey @andreskrey , thanks for this. There seems to be some issue with this sequence of changes because github is telling me "This branch cannot be rebased due to conflicts". Looking at the diff view, it looks OK to me. Do you know what's up, or shall I just try to do a separate merge commit myself to land this?

andreskrey:
@gijsk Yep yep yep, I'm aware of that, I'm not done yet tough! :)

I'm working on the test case and then I'll rewrite my commits, somewhere down the line I screwed the git log.

I'll let you know once it's done

andreskrey:
Also, thank you a lot for the review. Helped me a lot to understand the issues I had with my code :)

andreskrey:
Somehow I fucked the git log beyond repair. I'm going to create a new PR and close this one.

XKCD

PR #425, opened 2018-02-27
Improve URL handling in JSDOMParser and Readability.js

gijsk:
This change ups the required node version to 7.0 because it relies on the builtin url module.

We now pass a url when constructing a jsdom document or JSDOMParser document.
Because this is an API change, I'm increasing the package version.

Ultimately, I would like to remove the `url` argument from the readability constructor. It should
use the `documentURI` from the document it is passed.

@msujaws and @bradphilips, could you take a look at this?

### DIFF
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -229,6 +229,7 @@ describe("Test pages", function() {
 
       runTestsWithItems("jsdom", function(source) {
         var doc = jsdom(source, {
+          url: uri.spec,
           features: {
             FetchExternalResources: false,
             ProcessExternalResources: false
@@ -240,7 +241,7 @@ describe("Test pages", function() {
 
       runTestsWithItems("JSDOMParser", function(source) {
         var parser = new JSDOMParser();
-        var doc = parser.parse(source);
+        var doc = parser.parse(source, uri.spec);
         if (parser.errorState) {
           console.error("Parsing this DOM caused errors:", parser.errorState);
           return null;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - '6.5'
+  - '7.0'
 script:
   - npm run lint
   - npm run test
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -553,7 +553,8 @@
     },
   };
 
-  var Document = function () {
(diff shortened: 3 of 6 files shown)


gijsk:
@bradphilips I think this should mean that in `_fixRelativeURIs` in #422 you can just read `this._doc.baseURI` to get the base URI spec.

I suspect the easiest way to resolve the relative URL there is to just run `new URL(relativehref, baseURI)`. This should work as-is with the changes in this PR applied.

bradphilips:
@gijsk this is awesome. Thank you so much, this will definitely make resolving the relative path a lot easier. I assume this is reliant on using mozilla's version of `JSDOMParser`? At the moment I'm using the packaged version ([JSDOM](https://github.com/jsdom/jsdom)) although I have no qualms using Readability's version.

How do you want to handle the merge from #422? I could fork yours and implement the change for the `this._doc.baseURI` or you could just take my changes into your branch and push the upstream from there. I don't think there's much to do to make this work. This simplifies it greatly. Let me know. This is great. 👍

test/test-readability.js
```diff
@@ -239,7 +240,7 @@ describe("Test pages", function() {
       }, uri, testPage.source, testPage.expectedContent, testPage.expectedMetadata);
 
       runTestsWithItems("JSDOMParser", function(source) {
-        var parser = new JSDOMParser();
+        var parser = new JSDOMParser(uri.spec);
```
bradphilips: This I think should be:

```javascript
var parser = new JSDOMParser();
var doc = parser.parse(source, uri.spec);
```

gijsk: D'oh. Thanks for catching this...

gijsk:
> I assume this is reliant on using mozilla's version of JSDOMParser? At the moment I'm using the packaged version (JSDOM) although I have no qualms using Readability's version.

Nope, `jsdom` also supports `document.baseURI` so it should work there, too.

gijsk:
> How do you want to handle the merge from #422? I could fork yours and implement the change for the this._doc.baseURI or you could just take my changes into your branch and push the upstream from there. I don't think there's much to do to make this work. This simplifies it greatly. Let me know. This is great. 👍

I can look at pushing it upstream once this lands, sure. I just need review from someone. I think Jared's a bit busy at the moment, I'll see if I can find someone else...

PR #430, opened 2018-03-16
Fix _replaceBrs

davidar:
Previously, `nextElem` was not actually proceeding to the next element, and therefore aborting the paragraph at the first `<br>` (rather than the first `<br><br>` as the comment indicates).

This is currently causing some of the other tests to fail, but it does fix the behaviour for eg. `a<br><br>b<br>c<br><br>d<br>e`, which should become `a<p>b<br>c</p><p>d<br>e</p>` rather than `a<p>b</p>c<p>d</p>e`

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -457,7 +457,7 @@ Readability.prototype = {
         while (next) {
           // If we've hit another <br><br>, we're done adding children to this <p>.
           if (next.tagName == "BR") {
-            var nextElem = this._nextElement(next);
+            var nextElem = this._nextElement(next.nextSibling);
             if (nextElem && nextElem.tagName == "BR")
               break;
           }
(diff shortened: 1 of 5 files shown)


gijsk:
Hi, thank you for the PR! Apologies for the delay, my weekend availability is a bit intermittent. I intend to look at this more fully hopefully tomorrow (Monday) or worst-case Tuesday. At a glance, the change looks sensible - can you update the 2 tests that are failing as well (ie change the `expected.html` file for them to match what they output with your modification to `Readability.js`) and add a commit to this PR for that? Thank you!

davidar:
Done. I also modified `replace-brs` to test the change in behaviour.

gijsk:
Looks great, thank you!

PR #431, opened 2018-03-21
Clean <aside> tags in _prepArticle

andreskrey:
As mentioned in https://github.com/andreskrey/readability.php/issues/53, `<aside>` tags usually do not contribute towards the actual content of the article.

Adding `<aside>` to the list of tags to clean during `_prepArticle()` not only allows to parse correctly the new "Engadget" test case but also removes some empty tags we had in other test cases.

Let me know if you have any questions.


gijsk:
Makes sense to me, thanks for the PR! It's possible we'll want to tweak this in the future (ie perhaps there are sites that use `<aside>` for reasonable content), but I think certainly to start with this should be fine.

PR #432, opened 2018-03-31
Handle itemprop articles more fluidly

bradphilips:
Hey guys,

This started out as an endeavor to handle the `itemprop` specification for `articleBody` and `articlesBody` more effectively so that readability could properly parse these without having to do much work to find the content and handle subscription newsletters more effectively. 

It turned into a couple more enhancements to JSDOM that I found along the way, including the ability to search by attributes (needed for finding `articleBody` itemprop's) along with handling script tags that contained > or < symbols in the content. Also implemented a more effective (I think) handling of closing tags and fixed some regressions on weighting and allowed the ability to run a list of tests alone via an environment variable(ex: `READABILITY_TESTS=la-nacion,android-weekly npm test`).

I'm still working out some kinks with some of the subscription newsfeed articles but I think I might put that in a new pull request, figured I should get eyes on this one first. Let me know if you have questions and apologies for not splitting these up more.. Love this library. Thanks!


gijsk:
Hey, thanks for the patch. Unfortunately, the script tag stuff makes me suspect you're passing non-valid-XHTML content into JSDOMParser. We had some logic in the script parsing to deal with this in the past, and then removed it because it caused breakage when passed actual valid XHTML content. Ditto for self-closing tags. Is it possible that we should just be fixing the testcase content instead, similar to what would happen in Firefox if you used reader mode there?

If using this as a library and DOM parsing somewhere that doesn't have a DOM parser, I would strongly recommend using `jsdom` instead.

Let me know if I've misunderstood parts of this PR. Note that if some of these fixes are really required, they might be better off in dedicated PRs (because smaller PRs are easier to review and merge).

bradphilips:
@gijsk, looks like you're right and I thought I'd tested this in Firefox and got the same result. The site under question was the [Android Developer Blog](https://android-developers.googleblog.com/2018/03/android-things-developer-preview-7.html) and was due a `>` being in the script tag which was blowing up when attempting to generate the tests. Even if I pull the latest and attempt to generate the test against the rendered DOM it blows up -- but it does look like its working in Firefox and has always worked with JSDOM. I'll pull out these changes and submit another PR with the changes for the itemprop. I still think these would be beneficial. Thanks for your help!

PR #434, opened 2018-04-05
Fix absoule and data src in 'toAbsoluteURI', check for 'data-lazy-src…

brainm:
…' attribute


gijsk:
Thanks for the pull request. Can you elaborate on why the explicit checks for `data:...` and `http(s)` are necessary?

Note that unfortunately there are both test and eslint failures in the current version of the PR as built by travis ( https://travis-ci.org/mozilla/readability/builds/362581605 ).

PR #439, opened 2018-04-24
Make unused URI object optional

davidar:
The URI object isn't actually used anywhere in the code, other than to be [passed back unmodified to the caller](https://github.com/mozilla/readability/blob/master/Readability.js#L1751), so it seems to make sense to be able to call the constructor without it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,23 +16,14 @@ Please make sure to run [eslint](http://eslint.org/) against any proposed change
 
 ## Usage
 
-To parse a document, you must create a new `Readability` object from a URI object and a document object, and then call `parse()`. Here's an example:
+To parse a document, you must create a new `Readability` object from a document object, and then call `parse()`. Here's an example:
 
 ```javascript
-var loc = document.location;
-var uri = {
-  spec: loc.href,
-  host: loc.host,
-  prePath: loc.protocol + "//" + loc.host,
-  scheme: loc.protocol.substr(0, loc.protocol.indexOf(":")),
-  pathBase: loc.protocol + "//" + loc.host + loc.pathname.substr(0, loc.pathname.lastIndexOf("/") + 1)
-};
-var article = new Readability(uri, document).parse();
+var article = new Readability(document).parse();
 ```
 
 This `article` object will contain the following properties:
 
-* `uri`: original `uri` object that was passed to constructor
 * `title`: article title
 * `content`: HTML string of processed article content
 * `length`: length of article, in characters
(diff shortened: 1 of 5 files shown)


gijsk:
Yeah, this is straightforward after #425 . Really, perhaps we should just remove support for the old form... but I guess this is safest for now.

This looks good as-is, but would you mind also updating the code in the `test/` directory? Thanks!

davidar:
> Really, perhaps we should just remove support for the old form

That would certainly be the cleaner option, I just wasn't sure what the policy was on making breaking changes to the public API

> This looks good as-is, but would you mind also updating the code in the test/ directory? Thanks!

Done

gijsk:
> That would certainly be the cleaner option, I just wasn't sure what the policy was on making breaking changes to the public API

I think there isn't one (a policy, that is) or at least if there is I'm not aware of it. Whenever I've tried to get #43 resolved I've hit some walls, and so there's no proper semver stuff in place (and if there was, no way for people to use that rather than raw github pointers) and so it's not clear when we can/can't change things. I'll try to remember to chase it again this week...

Readability.js
```diff
@@ -27,6 +27,10 @@
  * @param {Object}       options The options object.
  */
 function Readability(uri, doc, options) {
+  if (arguments.length === 1) {
```
gijsk: Ah, hm. The test changes made me realize this way of checking arguments means we can't omit the URL when passing options... That seems unfortunate.

Let's be a little bit more forward and change the function signature, and then use duck typing to figure out what people are passing, so we get:

```js
function Readability(doc, options) {
  // In some older versions, people passed a URI object as the first argument. Cope:
  if (doc && !doc.documentElement && doc.spec) {
    doc = options;
    options = arguments[2];
  }
```

Please also update the doc comment above the constructor definition to omit the `uri` bit. :-)

Then I think we can remove the explicit `null` being passed from the callsites in `test/`.

README.md
```diff
@@ -16,23 +16,14 @@ Please make sure to run [eslint](http://eslint.org/) against any proposed change
...
-  scheme: loc.protocol.substr(0, loc.protocol.indexOf(":")),
-  pathBase: loc.protocol + "//" + loc.host + loc.pathname.substr(0, loc.pathname.lastIndexOf("/") + 1)
-};
-var article = new Readability(uri, document).parse();
+var article = new Readability(document).parse();
 ```
 
 This `article` object will contain the following properties:
 
-* `uri`: original `uri` object that was passed to constructor
```
gijsk: Can you then also remove the `uri` property from the returned result, please? :-)

davidar: Done

gijsk:
Awesome, thanks very much for cleaning this up, much nicer now!

Issue #441, opened 2018-04-27
Should test the incoming arguments and check that the document passed looks vaguely like a document


gijsk:
We could check that it's non-null, not a string, and has a `documentElement` property, for a start.

gijsk:
Fixed in #445.

PR #445, opened 2018-04-30
Check for a document in either of the first two arguments (fixes #444 and #441)

### LINKED ISSUE #444: jsdom version issues
As discussed in https://github.com/mozilla/readability/issues/376#issuecomment-385206370, the following code fails:

https://gist.github.com/bittlingmayer/7a14f3a9c17b42163d12ba7da9bd1866

```
	Content.parse( https://www.nzz.ch/international/nahost-und-afrika/komitee-zur-befreiung-der-levante-extremistengruppen-in-syrien-schliessen-sich-zusammen-ld.142403 )
Fetching  https://www.nzz.ch/international/nahost-und-afrika/komitee-zur-befreiung-der-levante-extremistengruppen-in-syrien-schliessen-sich-zusammen-ld.142403
Fetched  https://www.nzz.ch/international/nahost-und-afrika/komitee-zur-befreiung-der-levante-extremistengruppen-in-syrien-schliessen-sich-zusammen-ld.142403
Content._parse(...)
(node:5014) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'firstElementChild' of undefined
(node:5014) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

gijsk:
This should fix #444, I believe. @davidar and/or @bittlingmayer, can you have a look if this makes sense to you? Thanks!

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -26,10 +26,12 @@
  * @param {Object}       options The options object.
  */
 function Readability(doc, options) {
-  // In some older versions, people passed a URI object as the first argument. Cope:
-  if (doc && !doc.documentElement && doc.spec) {
+  // In some older versions, people passed a URI as the first argument. Cope:
+  if (options && options.documentElement) {
     doc = options;
     options = arguments[2];
+  } else if (!doc || !doc.documentElement) {
+    throw new Error("First argument to Readability constructor should be a document object.");
   }
   options = options || {};
 
--- a/test/test-readability.js
+++ b/test/test-readability.js
@@ -190,19 +190,20 @@ function removeCommentNodesRecursively(node) {
 
 describe("Readability API", function() {
   describe("#constructor", function() {
+    var doc = new JSDOMParser().parse("<html><div>yo</div></html>");
     it("should accept a debug option", function() {
-      expect(new Readability({})._debug).eql(false);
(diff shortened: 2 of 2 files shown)


Readability.js
```diff
@@ -26,8 +26,8 @@
  * @param {Object}       options The options object.
  */
 function Readability(doc, options) {
-  // In some older versions, people passed a URI object as the first argument. Cope:
-  if (doc && !doc.documentElement && doc.spec) {
+  // In some older versions, people passed a URI object or string as the first argument. Cope:
+  if (doc && ((!doc.documentElement && doc.spec) || typeof doc == "string")) {
```
davidar: Perhaps it would be worth fixing #441 at the same time and doing something like the following?

```js
if (doc && doc.documentElement) {
  // ok
} else if (options && options.documentElement) {
  doc = options;
  options = arguments[2];
} else {
  throw new Error("invalid arguments");
}
```

I think some of the tests pass in an empty object for the document though, so those would also need to be changed.

gijsk: Good point. Also thanks for the heads-up about the test!

I'll hold off printing a warning for now. Maybe when we publish to npm...

PR #449, opened 2018-05-16
Don't put non-phrasing content into paragraphs

davidar:
This fixes an issue where `_replaceBrs` was putting block-level elements like headings inside paragraphs.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -497,11 +497,15 @@ Readability.prototype = {
               break;
           }
 
+          if (!this._isPhrasingContent(next)) break;
+
           // Otherwise, make this node a child of the new <p>.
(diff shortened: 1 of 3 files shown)


Readability.js
```diff
@@ -497,11 +497,15 @@ Readability.prototype = {
               break;
           }
 
+          if (!this._isPhrasingContent(next)) break;
```
gijsk: Nit: please put the `break;` on the next line (indented, of course).

gijsk: I should probably tighten up the eslint rules to make this less frictionny, actually, so let's leave it for now and I'll try and sort it out at a later point.

Issue #451, opened 2018-05-17
Invalid HTML in test cases


davidar:
I noticed readability is producing invalid markup for some of the test cases:

- content wrapped in a `<td>` without a corresponding `<table>` in `lwn-1`

- paragraphs nested inside a paragraph in `yahoo-4`

- headings nested inside paragraphs in `engadget`:

```html
<p>
  <h2> But only hardcore gamers will appreciate it. </h2>
</p>
```

- and `tmz-1`:

```html
<p>
  <h2>Lupita Nyong'o</h2>
  <h4>$150K Pearl Oscar Dress ... STOLEN!!!!</h4>
</p>
```

davidar:
I think the second one is because `_replaceBrs` doesn't check if the `br`s are already inside a paragraph, but I haven't looked into the others yet.

Issue #343, opened 2017-01-21
Reduce down singly nested containers when returning final content from the readability algorithm


gijsk:
In #337 we started replacing candidate container elements with their parents if they had no siblings. This helped in cases where major blocks of text were deeply nested and we were therefore unable to detect their siblings.

However, there are 2 downsides:
1) we end up adding needless `<div>` and similar nesting to elements that, given the lack of styling for them, adds no value to the readability result.
2) we no longer correctly detect `dir` attributes on the container element, because the container element has changed.

We should fix this by having an algorithm (after all the cleanup etc., once we're sure we're returning a particular candidate element) that does 3 things:
1) while the candidate has exactly 1 element child node and no non-whitespace text children, make its child the top candidate
2) in that case, if the parent has a `dir` attribute and the child does not, set the `dir` on the child. (If the child has a `dir` attribute, that direction should "win", ie persist.)
3) if the final candidate we land on this way is not `<p>` or `<div>` or `<article>` or `<section>`, create a `<div>` and put all the child nodes (not just elements!) in that `<div>` and return that `<div>`. This will fix cases where we'd otherwise return `<table>` or `<td>` elements.

@evanxd, can you take this on?

evanxd:
Sure, let me investigate this issue...

PR #455, opened 2018-05-26
Don't nest paragraphs

davidar:
This fixes the second item of #451, along with some other issues I ran into in the process.


gijsk:
(Sorry for the delay, I've been on holiday. I will hopefully get to this tomorrow.)

Readability.js
```diff
@@ -506,6 +506,8 @@ Readability.prototype = {
         }
 
         while (p.lastChild && this._isWhitespace(p.lastChild)) p.removeChild(p.lastChild);
+
+        if (p.parentNode.tagName === "P") this._setNodeTag(p.parentNode, "DIV");
```
gijsk: So my understanding is that this code is meant to deal with `<br>` separated paragraphs by turning them into actual paragraphs. If we end up doing this *inside* a `<p>`, wouldn't the most semantically correct thing be to make the new `<p>` a sibling of the existing `<p>` ?

In other words, why did you end up going for making these `<div>`s ? Maybe there's a reason not to do what I'm suggesting that I'm missing (maybe just complexity) ?

test/generate-testcase.js
```diff
@@ -75,17 +75,21 @@ function fetchSource(url, callbackFn) {
...
   });
 }
 
+function sanitizeSource(html, callbackFn) {
+  htmltidy(serializeDocument(jsdom(html)), {
+    "indent": true,
+    "indent-spaces": 4,
+    "numeric-entities": true,
+    "output-xhtml": true,
+    "wrap": 0
```
gijsk: Nitpick: can you add a trailing comma after the last property as well? This makes it easier to add/remove entries in objects/arrays.

test/test-pages/yahoo-4/expected.html
```diff
@@ -1,18 +1,15 @@
...
-         　トレンドマイクロがスマートフォン保持者でフリーWi-Fiの利用経験がある人に実施した調査では、回答者の約85％が安全なフリーWi-Fiと危険なフリーWi-Fiは「見分けられない」と回答。さらに、約65％がフリーWi-Fiの利用に不安を感じていると回答している。
-        </p><p>
-         　こうした環境の変化やユーザの状況を鑑み、フリーWi-Fiプロテクションの提供を開始する。同アプリをインストールすることで利用者は、万が一安全性の低いフリーWi-Fiのアクセスポイントに接続してしまった場合でも、その通信を暗号化でき、通信の盗み見やそれによる情報漏えいのリスクを低減できるようになる。
-        </p><p>
-         　具体的には、フリーWi-Fi利用時に、スマートフォンがフリーWi-Fiプロテクションインフラに接続することにより、フリーWi-Fiのアクセスポイントを介した通信がVPN（Virtual Private Network）で暗号化される。これにより利用者は、第三者から通信を傍受されることやデータの情報漏えいを防ぐことが可能。さらに、かんたん自動接続の機能により、通信を暗号化していない安全性が低いフリーWi-Fi接続時や利用者が指定したWi-Fiへ接続する際に、自動的に通信を暗号化し、利用者の通信を保護する。</p><p>
-         　また、フリーWi-Fiプロテクションインフラと、莫大なセキュリティ情報のビッグデータを保有するクラウド型セキュリティ技術基盤「Trend Micro Smart Protection Network」（SPN）が連携することで、フリーWi-Fiプロテクションインフラを経由してインターネットを利用する際に、利用者がフィッシング詐欺サイトや偽サイトなどへの不正サイトへアクセスすることをブロックできるという。</p>
-    </p>
-</div>
+    <div>
+        <div>
```
gijsk: Hm, why don't we unify the 2 divs (with no other kids), when we clearly previously unified the `<div><p></p></div>` nesting to remove the `<div>` ?

gijsk:
Thank you for splitting these changes into logical commits. I'm wishing that I could just merge the first 3 commits, but I assume that would cause test failures because of the missing commit 5, which then also changes things for commit 4 which I'm less sure about. :-(

Readability.js
davidar: I did it this way mostly because it was the simplest - the end result is the same, as a new paragraph gets created for the content before the `br` chain by

https://github.com/mozilla/readability/blob/d4b842c82a86c09d1f5c9f8ea8781692ba3cd147/Readability.js#L794-L799

the only difference being that all those new paragraphs get grouped together in a `div` (`<p>foo<br><br>bar</p>` -> `<div><p>foo</p><p>bar</p></div>`)

test/test-pages/yahoo-4/expected.html
davidar: I'm not sure, is it because of #343?

gijsk: Ah, yes. Yes, it is.

Readability.js
gijsk: OK, fair enough. Let's do this then. :-)

gijsk:
(I don't care enough about the nitpick not to merge, so I just merged as-is. :-) )

PR #456, opened 2018-06-08
Remove single-cell tables

davidar:
These seem to just clutter the output for no benefit.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -615,6 +615,19 @@ Readability.prototype = {
       if (next && next.tagName == "P")
         br.parentNode.removeChild(br);
     });
+
+    // Remove single-cell tables
+    this._forEachNode(this._getAllNodesWithTag(articleContent, ["table"]), function(table) {
+      var tbody = this._hasSingleTagInsideElement(table, "TBODY") ? table.firstElementChild : table;
+      if (this._hasSingleTagInsideElement(tbody, "TR")) {
+        var row = tbody.firstElementChild;
+        if (this._hasSingleTagInsideElement(row, "TD")) {
+          var cell = row.firstElementChild;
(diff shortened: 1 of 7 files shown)


Readability.js
```diff
@@ -615,6 +615,19 @@ Readability.prototype = {
       if (next && next.tagName == "P")
         br.parentNode.removeChild(br);
     });
+
+    // Remove single-cell tables
+    this._forEachNode(this._getAllNodesWithTag(articleContent, ["table"]), function(table) {
+      var tbody = (table.children.length === 1 && table.children[0].tagName === "TBODY") ? table.children[0] : table;
```
gijsk: Nit: `==` rather than `===` please.

Also, I wonder if for 'real' DOM implementations it'd be cheaper to compare `firstElementChild == lastElementChild`. I guess that's longer to write though...

test/test-pages/table-style-attributes/expected.html
```diff
@@ -3,35 +3,29 @@
     <span size="4"><br/>...or, why do I bother.</span></b>
         </span><br/> © 2002, 2003 <a href="mailto:/dev/null@jwz.org?subject=Please%20delete%20this%20message%20without%20reading%20it.">Jamie Zawinski</a> </p>
     <div>
-        <table>
```
gijsk: This test is supposed to check that we remove attributes like `bgcolor` and `border`. Now the test no longer checks this. Instead, can you modify both `source.html` and `expected.html` to include more than 1 cell so the table stays? (maybe just split some of the paragraphs off into their own cell or something)

Readability.js
davidar: Now defers to a generalised version of `_hasSinglePInsideElement` to keep things consistent

PR #458, opened 2018-06-08
Strip XML namespaces from tag names to deal with broken serializations

gijsk:
Per discussion with @bzbarsky on IRC re https://bugzilla.mozilla.org/show_bug.cgi?id=1394941 , this seems like the simplest solution to the issue we're facing there. The `XMLSerializer` does the right thing per the spec, even if it's... confusing. And JSDOMParser can cope relatively easily, so let's just do that.

@msujaws , can you doublecheck this looks OK to you? I added some minor unit tests. :-)

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -611,6 +611,13 @@
   };
 
   var Element = function (tag) {
+    // We use this to find the closing tag.
+    this._matchingTag = tag;
+    // We're explicitly a non-namespace aware parser, we just pretend it's all HTML.
+    var lastColonIndex = tag.lastIndexOf(":");
+    if (lastColonIndex != -1) {
+      tag = tag.substring(lastColonIndex + 1);
+    }
     this.attributes = [];
     this.childNodes = [];
     this.children = [];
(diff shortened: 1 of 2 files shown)


gijsk:
Wow, `let` has become such an automatic thing now...

JSDOMParser.js
```diff
@@ -1118,7 +1125,7 @@
       // If this isn't a void Element, read its child nodes
       if (!closed) {
         this.readChildren(node);
-        var closingTag = "</" + localName + ">";
+        var closingTag = "</" + node._matchingTag + ">";
```
msujaws: I don't think this will match closing tags such as `</p   >` which is valid according to https://www.w3.org/TR/html5/syntax.html#end-tags

bzbarsky: XMLSerializer never produces those, though.

msujaws: Okay thanks for the clarification.

gijsk: This was a pre-existing issue, too, so I'll just merge this as-is, but @msujaws let me know if I should file an issue to change this to be defensive in case XMLSerializer ever changes.

msujaws: Meh, I'm fine with not worrying about this change. I was more concerned if this was unsanitized markup.

PR #461, opened 2018-06-12
Add some title separators "«<"

naiba:
Add title separators: "«<"


gijsk:
Thanks for the PR! Could you add or modify a testcase that uses these separators?

naiba:
Weigang Wan is a very interesting scholar.

gijsk:
Thanks for the quick update! It looks like the test you added is failing. Is the source testcase valid XHTML? It looks like it might not be.

> Weigang Wan is a very interesting scholar. @gijsk

I'm not sure what to do with this statement...

PR #478, opened 2018-08-07
Improve metadata extraction

da2x:
* Recognize meta[property] as a space-separated list.
* Recognize Dublin Core (dc|dcterm): metadata.
* Prefer Dublin Core, Open Graph, HTML5, and Twitter in that order.
* _getArticleTitle() is now only used as fallback if document doesn't provide good metadata. This function has many localization issues and is “magical” in nature.
* Check for matches in both name and property attributes (both OGP and Twitter have mixed these up in recommendations over the years).

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1202,7 +1202,7 @@ Readability.prototype = {
     var propertyPattern = /\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*/gi;
 
     // name is a single value
-    var namePattern = /^\s*(?:(dc|dcterm|og|twitter)\s*[\.:]\s*)?(author|creator|description|title)\s*$/gi;
+    var namePattern = /^\s*(?:(dc|dcterm|og|twitter)\s*[\.:]\s*)?(author|creator|description|title)\s*$/i;
 
     // Find description tags.
     this._forEachNode(metaElements, function(element) {
@@ -1224,7 +1224,7 @@ Readability.prototype = {
           }
         }
       }
-      if (!matches && namePattern.test(elementName)) {
+      if (!matches && elementName && namePattern.test(elementName)) {
         name = elementName;
         if (content) {
           // Convert to lowercase, remove any whitespace, and convert dots
--- a/Readability.js
+++ b/Readability.js
@@ -1198,61 +1198,66 @@ Readability.prototype = {
     var values = {};
     var metaElements = this._doc.getElementsByTagName("meta");
 
-    // Match "description", or Twitter's "twitter:description" (Cards)
-    // in name attribute.
-    var namePattern = /^\s*((twitter)\s*:\s*)?(description|title)\s*$/gi;
+		// property is a space-separated list of values
+    var propertyPattern = /\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*/gi;
 
-    // Match Facebook's Open Graph title & description properties.
-    var propertyPattern = /^\s*og\s*:\s*(description|title)\s*$/gi;
+    // name is a single value
+    var namePattern = /^\s*(?:(dc|dcterm|og|twitter)\s*[\.:]\s*)?(author|creator|description|title)\s*$/gi;
 
     // Find description tags.
     this._forEachNode(metaElements, function(element) {
       var elementName = element.getAttribute("name");
       var elementProperty = element.getAttribute("property");
-
-      if ([elementName, elementProperty].indexOf("author") !== -1) {
-        metadata.byline = element.getAttribute("content");
-        return;
-      }
-
+      var content = element.getAttribute("content");
+      var matches = null;
       var name = null;
-      if (namePattern.test(elementName)) {
-        name = elementName;
-      } else if (propertyPattern.test(elementProperty)) {
-        name = elementProperty;
-      }
 
-      if (name) {
+      if (elementProperty) {
+        matches = elementProperty.match(propertyPattern);
+        if (matches) {
+          for (var i = matches.length - 1; i >= 0; i--) {
+            // Convert to lowercase, and remove any whitespace
+            // so we can match below.
+            name = matches[i].toLowerCase().replace(/\s/g, "")
+            // multiple authors
+            values[name] = content.trim();
+		  		}
+        }
+      }
+			if (!matches && namePattern.test(elementName)) {
+        name = elementName;
         var content = element.getAttribute("content");
         if (content) {
-          // Convert to lowercase and remove any whitespace
-          // so we can match below.
-          name = name.toLowerCase().replace(/\s/g, "");
+          // Convert to lowercase, remove any whitespace, and convert dots
+          // to colons so we can match below.
+          name = name.toLowerCase().replace(/\s/g, "").replace(/\./g, ":");
           values[name] = content.trim();
         }
       }
     });
 
-    if ("description" in values) {
-      metadata.excerpt = values["description"];
-    } else if ("og:description" in values) {
-      // Use facebook open graph description.
-      metadata.excerpt = values["og:description"];
-    } else if ("twitter:description" in values) {
-      // Use twitter cards description.
-      metadata.excerpt = values["twitter:description"];
+    // get title
+    metadata.title = values["dc:title"] ||
+                     values["dcterm:title"] ||
+                     values["og:title"] ||
+                     values["title"] ||
+                     values["twitter:title"];
+
+		if (!metadata.title) {
+      metadata.title = this._getArticleTitle();
     }
 
-    metadata.title = this._getArticleTitle();
-    if (!metadata.title) {
-      if ("og:title" in values) {
-        // Use facebook open graph title.
-        metadata.title = values["og:title"];
-      } else if ("twitter:title" in values) {
-        // Use twitter cards title.
-        metadata.title = values["twitter:title"];
-      }
-    }
+    // get author
+    metadata.byline = values["dc:creator"] ||
+                      values["dcterm:creator"] ||
+                      values["author"];
+
+    // get description
+    metadata.excerpt = values["dc:description"] ||
+                       values["dcterm:description"] ||
+                       values["og:description"] ||
+                       values["description"] ||
+                       values["twitter:description"];
 
     return metadata;
   },
--- a/Readability.js
+++ b/Readability.js
@@ -1198,7 +1198,7 @@ Readability.prototype = {
     var values = {};
     var metaElements = this._doc.getElementsByTagName("meta");
 
-		// property is a space-separated list of values
+    // property is a space-separated list of values
     var propertyPattern = /\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*/gi;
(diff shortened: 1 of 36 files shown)


gijsk:
Hi! Sorry for the extremely late response. I went on vacation and it seems that our regular triage filter from the folks who took over in my absence includes new issues filed, but not PRs. D'oh!

So I've literally just seen this (I returned to work today). I'll do a review straightaway...

Readability.js
```diff
@@ -1233,20 +1224,39 @@ Readability.prototype = {
...
 
-    if ("description" in values) {
-      metadata.excerpt = values["description"];
+    // get author
+    if ("author" in values) {
+      metadata.excerpt = values["author"];
+    } else if ("dc:author" in values) {
+      // Use dublin core author.
+      metadata.excerpt = values["dc:author"];
+    } else if ("article:author" in values) {
```
gijsk: I'm probably missing something, but it doesn't look to me like a meta tag with `article:author` would pass the regular expression you've used above - I assume the `author` in the first group in the regex should be `article` instead. In fact, if I'm not missing something:

```js
 /^\s*(?:author|dc|og|twitter)\s*:\s*(author|description|title)\s*$/gi;
```
looks to me like it won't match plain `"author"` either, because you've removed the question mark that used to be after the first group.

gijsk: Ah, also, as pointed out by someone else in #480, can you remove the `g` flag from the new regex? It shouldn't be necessary and it causes bugs (like that one).

da2x:
There are some frankly embarrassing bugs here too. Opening pull requests at 4 AM is a bad idea. I’ll fix it. Some tests does needs to be updated too. I’ll check what the effects of removing /g are.

da2x:
Heads up: Some of the tests have og:title values that include the site name in them. They’re not supposed to be there according to the Open Graph Protocol. I checked live versions of these sites, and all but one (Breitbart) have made changes to their websites and have removed their site names from their og:title values. Therefor some of the test cases may look like they’re regressing in terms of article title detection, but these tests don’t reflect how these websites appear today.

da2x:
@gijsk, I believe I should be done now. This is a big behavioral change for metadata extraction. Web developers now have more control over their document titles. I’ve added two new tests in addition to having updated existing tests.

gijsk:
Thanks very much! I'm going to take your advice and not look at this right now at 11.30pm at night, but I promise to have a look tomorrow morning. :-)

test/test-pages/buzzfeed-1/expected-metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "Student Dies After Diet Pills She Bought Online \"Burned Her Up From Within\"",
-  "byline": "Mark Di Stefano",
-  "excerpt": "An inquest into Eloise Parry's death has been adjourned until July...",
+  "byline": null,
```
gijsk: The source has ` <meta property="author" content="Mark Di Stefano"/>`. Can we continue matching that if we have nothing else?

Readability.js
```diff
@@ -1198,62 +1198,66 @@ Readability.prototype = {
...
-    // Match "description", or Twitter's "twitter:description" (Cards)
-    // in name attribute.
-    var namePattern = /^\s*((twitter)\s*:\s*)?(description|title)\s*$/i;
+    // property is a space-separated list of values
+    var propertyPattern = /\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*/gi;
 
-    // Match Facebook's Open Graph title & description properties.
-    var propertyPattern = /^\s*og\s*:\s*(description|title)\s*$/i;
+    // name is a single value
+    var namePattern = /^\s*(?:(dc|dcterm|og|twitter)\s*[\.:]\s*)?(author|creator|description|title)\s*$/gi;
```
gijsk: Can you remove the `g` flag here? Otherwise, I believe it'll reintroduce issue #480 , see the discussion in the first answer here: https://stackoverflow.com/questions/209732/why-am-i-seeing-inconsistent-javascript-logic-behavior-looping-with-an-alert-v - you're using `namePattern` with `.test()` which will keep the last matching index and won't match the next time you run it.

Readability.js
```diff
@@ -1198,62 +1198,66 @@ Readability.prototype = {
     var values = {};
     var metaElements = this._doc.getElementsByTagName("meta");
 
-    // Match "description", or Twitter's "twitter:description" (Cards)
-    // in name attribute.
-    var namePattern = /^\s*((twitter)\s*:\s*)?(description|title)\s*$/i;
+    // property is a space-separated list of values
+    var propertyPattern = /\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*/gi;
```
gijsk: You never use the groups inside this regex, so they can be made non-capturing with `?:` immediately after the opening `(`.

Readability.js
```diff
@@ -1198,62 +1198,66 @@ Readability.prototype = {
...
-    metadata.title = this._getArticleTitle();
     if (!metadata.title) {
-      if ("og:title" in values) {
-        // Use facebook open graph title.
-        metadata.title = values["og:title"];
-      } else if ("twitter:title" in values) {
-        // Use twitter cards title.
-        metadata.title = values["twitter:title"];
-      }
+      metadata.title = this._getArticleTitle();
```
gijsk: There's quite a bit of logic in `_getArticleTitle` to strip out the prefixed/suffixed junk (usually publication name or something like that) that we're now seeing in some of the metadata things. One thing we could potentially do is run that logic against the metadata as well. Doesn't need to happen here (though I wouldn't object if you did it here either) but might be nice as a separate PR once this lands, if you're interested in tidying this up further.

Readability.js
da2x: Fixed.

Readability.js
da2x: This is only a problem with the tests. See [this comment](https://github.com/mozilla/readability/pull/478#issuecomment-415156928). `_getArticleTitle` only really works for English-language content and web developers have no control over it. It should only be used if no better data source available.

Readability.js
da2x: No. We need global matches so we can find all the metadata candidates. It lets the logic find the best data source and use them in a predictable order (Dublin Core first, then Open Graph, then HTML5/native, and lastly Twitter).
https://github.com/mozilla/readability/pull/478/files/5b1dabb1700510e9466e79ea0e03cc8542cbcb8d#diff-06d8d22df421dacde90a2268083424abR1255

da2x: We need the global modifier. But I think I can fix the original issue here without loosing the g.

The global lets us find every metadata candidate so that we can select from the candidates in a defined order (Dublin Core first followed by Open Graph, HTML5 native, and lastly Twitter). This makes the metadata selection process less random to web developers.

gijsk: But we don't use `namePattern` with `match`, only with `test`, which doesn't find more than 1 candidate, only returns true/false, right?

gijsk: That is, I can see you want the global modifier with `propertyPattern`, but not (in the current implementation) with `namePattern.

test/test-pages/buzzfeed-1/expected-metadata.json
da2x: This is a mistake by the document author. The property has no prefix and the element doesn’t have a different namespace. This is an unknown property. The website have removed this tag today and only supply the author through JSON-LD. Adding support for JSON-LD would resolve this and detect authors for many test cases that lack them.

gijsk:
I'd like to merge this, but it seems there are conflicts. If you don't have an opportunity to rebase beforehand, I'll try to do it myself on Monday. Thanks again for the patch!

da2x:
Github claims that “this branch has no conflicts with the base branch”. I’m not sure which conflicts I should be fixing. 🙃

gijsk:
Oh, wow. So the issue is that github's UI is ... not very obvious?

For me it said:

"This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch."

and the "Rebase and merge" button was greyed out.

However, it turns out I could still click the dropdown on that greyed-out-button, and select "squash and merge", which turns the button green and makes the warning disappear. That was... not obvious. I'd have thought if there are conflicts, merging in whatever way wouldn't work... Anyway.

da2x:
GitHub is under new management, you know. 😉

Thanks for merging! Now I’ll need to update [this article](https://www.ctrl.blog/entry/browser-reading-mode-metadata) just a day after publishing it. Or, hm. How long will it be before any browser update their copy of Readability?

FarmaanElahi:
I have found an issue in `propertyPattern` regex. It should end with `$` like 
```
/\s*(dc|dcterm|og|twitter)\s*:\s*(author|creator|description|title)\s*$/gi;
```
If we dont add, it can match with anything start from the group item like `author:image` or `og:author:image`

gijsk:
I'll merge the changes to Mozilla's main repository "soon", I hope. Tomorrow is a public holiday here in the UK, but hopefully I'll get around to it later that week, and then it'll ship as part of Firefox 63, in late October.

Thanks for the link to your articles. I can offer some more updates (e.g. Firefox shipped reader mode on Android as part of Firefox 14, in July 2012 ( https://bugzilla.mozilla.org/show_bug.cgi?id=696921 ), and also offers it on iOS). There's probably more I can comment on if you're interested. Mozilla plan to open source Pocket's code (quickest non-primary link I can find, so it's not just me saying it: https://www.omgubuntu.co.uk/2017/02/mozilla-acquire-pocket-open-source ) - it just seems like it's a slow process. I'm not involved with it, so I don't have any more details myself, but I'll try asking what the plans are for their readability-style code.

da2x:
> There's probably more I can comment on if you're interested.

Corrections and comments about the article are absolutely welcome.

> I'll merge the changes to Mozilla's main repository "soon", I hope. Tomorrow is a public holiday here in the UK, but hopefully I'll get around to it later that week, and then it'll ship as part of Firefox 63, in late October.

That is much earlier than I’d expected. Thanks for the update!

> Thanks for the link to your articles. I can offer some more updates (e.g. Firefox shipped reader mode on Android as part of Firefox 14, in July 2012 ( https://bugzilla.mozilla.org/show_bug.cgi?id=696921 ), and also offers it on iOS).

Firefox for Android’s market share makes it little more interesting than a footnote, though. (I’m saying tht as a Firefox for Android user myself.) I remember there was a working reading list/special bookmarks folder system in Firefox for Android that never made it to sync or desktop.

> Mozilla plan to open source Pocket's code (quickest non-primary link I can find, so it's not just me saying it: https://www.omgubuntu.co.uk/2017/02/mozilla-acquire-pocket-open-source ) - it just seems like it's a slow process. I'm not involved with it, so I don't have any more details myself, but I'll try asking what the plans are for their readability-style code.

I’ve read about the plans to open source Pocket but as far as I can tell it hasn’t been mentioned again in like 18 months. I’m hard on Readability in the article because, frankly, they’re really bad at parsing documents and they don’t offer any details on how publishers can optimize their content for Pocket.

PR #482, opened 2018-08-22
Recognize more iframe video embed video services

da2x:
Added:

* TenCent QQ Video, Alexa Rank 8
* Twitch clips and streams, Alexa Rank 33
* Internet Archive, Alexa Rank 265
* Wikimedia, Alexa Rank 347

[List of even more services](https://en.wikipedia.org/wiki/List_of_video_hosting_services) including some popular omissions: Facebook, VKontakte (Russia), Flickr, and Naver (South Korea). I’ve not got an account with these services so I can’t test their embeds properly.


gijsk:
Thanks! If you're able to update your patch in #478 that would be great, if not, please let me know and I can try to do it myself...

PR #483, opened 2018-08-26
Recognize Sina Weibo meta tags

da2x:
For Chinese content.

http://open.weibo.com/wiki/Weibo_meta_tag

Alexa rank in China: 7, globally: 20 (equivalent to Twitter which is 6 in US)


gijsk:
Thanks! I'll merge this now, though if you have a testcase for this that'd be great.

PR #486, opened 2018-10-17
Add .gitattributes file

Calinou:
This ignores HTML (test data) so the repository is considered to use JavaScript instead of HTML on GitHub.

(Note that this won't display on my fork, since it must be pushed to the `master` branch to be effective.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+**/*.html linguist-detectable=false


msujaws:
@gijsk What do you think? I'm not sure this is something we would normally take but I'm not too active in the Github community to know how important it is that the project appears as HTML or JavaScript. Merging this pull request would cause this file to make its way in to the mozilla-central codebase at some point. I do see that we have .gitattributes files for two third-party imports and one for devtools.

gijsk:
> 
> 
> @gijsk What do you think? I'm not sure this is something we would normally take but I'm not too active in the Github community to know how important it is that the project appears as HTML or JavaScript. Merging this pull request would cause this file to make its way in to the mozilla-central codebase at some point. I do see that we have .gitattributes files for two third-party imports and one for devtools.

We don't need to import it - imports to m-c basically copy JSDOMParser and Readability.js and then revert the warning header we have at the top of the m-c files that say they shouldn't be modified in-place.

So I think we can just merge this. Thank you! :)

Issue #496, opened 2018-11-19
Sorting relies on a boolean return value


jemrobinson:
As described here (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) the `compareFunction` argument to `sort` discriminates based on whether its input is less than, equal to or greater than 0. When passing this a boolean, true and false are evaluated to 1 and 0, with 0 then leaving

> a and b unchanged with respect to each other

This means that the details of the sort function used become important. In the move from Node.js 10.13 to Node.js 11.3 the sort behaviour has changed and thus the output of standalone Readability.js has changed on trivial inputs like:

```
<html>
    <head></head>
    <body>No article here!</body>
</html>
```

With Node 10.13 this returns

`{"title":"","byline":null,"dir":null,"content":"<div id=\"readability-page-1\" class=\"page\">No article here!\n</div>","textContent":"No article here!\n","length":17}`

while with Node 11.1 this returns:

`null`

I have made a pull request which restores the behaviour from 10.13 in 11.1 by changing the `compareFunction`.

gijsk:
Closing thanks to the fix in #497. Thank you!

PR #497, opened 2018-11-19
Switched sort function from boolean to explicit -1 and 1

jemrobinson:
This avoids changes in behaviour between Node.js 10.13 and Node.js 11.1, closing https://github.com/mozilla/readability/issues/496

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1141,7 +1141,7 @@ Readability.prototype = {
           this._attempts.push({articleContent: articleContent, textLength: textLength});
           // No luck after removing flags, just return the longest text we found during the different loops
           this._attempts.sort(function (a, b) {
-            return (a.textLength < b.textLength) ? 1 : -1;
+            return b.textLength - a.textLength;
           });
 
           // But first check if we actually have something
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -1141,7 +1141,7 @@ Readability.prototype = {
           this._attempts.push({articleContent: articleContent, textLength: textLength});
           // No luck after removing flags, just return the longest text we found during the different loops
           this._attempts.sort(function (a, b) {
-            return a.textLength < b.textLength;
+            return (a.textLength < b.textLength) ? 1 : -1;
```
gijsk: Thanks for the patch! How about:

```suggestion
            return b.textLength - a.textLength;
```

Does that also work?

jemrobinson: Yes, that also works. Please feel free to go with whichever you think is clearer.

gijsk: Hm, @jemrobinson seems that I can't accept my own suggestion because I (of course!) don't have permission to push to your clone of the repo - would you mind accepting it so I can just merge this from the web UI? Sorry for the confusion, I'll give github some feedback on this...

jemrobinson: OK, I *think* that should be done now. I also realised that your solution deals with the case `a.textLength == b.textLength` which the original code doesn't, so it's definitely better. Thanks for reviewing this!

gijsk:
Fantastic, thank you!

PR #504, opened 2018-12-21
Add website name feature

maludecks:
Hi!

I am working in a project on my own, and it would be great if I could retrieve also the website name from the article using Readability, since it already gets all the metadata. At first I developed a decorator for it, but I think this could also be useful for more people :)

What do you guys think?

Thank you!


Readability.js
```diff
@@ -1811,6 +1815,7 @@ Readability.prototype = {
     }
 
     var textContent = articleContent.textContent;
+
```
gijsk: Nit: please avoid the extra line here.

maludecks: removed :)

PR #505, opened 2018-12-22
example added

mlda065:
I found it very confusing to figure out how to initialise and use this library. (Especially since `jsdom` is [currently broken](https://github.com/jsdom/jsdom/issues/2429)) So I thought it would be very useful to have an example.


gijsk:
I don't think we want to do this. We really don't want to encourage outside consumers to use `JSDOMParser`. It can only parse XML-compatible HTML DOM, and will fail to work on most of the public web. `jsdom` is a much better, more complete implementation. I don't know why it doesn't work for you; it seems to work in the unit tests for this project and for most other people. Hopefully the maintainers of `jsdom` can help you figure out what's broken there.

I'd be open to an example using `jsdom`, next to the one that uses an in-page document object, but it should probably just live in the readme as a snippet.

PR #509, opened 2019-01-07
Avoid setAttribute errors from invalid attributes, fixes #392

### LINKED ISSUE #392: Uncaught DOMException on https://losst.ru/obnovlenie-debian-9
I am getting

```
Uncaught DOMException: Failed to execute 'setAttribute' on 'Element': 'adsbygoogle.js"' is not a valid attribute name.
    at Readability._setNodeTag (Readability.js:477:19)
    at Readability._grabArticle (Readability.js:774:25)
    at Readability.parse (Readability.js:2062:31)
    at 
```

which refers to

```js
  _setNodeTag: function (node, tag) {
    this.log("_setNodeTag", node, tag);
    if (node.__JSDOMParser__) {
      node.localName = tag.toLowerCase();
      node.tagName = tag.toUpperCase();
      return node;
    }

    var replacement = node.ownerDocument.createElement(tag);
    while (node.firstChild) {
      replacement.appendChild(node.firstChild);
    }
    node.parentNode.replaceChild(replacement, node);
    if (node.readability)
      replacement.readability = node.readability;

    for (var i = 0; i < node.attributes.length; i++) {     /*the next line*/
      replacement.setAttribute(node.attributes[i].name, node.attributes[i].value);
    }
    return replacement;
  },

```

gijsk:
The original markup that trips these exceptions is things that look like `<table width="90% border="0>` which gets you an attribute with name `0`; passing that (or something with a "." or other invalid characters) to `setAttribute` can throw an exception, which we should catch.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -535,7 +535,16 @@ Readability.prototype = {
       replacement.readability = node.readability;
 
     for (var i = 0; i < node.attributes.length; i++) {
-      replacement.setAttribute(node.attributes[i].name, node.attributes[i].value);
+      try {
+        replacement.setAttribute(node.attributes[i].name, node.attributes[i].value);
+      } catch (ex) {
+        /* it's possible for setAttribute() to throw if the attribute name
+         * isn't a valid XML Name. Such attributes can however be parsed from
+         * source in HTML docs, see https://github.com/whatwg/html/issues/4275,
+         * so we can hit them here and then throw. We don't care about such
+         * attributes so we ignore them.
+         */
+      }
     }
     return replacement;
   },


gijsk:
As noted in #392, I tried to use `element.attributes.setNamedItem` first, but I can't `cloneNode` attribute instances in jsdom ( https://github.com/jsdom/jsdom/issues/2478 ) and so I couldn't figure out how to make that work. I think in general we don't care about these attributes anyway (only really about valid HTML attrs), so we might as well drop them on the floor when setting them doesn't work.

PR #511, opened 2019-01-28
update JSDOM example for node

buley:
throws error "First argument to Readability constructor should be a document object" unless you reference the JSDOM `document` object instead of the `window` object per https://github.com/jsdom/jsdom/blob/master/README.md#basic-usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ var JSDOM = require('jsdom').JSDOM;
 var doc = new JSDOM("<body>Here's a bunch of text</body>", {
   url: "https://www.example.com/the-page-i-got-the-source-from",
 });
-let reader = new Readability(doc);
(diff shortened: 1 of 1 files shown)


gijsk:
Is `doc.window` a `document` object? That seems surprising, ie shouldn't it be `dom.window.document` if not `dom.document` ?

buley:
yes, @gijsk, absolutely; sorry I submitted the wrong patch. It's been fixed.

Before: https://gist.github.com/buley/6fe57bca09d8c469425b771866633021

After: https://gist.github.com/buley/f93a5e9fcf4a47a539a4e2b90c67b324

gijsk:
Thanks, and sorry for getting the example wrong in the first place... that's kind of embarrassing...

PR #521, opened 2019-02-23
Check itemprop when determining whether a node is a byline

PalmerAL:
The New York Times uses this markup for its byline:
```
<p class="css-16vrk19 e1jsehar1" itemprop="author creator" itemscope="" itemtype="http://schema.org/Person" itemid="">
By <!-- --><a href="https://www.nytimes.com/by/corey-kilgannon" class="css-1riqqik e1jsehar0"><span class="css-1baulvz" itemprop="name">Corey Kilgannon</span></a>
</p>
```

Because checkByline currently only checks for the `rel` attribute, it doesn't detect that this is a byline. This PR adds a check for `itemprop=author`, which makes the byline detection work correctly.

Doing this also fixes the byline detection on the lemonde testcase, which has similar markup.

This isn't an issue on the existing NYT testcases because the site used to include a `<meta name="author">` tag, which has since been removed. I've added a new testcase and left the testcases using an old version of the site in place, but I'm happy to update those as well if that would be better.


gijsk:
Thanks, this looks great!

PR #523, opened 2019-02-25
Fix tabular data got removed

### LINKED ISSUE #66: We shouldn't strip tables agressively, as they're most often actually part of content

RadhiFadlillah:
This PR should fix #66. 

Previously, in [`_cleanConditionally`](https://github.com/mozilla/readability/blob/master/Readability.js#L1619) we only check if the `node` is inside data table, but we didn't check if the `node` itself is data table.

This PR add a simple check to see if the `node` is table and marked as tabular data.

In other note, I'm not sure if the test files that I prepared is correct, so feels free to correct it if needed.


RadhiFadlillah:
@gijsk I've fixed the tests, hopefully it's correct now.

gijsk:
> @gijsk I've fixed the tests, hopefully it's correct now.

Thanks for the quick update. I'm a little confused; you changed the source files in quite a few cases (and it's making the diff a bit trickier to review). Why did you need to change the source files?

gijsk:
> @gijsk I've fixed the tests, hopefully it's correct now.

@RadhiFadlillah Thanks for the quick update. I'm a little confused; you changed the source files in quite a few cases (and it's making the diff a bit trickier to review). Why did you need to change the source files?

RadhiFadlillah:
@gijsk that's ... a good question :confounded: 

I think I was too focused to make the test passed that I didn't care which file that I changed. Not to mention it's my first time creating unit test for JavaScript project, which make me feel a bit confused about the tooling. I apologize for the trouble.

Right now I've reverted the source files for `lwn-1` and `wikipedia` and make sure all tests have passed successfully.

gijsk:
Thanks, this looks great!

RadhiFadlillah:
@gijsk yep, you're welcome :smile: 

By the way, the old PR #67 could be closed as well since #66 has been solved by this.

PR #527, opened 2019-03-02
Don't remove sharing elements with a large amount of text content

### LINKED ISSUE #466: Only the header is displayed on Mercurial docs
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:60.0) Gecko/20100101 Firefox/60.0
Build ID: 20180605171542

Steps to reproduce:

1. Open https://www.mercurial-scm.org/doc/evolution/sharing.html
2. Enter "read mode" (F9)

Actual results:
The header is displayed.

Expected results:
The whole page should be displayed.

Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1471565

PalmerAL:
Fixes #466.

The entire content of the page is wrapped in a div that contains `share` in the ID:

```
<div class="content">
    <div class="section" id="evolve-shared-mutable-history">
...
```

Currently, any children of the top candidate with "share" in the classname or ID are removed, which results in all of the page content being removed. Since share buttons probably won't normally have a lot of text content, any match with lots of text inside is probably a false-positive, so I think it's safe to assume that the node shouldn't be removed.

The 500-character limit is pretty arbitrary - all the tests pass until it is reduced to around 60 - but I didn't want to diverge too much from the existing behavior.

I also ended up modifying `_cleanMatchedNodes` to allow additional conditions beyond just the regex, although given that this is only used in one place I wonder if it's even worth keeping this (versus using _removeNodes).

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -580,8 +580,8 @@ Readability.prototype = {
     // Clean out elements with little content that have "share" in their id/class combinations from final top candidates,
     // which means we don't remove the top candidates even they have "share".
 
-    var shareElementThreshold = this.DEFAULT_CHAR_THRESHOLD
-    
+    var shareElementThreshold = this.DEFAULT_CHAR_THRESHOLD;
+
     this._forEachNode(articleContent.children, function (topCandidate) {
       this._cleanMatchedNodes(topCandidate, function (node, matchString) {
         return /share/.test(matchString) && node.textContent.length < shareElementThreshold;
--- a/Readability.js
+++ b/Readability.js
@@ -577,10 +577,12 @@ Readability.prototype = {
     this._clean(articleContent, "link");
     this._clean(articleContent, "aside");
 
-    // Clean out elements have "share" in their id/class combinations from final top candidates,
+    // Clean out elements with little content that have "share" in their id/class combinations from final top candidates,
     // which means we don't remove the top candidates even they have "share".
-    this._forEachNode(articleContent.children, function(topCandidate) {
-      this._cleanMatchedNodes(topCandidate, /share/);
+    this._forEachNode(articleContent.children, function (topCandidate) {
+      this._cleanMatchedNodes(topCandidate, function (node, matchString) {
+        return /share/.test(matchString) && node.textContent.length < 500;
+      });
     });
 
     // If there is only one h2 and its text content substantially equals article title,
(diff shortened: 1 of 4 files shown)


Readability.js
```diff
@@ -577,10 +577,12 @@ Readability.prototype = {
...
     this._clean(articleContent, "aside");
 
-    // Clean out elements have "share" in their id/class combinations from final top candidates,
+    // Clean out elements with little content that have "share" in their id/class combinations from final top candidates,
     // which means we don't remove the top candidates even they have "share".
-    this._forEachNode(articleContent.children, function(topCandidate) {
-      this._cleanMatchedNodes(topCandidate, /share/);
+    this._forEachNode(articleContent.children, function (topCandidate) {
+      this._cleanMatchedNodes(topCandidate, function (node, matchString) {
+        return /share/.test(matchString) && node.textContent.length < 500;
```
gijsk: Could we reuse `DEFAULT_CHAR_THRESHOLD` here? (Note: `_removeNodes`, unlike `_cleanMatchedNodes`, keeps `this` the same when applying the filter function so we can just access `this.DEFAULT_CHAR_THRESHOLD`, I think - but worth doublechecking.)

Readability.js
```diff
@@ -577,10 +577,12 @@ Readability.prototype = {
     this._clean(articleContent, "link");
     this._clean(articleContent, "aside");
 
-    // Clean out elements have "share" in their id/class combinations from final top candidates,
+    // Clean out elements with little content that have "share" in their id/class combinations from final top candidates,
     // which means we don't remove the top candidates even they have "share".
-    this._forEachNode(articleContent.children, function(topCandidate) {
-      this._cleanMatchedNodes(topCandidate, /share/);
+    this._forEachNode(articleContent.children, function (topCandidate) {
+      this._cleanMatchedNodes(topCandidate, function (node, matchString) {
```
gijsk: Yeah, if this is the only use we might as well switch it over to `_removeNodes` and remove `_cleanMatchedNodes`.

PalmerAL: I think `_removeNodes` won't work as-is, because it takes a nodeList, and all of the children need to be checked recursively here. We could either add a new `_getNodeChildren` method (like _getNodeAncestors) that gets all the children recursively, and then pass the result to `_removeNodes`, or we could just move `_cleanMatchedNodes` into the loop (but that seems kind of messy):

```
this._forEachNode(articleContent.children, function (topCandidate) {
      var endOfSearchMarkerNode = this._getNextNode(topCandidate, true);
      var next = this._getNextNode(topCandidate);
      while (next && next != endOfSearchMarkerNode) {
        if (/share/.test(next.className + " " + next.id) && next.textContent.length < this.DEFAULT_CHAR_THRESHOLD) {
          next = this._removeAndGetNext(next);
        } else {
          next = this._getNextNode(next);
        }
      }
    });
```

gijsk: Oh, gah, sorry for missing that. Yeah, then let's leave it as you have it in the patch. @PalmerAL can we still go with using the constant? You can probably just store a copy as a local var in the outer function.

PalmerAL: Sounds fine to me; I've added that in 3a2d258bbb2ae7f1efa02e569fa3a45b2cebff31.

Readability.js
PalmerAL: Done.

gijsk:
Thanks!

PR #530, opened 2019-03-06
Fix table header missing

RadhiFadlillah:
As title said, this PR fixed the missing table header from #66.

I've also updated the expected result in test file. This is because in my previous PR #523, I create the test file manually instead of using the [node script](https://github.com/mozilla/readability/blob/master/test/generate-testcase.js), so I update it here.


gijsk:
Thanks!

Issue #533, opened 2019-03-17
document the install procedure


SvetlanaTkachenko:
how to install using npm ( nodejs ); how to install for usage from javascript - please add this to the documentation

gijsk:
There's no npm package atm (see #43 ). I don't think adding instructions for using github to install it is useful - hopefully we can sort out the NPM situation soon instead. Otherwise, people just end up stuck on old versions of the package sourced via github (or use `master` and get constant breakage, or something).

For non-node envs, there's no real "installation"... you load the script file the same way you load any other script file, and then you use it as described in the docs, so I don't see what other explanations are needed...

SvetlanaTkachenko:
Thank you for describing the situation. I am glad there will be an npm version possible soon.

How do I install the stable version today? Could you please provide me with installation instructions that work at the current time?

gijsk:
https://medium.com/@jonchurch/use-github-branch-as-dependency-in-package-json-5eb609c81f1a may be helpful.

There isn't a stable version today; If you're experimenting with things I'd suggest either using master as the branch tag and accepting breakage, or picking the current revision and updating when necessary/desired.

SvetlanaTkachenko:
@gijsk 

I don't have package.json, which makes the article you linked more confusing.

I am looking for the same 'stable' version as ships with Firefox for its reading view.

I face confusion here. I understand that 'master' includes breakage, but I don't see what is the definition of 'current revision', or how to use it.

gijsk:
I don't know how to help, sorry. The `npm install` instruction at the top of the article seems like it should work, irrespective of whether you have a package.json . If it doesn't, I don't know enough about npm to help.

For the "stable" version that ships with Firefox - we import from 'master' on this repository, and fix any issues that come up. The latest import (that will ship with Firefox 67) is from what is 'master' a few days ago, see https://bugzilla.mozilla.org/show_bug.cgi?id=1531712 . (Compared to current 'master', we're only missing a test-only commit, ie https://github.com/mozilla/readability/commit/cb5771fd4a832bf257bf1b86f7fbdcb74068d3c5 .

PR #535, opened 2019-03-24
added npm install instructions from github

SvetlanaTkachenko:
From https://github.com/mozilla/readability/issues/533

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ var article = new Readability(documentClone).parse();
 
 ## Usage from node.js
 
+To install this library using npm, run
+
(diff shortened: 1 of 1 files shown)


gijsk:
So, as I said in #533, I don't want to document this method. We should get the NPM package published the normal way. Telling people to import from github is a recipe for confusion and things being broken when the developer least expects it, because github will change.

People who want to use this package directly from github can already do so (and will hopefully know the risks involved), but we should not encourage this in the documentation.

wong2:
when will the npm package be published the normal way? been waiting for too long..

Issue #299, opened 2016-07-14
Images missing from medium.com articles


designakt:
Any images shown in medium.com articles are lost when read via Reader View.
examples:
https://medium.com/firefox-ux/this-is-our-reading-list-c81c4238dd3d
https://medium.com/the-year-of-the-looking-glass/quality-is-not-a-tradeoff-bcddf7c85553
https://medium.com/@joulee/metrics-versus-experience-a9347d6b80b

![readerviewimagesmedium](https://cloud.githubusercontent.com/assets/8145891/16836277/119af5c4-49bd-11e6-9f07-d79dbd2ba074.gif)

tigt:
Medium does this awful lazy-loading strategy, where their image markup looks like this (whitespace added for readability):

```html
<figure class="graf--figure graf--leading">
  <div class="aspectRatioPlaceholder is-locked" style="max-width: 699px; max-height: 392px;">
    <div class="aspectRatioPlaceholder-fill" style="padding-bottom: 56.10000000000001%;"></div>
    <div class="progressiveMedia js-progressiveMedia graf-image" data-image-id="0*9lG_11ELAmuIaO-Z.jpeg" data-width="699" data-height="392">
      <img src="https://cdn-images-1.medium.com/freeze/max/30/0*9lG_11ELAmuIaO-Z.jpeg?q=20" crossorigin="anonymous" class="progressiveMedia-thumbnail js-progressiveMedia-thumbnail">
      <canvas class="progressiveMedia-canvas js-progressiveMedia-canvas"></canvas>
      <img class="progressiveMedia-image js-progressiveMedia-image" data-src="https://cdn-images-1.medium.com/max/800/0*9lG_11ELAmuIaO-Z.jpeg">
      <noscript class="js-progressiveMedia-inner">
        <img class="progressiveMedia-noscript js-progressiveMedia-inner" src="https://cdn-images-1.medium.com/max/800/0*9lG_11ELAmuIaO-Z.jpeg">
      </noscript>
    </div>
  </div>
  <figcaption class="imageCaption">From the movie “Jiro Dreams of Sushi”</figcaption>
</figure>
```

Since the only real `<img src>` in there is a 20-pixel-wide file, or wrapped in a `<noscript>`, this might take some doing.

designakt:
I wonder how Firefox on Android solves this.
They show the picture in their Reader View: 

[...]

PR #542, opened 2019-05-01
Add support for detecting lazy-loaded images

PalmerAL:
This PR adds the ability to detect some lazy-loaded images and replace them with regular HTML based images that can be loaded inside reader view. Specifically, it:

* Checks for image elements that have an attribute that isn't an `src` or `srcset`, but looks like one, and replaces the `src`/`srcset` with that value (similar to what [mercury-parser does](https://github.com/postlight/mercury-parser/blob/0940971069290b5bce4dc9422d8b4d0d20f7d3b5/src/resource/utils/dom/convert-lazy-loaded-images.js)).
* Checks for `<figure>` elements that have an attribute with a value that looks like an src/srcset, but that don't contain any image nodes. In this case, a new image will be created and added to inside the `<figure>`

The first change fixes missing images on these articles:

* https://www.seattletimes.com/business/amazon/alaskan-halibut-caught-by-a-century-old-seattle-boat-provides-a-glimpse-of-amazons-strategy-with-whole-foods/
* https://www.citylab.com/design/2019/04/neon-signage-20th-century-history/588400/

The second change fixes missing images on the existing nytimes-3 testcase.

My original intent was to make this fix medium as well (#299), but doing so requires removing "media" from the negative candidates regex, which causes a bunch of other test failures; I might look into that later at some point if I have time.

Currently the nytimes-3 testcase is failing; it appears that jsdom and jsDOMParser disagree on how to escape attribute values, so the test ends up comparing `&` to `&amp;` inside the image URL. I'm not really sure which one is right, @gijsk do you have any ideas on how to fix this?


Readability.js
```diff
@@ -1616,6 +1618,44 @@ Readability.prototype = {
...
+            //if this is an img or picture, set the attribute directly
+            if (elem.tagName === "IMG" || elem.tagName === "PICTURE") {
+              elem.setAttribute(copyTo, attr.value)
+            } else {
+              //if the item is a <figure> that does not contain an image or picture, create one and place it inside the figure
+              //see the nytimes-3 testcase for an example
+              if (elem.tagName === "FIGURE" && elem.getElementsByTagName("img").length === 0 && elem.getElementsByTagName("picture").length === 0) {
+                var img = this._doc.createElement("img")
+                img.setAttribute(copyTo, attr.value)
+                elem.appendChild(img)
```
PalmerAL: It would be better to make the image the first child in the figure, but that would require using `insertBefore`, which JSDOMParser doesn't support.

gijsk:
Thanks for the PR!

For the `&amp;` issue... If the markup has `&amp;` then the `attribute.value` (which it looks like we're using?) should be "text" rather than "html", ie the `&amp;` should be rendered as `&`, as far as I can tell. I don't know what that means in terms of whether JSDOMParser or jsdom is wrong. In the former case, we can fix it, in the latter, we could file an issue, I guess?

For convenience in terms of this test failure, I'd just cheat and strip the `&amp` bits from the source and expected values - they don't matter for testing that readability works.

gijsk:
Oh, also, fair warning, someone will need to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1546235 before the `srcset` side of things will work properly in Firefox, unfortunately. :-(

Readability.js
PalmerAL: Yes, the NYT markup has a caption and some other elements inside the figure, so the caption ends up being above the image. Adding insertBefore in a separate PR sounds fine to me.

Readability.js
```diff
@@ -1616,6 +1618,44 @@ Readability.prototype = {
...
   },
 
+  /* convert images and figures that have properties like data-src into images that can be loaded without JS */
+  _fixLazyImages: function(root) {
+    var nodesToCheck = [].slice.call(root.getElementsByTagName("img"))
+    .concat([].slice.call(root.getElementsByTagName("picture")))
+    .concat([].slice.call(root.getElementsByTagName("figure")))
+
+    this._forEachNode(nodesToCheck, function(elem) {
+      if ((!elem.src && !elem.srcset) || elem.className.indexOf("lazy") !== -1) {
```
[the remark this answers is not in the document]
PalmerAL: > Nit: could we toLowerCase the className so we catch Lazy as well?

Done.

> Also, are the src and srcset properties ever defined in the JSDOMParser case? I would kind of expect them not to be, but I could be wrong...

`src` is defined here: https://github.com/mozilla/readability/blob/master/JSDOMParser.js#L661, but it looks like `srcset` isn't, I'll go ahead and fix that.

Readability.js
```diff
@@ -1616,6 +1618,44 @@ Readability.prototype = {
...
+        for (var i = 0; i < elem.attributes.length; i++) {
+          var attr = elem.attributes[i]
+          if(attr.name === "src" || attr.name === "srcset") {
```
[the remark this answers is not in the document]
PalmerAL: I think I fixed all the instances of this, let me know if I missed one.

Readability.js
```diff
@@ -1616,6 +1618,44 @@ Readability.prototype = {
...
+            continue;
+          }
+          var copyTo = null;
+          if(/.*\.(jpg|png|bmp|tiff)\s\d+/.test(attr.value)) {
```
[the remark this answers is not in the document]
PalmerAL: I think you're right, there's nothing in the testcases that includes a bmp or tiff, so I've removed them (and added the ones you mentioned). srcset should work with multiple spaces, I've fixed that as well.

Readability.js
```diff
@@ -1616,6 +1618,44 @@ Readability.prototype = {
...
+            copyTo = "src"
+          }
+          if(copyTo) {
+            //if this is an img or picture, set the attribute directly
+            if (elem.tagName === "IMG" || elem.tagName === "PICTURE") {
+              elem.setAttribute(copyTo, attr.value)
+            } else {
+              //if the item is a <figure> that does not contain an image or picture, create one and place it inside the figure
+              //see the nytimes-3 testcase for an example
+              if (elem.tagName === "FIGURE" && elem.getElementsByTagName("img").length === 0 && elem.getElementsByTagName("picture").length === 0) {
```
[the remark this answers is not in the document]
PalmerAL: `_getAllNodesWithTag` is nice, I didn't realize that existed before.

The rest of readability seems to mostly use `===` already (55 instances of === vs 14 instances of ==), so I'm not sure it's worth changing this, but I'm happy to do so it you want me to.

ESLint might work here, although I'm not sure if it has rules for everything or not.

PalmerAL:
> The only thing I'm a little worried by is the difference in terms of protocol requirements in the src vs. srcset regexes.

Removing the protocol requirement for `src` doesn't cause any regressions in the testcases, and fixes a missing image in `mozilla-1`, so I'm guessing it would be fine to do. (However, not many of the tests include lazy-loading images in the first place, so I'm not sure how well this would actually work).

The encoding issue looks like a JSDOMParser bug: the attribute is encoded immediately when it is set, but only `getAttribute()` will decode it -  `attribute.value` just returns the encoded value, which is incorrect. I think the correct fix for this is to save the attribute in un-encoded form, and only encode it when it's needed for `innerHTML`; does that make sense to you?

For the project I'm working on, I'm loading the reader content in a sandboxed iframe where `srcset` should work, so I'd like to leave it included here if that's OK with you.

gijsk:
Sorry for the slow response, I was hoping to deal with this on Friday or over the weekend, but then things went a bit mad ( https://blog.mozilla.org/addons/2019/05/04/update-regarding-add-ons-in-firefox/ ).

> > The only thing I'm a little worried by is the difference in terms of protocol requirements in the src vs. srcset regexes.
> 
> Removing the protocol requirement for `src` doesn't cause any regressions in the testcases, and fixes a missing image in `mozilla-1`, so I'm guessing it would be fine to do. (However, not many of the tests include lazy-loading images in the first place, so I'm not sure how well this would actually work).

Worth a shot, it seems.

> The encoding issue looks like a JSDOMParser bug: the attribute is encoded immediately when it is set, but only `getAttribute()` will decode it - `attribute.value` just returns the encoded value, which is incorrect. I think the correct fix for this is to save the attribute in un-encoded form, and only encode it when it's needed for `innerHTML`; does that make sense to you?

That sounds right based on your description.

> For the project I'm working on, I'm loading the reader content in a sandboxed iframe where `srcset` should work, so I'd like to leave it included here if that's OK with you.

Very much OK; I was just concerned that if you were doing this hoping to get it to work in Firefox, there's additional work that needs doing so wanted to avoid disappointment. :-)

Readability.js
```diff
@@ -1616,6 +1618,40 @@ Readability.prototype = {
...
+          if (attr.name === "src" || attr.name === "srcset") {
+            continue;
+          }
+          var copyTo = null;
+          if (/\.(jpg|jpeg|png|webp)\s+\d/.test(attr.value)) {
+            copyTo = "srcset"
+          } else if (/https?:\/\/.*\.(jpg|jpeg|png|webp)/.test(attr.value)) {
+            copyTo = "src"
+          }
+          if(copyTo) {
```
gijsk: Nit: space before "("

Readability.js
```diff
@@ -1616,6 +1618,40 @@ Readability.prototype = {
...
+      if ((!elem.src && !elem.srcset) || elem.className.toLowerCase().indexOf("lazy") !== -1) {
+        for (var i = 0; i < elem.attributes.length; i++) {
+          var attr = elem.attributes[i]
+          if (attr.name === "src" || attr.name === "srcset") {
+            continue;
+          }
+          var copyTo = null;
+          if (/\.(jpg|jpeg|png|webp)\s+\d/.test(attr.value)) {
+            copyTo = "srcset"
+          } else if (/https?:\/\/.*\.(jpg|jpeg|png|webp)/.test(attr.value)) {
```
gijsk: Yeah, let's omit the protocol. Given this is specifically the `src` matching, perhaps we could check that there is no whitespace in the bit of the attribute with a filename, ie use something like `/^\s*\S+\.(jpg|jpeg|png|webp)\s*$/` (which allows for leading/trailing whitespace but otherwise expects exactly one URL) ?

PalmerAL: Sounds good to me; I modified your suggested regex slightly to allow non-whitespace characters immediately after the filetype so that query strings still work.

In the process of updating the mozilla testcase for this, I ran into an interesting bug in jsdom: if you have an image with an `src` but no `srcset`, and you try to read the `srcset` property, jsdom will return the string `"null"` instead of undefined. For now, I've just added a check to work around that, but maybe it's worth adding a comment explaining what's happening?

PalmerAL:
> Sorry for the slow response, I was hoping to deal with this on Friday or over the weekend, but then things went a bit mad ( https://blog.mozilla.org/addons/2019/05/04/update-regarding-add-ons-in-firefox/ ).

No worries! Things have definitely been fairly crazy with Firefox over the past few days; hopefully you have a chance to relax for a bit now.

I think I've fixed the JSDOMParser issue and the other issues you've mentioned; all the tests are passing now. It'd be nice it this worked in Firefox eventually, but I don't know enough about how the sanitizer code works to fix that, so I'm going to leave that to someone else.

JSDOMParser.js
```diff
@@ -494,19 +494,14 @@
     get value() {
       return this._value;
     },
+    getValue: function() {
```
gijsk: Hm, there's already a `.value` getter, do we really need to define `getValue` as well?

Readability.js
```diff
@@ -1616,6 +1618,38 @@ Readability.prototype = {
     }
   },
 
+  /* convert images and figures that have properties like data-src into images that can be loaded without JS */
+  _fixLazyImages: function (root) {
+    this._forEachNode(this._getAllNodesWithTag(root, ["img", "picture", "figure"]), function (elem) {
+      if ((!elem.src && (!elem.srcset || elem.srcset == "null")) || elem.className.toLowerCase().indexOf("lazy") !== -1) {
```
gijsk: Yeah, a comment here would be good. I'm sure the jsdom people would appreciate you filing a bug if you have time, and then we could reference that here. If not, let me know and I can do it.

JSDOMParser.js
PalmerAL: Probably not, I just figured it would be better to be consistent with `setValue`, although I suppose that could be a setter instead.

Readability.js
PalmerAL: Good point, I've opened an issue here: https://github.com/jsdom/jsdom/issues/2580.

gijsk:
Thanks very much!

ldenoue:
Other sites like https://www.technologyreview.com/s/614057/china-squirrel-has-started-a-grand-experiment-in-ai-education-it-could-reshape-how-the/ wrap content to be loaded inside NOSCRIPT elements.
We could loop through them, set their textContent as innerHTML of a fake DIV and extract the IMG elements from there.
That’s wha I used for my app ReaderView https://www.appblit.com/readerview

PalmerAL:
@ldenoue I guess more generally, it might make sense to add an option to keep noscript tags (which are currently removed here: https://github.com/mozilla/readability/blob/master/Readability.js#L1311). This would probably have to be an option, since Firefox uses the existing page content when you switch to reader view (meaning scripts would already have a chance to run, so noscript tags shouldn't be important), but if you're just fetching the article HTML and sending it directly to readability it seems like it would make sense to keep them.

ldenoue:
Even with the existing page loaded in Firefox, this page for example https://www.technologyreview.com/s/614057/china-squirrel-has-started-a-grand-experiment-in-ai-education-it-could-reshape-how-the/ doesn’t touch the noscript tags until the user scrolls close to them.

So readability wouldn’t get the images in them.

Without touching the DOM, when we build the readable version, we could look at them and get images out.

PR #545, opened 2019-05-21
Remove sharing elements in fewer situations

### LINKED ISSUE #534: Missing images on The Guardian
On https://www.theguardian.com/us-news/2019/mar/22/uber-lyft-ipo-drivers-unionize-low-pay-expenses, the two images inside the article are removed.

The reason is that the images are inside an element with "share" in the class, resulting in the same issue as #466, and don't have sufficient text content to be fixed by #527:

```
<figure itemprop="associatedMedia image" itemscope="" itemtype="http://schema.org/ImageObject" data-component="image" class="element element-image img--landscape  fig--narrow-caption fig--has-shares " data-media-id="f2a65a345e22d705bfc5c65a780f42f8ac2f7286" id="img-2">
...
```

I see a couple of potential solutions to this:
* Decrease the share element threshold.
    * The figure in the article has 260 characters of text, so 250 might be a reasonable threshold.
* Make the removal of sharing elements also depend on other factors (tag name, etc.)
* Try to come up with a more specific list of selectors, rather than just "share", since that seems to match a lot of things that aren't actually sharing-related.

@gijsk What do you think?

PalmerAL:
With this change, elements will only be removed if the word "share" occurs between two non-word characters (or the beginning or end), meaning that things like `share` or `share-button` will be removed, but `has-shares` (like the Guardian testcase) will not.

This fixes #534, but regresses #117. I think that's an acceptable tradeoff, but if it's necessary I could also add an exception for `sharedaddy`.


Readability.js
```diff
@@ -586,7 +586,7 @@ Readability.prototype = {
 
     this._forEachNode(articleContent.children, function (topCandidate) {
       this._cleanMatchedNodes(topCandidate, function (node, matchString) {
-        return /share/.test(matchString) && node.textContent.length < shareElementThreshold;
+        return /[^a-zA-Z0-9]share[^a-zA-Z0-9]/g.test(" " + matchString + " ") && node.textContent.length < shareElementThreshold;
```
gijsk: Could this just be `/\bshare\b/` to look for word breaks? Or does that not work? Also, I assume the regex should be case insensitive - could you add the `i` flag?

PalmerAL: Good point! `\b` alone won't work because it doesn't check for underscores, but I've tried to simplify this a bit and fix the case issue in 55892033bc5148d1f1ae350b0957136f4c49cab4.

PR #552, opened 2019-07-31
Added "keepClasses" option to prevent cleaning of classes

jvandenaardweg:
This PR adds a new option `keepClasses` to prevent cleaning of classes.

Use:
```javascript
new readability(document, { keepClasses: true }).parse()
```

When not passing `keepClasses` it will fallback to the default, which is cleaning of the classes. So this PR is backward compatible.

**The use cases**
I have these use cases where I use readability to get the article from a website and then further process the article's HTML. I've noticed some elements that are unwanted (for my use case) are still in the `content` which I want to be stripped out. However, I cannot clean the document's complete HTML using my blacklist of selectors as that might strip out the article before giving it to readability.
Using `classesToPreserve` does not work for me, as I need it the other way around; I have a list of blacklisted elements I want to have stripped out.

Prevent the cleaning of classes allows me to further process the article with the classes still intact inside `content`.


test/test-readability.js
```diff
@@ -201,6 +201,12 @@ describe("Readability API", function() {
       expect(new Readability(doc)._maxElemsToParse).eql(0);
       expect(new Readability(doc, {maxElemsToParse: 42})._maxElemsToParse).eql(42);
     });
+
+    it("should accept a keepClasses option", function() {
```
gijsk: Can you also add a test that checks that the option actually works?

jvandenaardweg: Sure, done! 👍

jvandenaardweg:
> (I only had a brief look, I'm away until August 12, sorry for the delay.)
> 
> I'm fine adding this option, but please can you add a test for the option working, and could you please remove the added package-lock.json for now? I'm not clear on why it has various babel bits etc., and if we do need/want to add one, it should probably be in a separate commit.

I've added a few tests and removed the package-lock.json 👍

PR #555, opened 2019-08-28
Remove aria-hidden="true" nodes (fixes #541)

### LINKED ISSUE #541: Items with aria-hidden="true" (e.g. font awesome icons) should be removed
Filed based on https://bugzilla.mozilla.org/show_bug.cgi?id=1545453 .

We should be able to remove all these nodes.

jwinett:
Fixes #541

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -31,7 +31,7 @@ var REGEXPS = {
 function isNodeVisible(node) {
   // Have to null-check node.style to deal with SVG and MathML nodes.
   return (!node.style || node.style.display != "none") && !node.hasAttribute("hidden")
-		&& (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") !== "true");
+    && (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") !== "true");
 }
 
 /**
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -30,7 +30,8 @@ var REGEXPS = {
 
 function isNodeVisible(node) {
   // Have to null-check node.style to deal with SVG and MathML nodes.
-  return (!node.style || node.style.display != "none") && !node.hasAttribute("hidden");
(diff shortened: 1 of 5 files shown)


Readability-readerable.js
```diff
@@ -30,7 +30,8 @@ var REGEXPS = {
 
 function isNodeVisible(node) {
   // Have to null-check node.style to deal with SVG and MathML nodes.
-  return (!node.style || node.style.display != "none") && !node.hasAttribute("hidden");
+  return (!node.style || node.style.display != "none") && !node.hasAttribute("hidden")
+		&& (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") !== "true");
```
gijsk: Nitpick: Could you use spaces instead of tabs, please? I thought eslint covered this, but clearly I was wrong and need to fix that...

Also, could we just use `node.getAttribute("aria-hidden") != "true"` ? If the node doesn't have the attribute, the value will be null, so the comparison still works.

Readability.js
```diff
@@ -1786,7 +1786,10 @@ Readability.prototype = {
   },
 
   _isProbablyVisible: function(node) {
-    return (!node.style || node.style.display != "none") && !node.hasAttribute("hidden");
+    return (!node.style || node.style.display != "none")
+	  && !node.hasAttribute("hidden")
+	  && (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") !== "true")
```
gijsk: Same comment here as above.

Readability-readerable.js
gijsk: Hm, odd. I just pulled your branch, and there's definitely actual tabs... It looks like they get picked up if I add the eslint rule `no-tabs`; perhaps the existing indenting rules are OK with them because they're indenting the second and later lines of the statement or something.

gijsk:
Thanks!

Readability-readerable.js
gijsk: It worked, thanks again!

PR #560, opened 2019-09-28
Fix missing content on Wikipedia articles

PalmerAL:
The "foot" keyword in unlikelyCandidates matches "footnote" on Wikipedia, which results in some headings and citations being incorrectly removed from the output. Changing this to "footer" fixes the issue, and seems closer to the original intention of the filter.


Readability.js
```diff
@@ -115,7 +115,7 @@ Readability.prototype = {
   REGEXPS: {
     // NOTE: These two regular expressions are duplicated in
     // Readability-readerable.js. Please keep both copies in sync.
```
gijsk: Please also update Readability-readerable.

PalmerAL:
No problem! There's a couple different things that get incorrectly removed:

* Some of the citation markers:
```
<sup id="cite_ref-FOOTNOTEMein_Smith200523_37-0"><a href="#cite_note-FOOTNOTEMein_Smith200523-37">[31]</a></sup>
```
* The actual "footnotes" heading:
```
                <span id="Footnotes">Footnotes</span>
```

* Some of the citations at the end:
```
	                    <li id="cite_note-FOOTNOTEKennedy2007398-304">
                        <span><b><a href="#cite_ref-FOOTNOTEKennedy2007398_304-0">^</a></b></span> <span><a href="#CITEREFKennedy2007">Kennedy 2007</a>, p.&#160;398.</span>
                    </li>
```

This change fixes all 3. Here's a link to the full diff of the file: https://www.diffchecker.com/iaer3auO

It looks like only citations to books (and not websites) have "footnote" in the ID, which is why you only see this on certain pages (and only some pages have a "footnotes" header).

PR #562, opened 2019-10-17
Proper null check on removeChild

jvandenaardweg:
Fixing an error I was seeing: Cannot read property 'removeChild' of null

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -706,7 +706,9 @@ Readability.prototype = {
 
   _removeAndGetNext: function(node) {
     var nextNode = this._getNextNode(node, true);
-    node.parentNode.removeChild(node);
+    if (node.parentNode) {
+      node.parentNode.removeChild(node);  
+    }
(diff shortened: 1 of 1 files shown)


gijsk:
Can you provide a testcase that shows this error? I think it indicates something went wrong somewhere else and would prefer to fix that problem over adding a nullcheck - nodes should always have a parent node, unless they're the document element, which we shouldn't be removing.

gijsk:
(or unless they've already been removed from the doc, in which case, why are we removing them again?)

jvandenaardweg:
It did happen on a document with almost no text and just a few elements. Maybe that gives a better idea on the error case.

> Can you provide a testcase that shows this error? I think it indicates something went wrong somewhere else and would prefer to fix that problem over adding a nullcheck - nodes should always have a parent node, unless they're the document element, which we shouldn't be removing.

I'll try to see if i can recreate this error.

PR #578, opened 2020-02-27
Add exception for wikimedia math images

### LINKED ISSUE #561: Wikipedia math rendering broken after #555
Example URL: https://en.wikipedia.org/wiki/Hermitian_matrix

Wikipedia renders math equations as MathML, followed by an image like the following:

```
<img src="https://wikimedia.org/api/rest_v1/media/math/render/svg/6ca00b61ff0e264e6c1e5adc9a00c0d2751feecf" class="mwe-math-fallback-image-inline" aria-hidden="true" style="vertical-align: -0.338ex; width:32.194ex; height:3.509ex;" alt="{\displaystyle A{\text{ Hermitian}}\quad \iff \quad A={\overline {A^{\mathsf {T}}}}}"/>
```

Readability always removes the MathML nodes (I haven't investigated to find out why, but since they're not supported outside of Firefox, it's probably better not to rely them anyway), leaving the image as the only way to display the equation. Prior to #555, the image would be included in the readability output, but since the image has `aria-hidden`, it will now be removed, so there's no way to see the included math.

I don't see a good general solution to this problem, but Wikipedia is a popular enough site that I'm thinking it would be fine to just create a workaround for this particular instance - maybe look for "fallback image" and assume it should be visible.

PalmerAL:
Fixes #561

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -31,9 +31,9 @@ var REGEXPS = {
 function isNodeVisible(node) {
   // Have to null-check node.style and node.className.indexOf to deal with SVG and MathML nodes.
   return (!node.style || node.style.display != "none")
-  && !node.hasAttribute("hidden")
-  //check for "fallback-image" so that wikimedia math images are displayed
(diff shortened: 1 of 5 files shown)


gijsk:
Thank you *very* much!

PR #581, opened 2020-02-28
Fix #580 by not using live node lists when removing items

### LINKED ISSUE #580: Readability freezes with 100% CPU on specific URL
```typescript
import jsdom from 'jsdom';
import Readability from 'mozilla-readability';

const url = 'https://www.sec.gov/Archives/edgar/data/1794515/000162828020002344/zoominfos-1.htm';

jsdom.JSDOM.fromURL(url, {
  resources: new jsdom.ResourceLoader()
}).then(doc => {
  const reader = new Readability(doc.window.document);

  return reader.parse(); // this freezes @ 100% CPU
});
```

Node.js 12
jsdom 16.1.0

gijsk:
It turns out that having a 15k live node list which we remove items from 1 at a time, which then causes us to recompute the length and all the members, is really slow. This change:

- makes us use `querySelectorAll` in places that support it (e.g. jsdom), which returns a static node list (as opposed to the live node list from `getElementsByTagName`), by using the `_getAllNodesWithTag` helper. In the JSDOMParser case (ie in Firefox), it'll return an array, which is also static.
- makes the node removal/replacement helpers throw errors if you pass the result of `getElementsByTagName` to them, to avoid hitting this problem when we implement other removals.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -308,6 +308,7 @@
       }
     }
     getElems(this);
+    elems._isLiveNodeList = true;
     return elems;
   }
 
--- a/Readability.js
+++ b/Readability.js
@@ -36,6 +36,7 @@ function Readability(doc, options) {
   options = options || {};
 
   this._doc = doc;
+  this._docJSDOMParser = this._doc.firstChild.__JSDOMParser__;
   this._articleTitle = null;
   this._articleByline = null;
   this._articleDir = null;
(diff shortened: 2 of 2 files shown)


JSDOMParser.js
```diff
@@ -308,6 +308,7 @@
       }
     }
     getElems(this);
+    elems._noLiveNodeListsPlease = true;
```
msujaws: Can you rename this to `isLiveNodeList`? Booleans that start with `no` are confusing when negated, and really this boolean doesn't describe the policy (like it's named) but it describes the behavior of this list.

PR #588, opened 2020-03-26
Fix low-res images in Medium pages that uses lazy loading

RadhiFadlillah:
Several months ago, an [issue](https://github.com/go-shiori/shiori/issues/230) is submitted to my project that uses `go-readability`, the port of `readability.js` for Go. I think I've fixed it there, so might as well put it here as well.

The summary of that issue is, Medium pages lazy load the image using JavaScript. However, they also provide the high quality image inside `noscript` tag in case the user disabled the JS in their browser. Unfortunately, right now Readability removes all `noscript` tags without exception.

By the way, this issue is only happened when Readability is used in server, and works fine in Firefox reader mode.

---

To fix this issue, I decide to unwrap `img` from inside `noscript` if it fulfills two criteria :

- the previous direct element sibling of the `noscript` is `img`; and
- `noscript` only contains exactly one `img` element.

To prevent wrong unwrapping, I also did these steps before unwrapping the `noscript` :

- Find all `div` that contains exactly one image, then put it out outside the `div`. This is done to 
prevent case where a `noscript` has direct previous sibling of a `div` that doesn't have any use except containing a single `img`.
- Find all `img` that doesn't have any sources. This is done, because in Medium they use empty `img` as placeholder. If it doesn't removed, after unwrapping we will ended up with two image, both the low quality and high quality image.

For sample, I use this Medium [article](https://medium.com/voodoo-engineering/node-js-and-cpu-profiling-on-production-in-real-time-without-downtime-d6e62af173e2) and here is the result :

<table>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77654732-03f61500-6fa4-11ea-8394-a941e4501867.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77654888-3b64c180-6fa4-11ea-86bb-a098cd63288e.png"></a></td>
    </tr>
</table>

---

Unfortunately, the test is failed for 5 pages :

- bug-1255978
- citylab-1
- mozilla-1
- seattletimes-1
- yahoo-1

However, from those failed tests, all changes is only happened in its DOM structure, yet visually there are (almost) no change as can be seen below :

<table>
    <tr valign="top" align="center">
        <td colspan="2"><b>bug-1255978</b></td>
    </tr>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77655481-0c028480-6fa5-11ea-92d5-f368435d9ab9.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77655529-1a50a080-6fa5-11ea-9cf4-0d352b7547f4.png"></a></td>
    </tr>
</table>

<table>
    <tr valign="top" align="center">
        <td colspan="2"><b>citylab-1</b></td>
    </tr>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77655651-4a983f00-6fa5-11ea-9537-c5faa30718b1.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77655712-58e65b00-6fa5-11ea-9501-a63744c13d3b.png"></a></td>
    </tr>
</table>

<table>
    <tr valign="top" align="center">
        <td colspan="2"><b>mozilla-1</b></td>
    </tr>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657152-774d5600-6fa7-11ea-89de-928d34d3d3d0.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657218-8fbd7080-6fa7-11ea-8238-d72a6c1336fd.png"></a></td>
    </tr>
</table>

<table>
    <tr valign="top" align="center">
        <td colspan="2"><b>seattletimes-1</b></td>
    </tr>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657284-acf23f00-6fa7-11ea-84df-bdb4ebd1bdde.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657329-bbd8f180-6fa7-11ea-91e6-2aa0acee87b9.png"></a></td>
    </tr>
</table>

<table>
    <tr valign="top" align="center">
        <td colspan="2"><b>yahoo-1</b></td>
    </tr>
    <tr align="center">
        <td><b>Current Readability.js</b></td>
        <td><b>After PR</b></td>
    </tr>
    <tr valign="top" align="center">
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657457-e88d0900-6fa7-11ea-8a8d-1983ba59b98a.png"></a></td>
        <td width="50%"><a href="https://user-images.githubusercontent.com/6129042/77657487-f5a9f800-6fa7-11ea-8a66-6b3ca32b9ff5.png"></a></td>
    </tr>
</table>

---

So, what do you think ? If you think it's fine, I will update the expected result for those tests later.


gijsk:
Hm, the yahoo testcase seems to miss images in your screenshot?

I don't mind the small markup changes, but I'd prefer to see the diffs of the resulting `expected.html` cases (which you can regenerate with `npm run test/generate-testcase.js <slug>` ) as part of this patch.

RadhiFadlillah:
@gijsk I've updated the test for those pages.

> I'm a bit confused because the yahoo testcase seems to miss images in your screenshot?

No, it just Puppeeter being weird when taking screenshot. If you view it in browser normally, the Youtube image is still there.

test/test-pages/mozilla-1/expected.html
```diff
@@ -2,65 +2,76 @@
...
             </div>
         </section>
         <section id="designed" data-ga-label="Designed to be redesigned">
             <div>
                 <div id="designed-copy">
-                    <h2>Designed to <br/>be redesigned</h2>
+                    <h2>Designed to <br />be redesigned</h2>
                     <p>Get fast and easy access to the features you use most in the new menu. Open the “Customize” panel to add, move or remove any button you want. Keep your favorite features — add-ons, private browsing, Sync and more — one quick click away.</p>
-                    <p><img src="http://mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned-high-res.6efd60766484.png" data-processed="false" data-src="//mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned.fbd3ee9402e6.png" data-high-res="true" data-high-res-src="//mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned-high-res.6efd60766484.png" alt="" id="designed-mobile" /> </p>
+                    <p><img src="http://mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/designed-redesigned.fbd3ee9402e6.png" alt="" id="designed-mobile" />
```
gijsk: So it's hard to know what would happen here today, because the original images no longer load and the page has disappeared off google, but the file names here indicate that we're now loading the mobile version of the image instead of the high-res one. That's the opposite of what you wanted to accomplish, right?

I'm not sure there's a way to really solve both cases, short of trying to come up with heuristics to determine which URL is better - or creating some kind of pluggability where the consumer of the library can fetch both and determine which they want... which seems tricky, because it more or less entails starting to use `await` and promises in the library, which is its own project.

Could we keep the old/other source available on the image in a separate attribute, so that consumers of the resulting markup could still make their own decisions later? That might be an easier way of giving control to the library consumers without needing too many architectural changes inside readability.

test/test-pages/mozilla-1/expected.html
```diff
@@ -2,65 +2,76 @@
...
-                        <p><a rel="external" href="https://addons.mozilla.org/firefox/themes/">Try it now</a> <br/> <a rel="external" href="https://support.mozilla.org/kb/use-themes-change-look-of-firefox">Learn more</a> </p>
+                        <p><a rel="external" href="https://addons.mozilla.org/firefox/themes/">Try it now</a>
+                            <br /> <a rel="external" href="https://support.mozilla.org/kb/use-themes-change-look-of-firefox">Learn more</a>
+                        </p>
                     </div>
-                    <p><a href="#add-ons" role="button">Next</a> </p>
-                    <p><img id="theme-demo" src="http://mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/theme-red.61611c5734ab.png" alt="Preview of the currently selected theme" /> </p>
+                    <p><a href="#add-ons" role="button">Next</a>
+                        <img id="theme-demo" src="http://mozorg.cdn.mozilla.net/media/img/firefox/desktop/customize/theme-red.61611c5734ab.png" alt="Preview of the currently selected theme" />
```
gijsk: This image used to be in its own `<p>` and now it's inline, which is probably not the right thing in the circumstances. Can we avoid this somehow? Perhaps we can check if the div/p which contains only the image and which we're removing has inline element or text siblings?

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+   * @param Element
+  **/
+  _unwrapNoscriptImages: function(doc) {
+    // First, find div which only contains single img element, then put it out.
+    var divs = doc.getElementsByTagName("div");
+    this._forEachNode(divs, function(div) {
+      if (div.children.length == 1 && div.children[0].tagName === "IMG") {
+        div.parentNode.replaceChild(div.children[0], div);
+      }
+    });
```
gijsk: I think this doesn't work correctly, because `getElementsByTagName` returns a live Node list, and you're then removing some of these nodes. On jsdom and other "proper" DOM implementations (unlike `JSDOMParser.js` in the readability repo), this would potentially skip some of the items.

I also think this is a little suboptimal from a performance perspective. In particular, I think there are normally fewer `img` elements than there are `div`s in the document, and so it would be more efficient to look for images, and then see if their parent is a div with no other children (and no inline siblings, see my other comment). Then you can also unify this loop with the next loop:

```js
var imgs = Array.from(doc.getElementsByTagName("img"));
this._forEachNode(imgs, function(img) {
  var src = img.getAttribute("src");
  var srcset = img.getAttribute("srcset");
  var dataSrc = img.getAttribute("data-src");
  var dataSrcset = img.getAttribute("data-srcset");

  if (!src && !srcset && !dataSrc && !dataSrcset) {
    img.parentNode.removeChild(img);
  } else if (img.parentNode.tagName == "DIV" && !img.nextElementSibling && !img.previousElementSibling) {
    var div = img.parentNode;
    // FIXME check for phrasing content that isn't empty/whitespace.
    div.parentNode.replaceChild(div, img);
  }
});
```

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+    // Next find img without source, and remove it. This is done to
+    // prevent a placeholder img is replaced by img from noscript in next step.
+    var imgs = doc.getElementsByTagName("img");
+    this._forEachNode(imgs, function(img) {
+      var src = img.getAttribute("src") || "",
+        srcset = img.getAttribute("srcset") || "",
+        dataSrc = img.getAttribute("data-src") || "",
+        dataSrcset = img.getAttribute("data-srcset") || "";
+      
+      if (src === "" && srcset === "" && dataSrc === "" && dataSrcset === "") {
```
gijsk: Nit: as suggested above, please use individual declarations. You don't need the `|| ""` or the `=== ""` because all you care about is whether the attribute value is truthy or not.

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
     return metadata;
   },
 
+  /**
+   * Find all <noscript> that located after <img> node, and contains exactly
+   * single <img> element. Once it found, this method will replace the previous <img>
+   * with <img> inside <noscript>, then finally remove the <noscript> tag. This is
+   * done because in some website (e.g. Medium), they use lazy load method like this.
+   *
```
gijsk: Some English nits here. Can you use:

```
Find all <noscript> that are located after <img> nodes, and which contain only one <img> element.
Replace the first image with the image from inside the <noscript> tag, and remove the <noscript> tag.
This improves the quality of the images we use on some sites (e.g. Medium).
```

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+      
+      if (src === "" && srcset === "" && dataSrc === "" && dataSrcset === "") {
+        img.parentNode.removeChild(img);
+      }
+    });
+
+    // Next find noscript and try to extract its image
+    var noscripts = doc.getElementsByTagName("noscript");
+    this._forEachNode(noscripts, function(noscript) {
+      // Make sure prev sibling is exist and it's image
```
gijsk: `exists` rather than `is exist`

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+      var prevElement = noscript.previousElementSibling;
+      if (prevElement == null || prevElement.tagName !== "IMG") {
+        return;
+      }
+      
+      // In jsdom content of noscript is treated as string, so here we parse it.
```
gijsk: This isn't just jsdom, it's any spec-compliant browser (Chrome, Firefox, etc.) - see https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody:scripting-flag . Let's have the comment reflect that.

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+      var tmp = doc.createElement("div");
+      tmp.innerHTML = noscript.innerHTML;
+
+      // Make sure noscript only has one children, and it's <img> element
```
gijsk: Nit: "one child", not "one children"

Readability.js
```diff
@@ -1316,6 +1316,61 @@ Readability.prototype = {
...
+        dataSrcset = img.getAttribute("data-srcset") || "";
+      
+      if (src === "" && srcset === "" && dataSrc === "" && dataSrcset === "") {
+        img.parentNode.removeChild(img);
+      }
+    });
+
+    // Next find noscript and try to extract its image
+    var noscripts = doc.getElementsByTagName("noscript");
+    this._forEachNode(noscripts, function(noscript) {
```
gijsk: Here too I'd expect off-by-one errors where not all the noscript tags get correctly replaced...

Readability.js
RadhiFadlillah: Fixed in [dc33dd6](https://github.com/mozilla/readability/pull/588/commits/dc33dd69793dc0af6e80b02e68af84a1609dd31b)

Readability.js
RadhiFadlillah: Fixed in [dc33dd6](https://github.com/mozilla/readability/pull/588/commits/dc33dd69793dc0af6e80b02e68af84a1609dd31b)

Readability.js
RadhiFadlillah: Fixed in [dc33dd6](https://github.com/mozilla/readability/pull/588/commits/dc33dd69793dc0af6e80b02e68af84a1609dd31b)

Readability.js
RadhiFadlillah: Fixed in [dc33dd6](https://github.com/mozilla/readability/pull/588/commits/dc33dd69793dc0af6e80b02e68af84a1609dd31b)

Readability.js
RadhiFadlillah: Fixed in [38bcb08](https://github.com/mozilla/readability/pull/588/commits/38bcb08e0c1c31f384d56c8b2c3098a07237dda1).

test/test-pages/mozilla-1/expected.html
RadhiFadlillah: It's fixed in [38bcb08](https://github.com/mozilla/readability/pull/588/commits/38bcb08e0c1c31f384d56c8b2c3098a07237dda1).

test/test-pages/mozilla-1/expected.html
RadhiFadlillah: Done in [d32f39f ](https://github.com/mozilla/readability/pull/588/commits/d32f39f6c6e73c4edff91fb958b4faf4c549e5ff)

Readability.js
RadhiFadlillah: It should be fixed by [1a2a0ab](https://github.com/mozilla/readability/pull/588/commits/1a2a0ab2262cc3f86e1f0fe72f12166abd08812d)

gijsk:
Apologies for the delay here; it's code freeze crunch time and I'm stretched a little thin at the moment. I hope to be able to look at this later today, worst case tomorrow.

test/test-pages/mozilla-1/expected.html
gijsk: I'm a bit puzzled because the image for which this conversation was started still only has 1 attribute?

gijsk: Looks like this conflicts with the code from #542 in `_fixLazyImages`, because your patch is replacing the original image with the one from inside the `noscript` element, without preserving the `data-src` etc. attributes.

I'll add a suggestion on how to fix this.

Readability.js
```diff
@@ -1316,6 +1316,89 @@ Readability.prototype = {
...
+
+        if (prevImgSrc && prevImgSrc !== newImgSrc) {
+          newImg.setAttribute("data-old-src", prevImgSrc);
+        }
+
+        if (prevImgSrcset && prevImgSrcset !== newImgSrcset) {
+          newImg.setAttribute("data-old-srcset", prevImgSrcset);
+        }
+
+        noscript.parentNode.replaceChild(tmp.children[0], prevElement);
```
gijsk, line 1397: Can we add code along the lines of:

```js
for (var i = 0; i < prevImg.attributes.length; i++) {
  var attr = prevImg.attributes[i];
  if (attr.name == "src" || attr.name == "srcset" || newImg.hasAttribute(attr.name)) {
    continue;
  }
  if (/\.(jpg|jpeg|png|webp)/i).test(attr.value)) {
    newImg.setAttribute(attr.name, attr.value);
  }
}
```

here so we preserve image-related attributes?

RadhiFadlillah: I've added it in [be67f18](https://github.com/mozilla/readability/pull/588/commits/be67f18f88c08501372c809032e768f9092683db)

test/test-pages/citylab-1/expected.html
```diff
@@ -12,7 +12,7 @@
             <meta itemprop="width" content="300" />
             <meta itemprop="url" content="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" />
             <picture>
-                <source srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" media="(max-width: 575px)" /><img srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" alt="" itemprop="contentUrl" onload="performance.mark(&quot;citylab_lead_image_loaded&quot;)" /></picture>
+                <source srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" media="(max-width: 575px)" /><img src="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" alt="" data-old-srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" /></picture>
```
RadhiFadlillah, line 15: I just realized this is wrong. The source for this part is look like this :

```html
<picture class="c-lead-image__image">
    <source srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" media="(max-width: 575px)" />
    <img class="c-lead-image__img" srcset="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" alt="" itemprop="contentUrl" onload="performance.mark(&quot;citylab_lead_image_loaded&quot;)" />
    <noscript>
        <img class="c-lead-image__img" src="https://cdn.citylab.com/media/img/citylab/2019/04/mr1/300.jpg?mod=1556645448" alt="" />
    </noscript>
</picture>
```

As can be seen, the old image has `srcset` attribute while the new one doesn't have it. Therefore the old `srcset` from previous image should be set as `srcset` attribute in new image instead of `data-old-srcset`.

RadhiFadlillah: Fixed in [75e0edc](https://github.com/mozilla/readability/pull/588/commits/75e0edc763dccaad74358983102b11cd7d1beb1b)

gijsk:
Looks great, thank you!

PalmerAL:
I'm having a bit of trouble getting this to work as expected; when I run it against a Medium page, the output includes both the low-res and the high-res images.

If I regenerate the lazy-image-1 testcase, I get a different source file than what's in this PR, so it's possible that Medium's changed their site in the past few days:



@RadhiFadlillah I'll try to look into fixing this soon if you don't have time to.

PR #606, opened 2020-07-25
Replace a Dead Link in Comment

SNikhill:
So, I was just reviewing JSDOMParser.js and came upon two dead links. I have replaced on of them and couldn't find a replacement for the other. (2nd Link is just above the .join() method. LINE: 708)

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -278,7 +278,7 @@
 
   var whitespace = [" ", "\t", "\n", "\r"];
 
-  // See http://www.w3schools.com/dom/dom_nodetype.asp
+  // See https://www.w3schools.com/jsref/prop_node_nodetype.asp
   var nodeTypes = {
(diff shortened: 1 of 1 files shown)


JSDOMParser.js
```diff
@@ -278,7 +278,7 @@
 
   var whitespace = [" ", "\t", "\n", "\r"];
 
-  // See http://www.w3schools.com/dom/dom_nodetype.asp
+  // See https://www.w3schools.com/jsref/prop_node_nodetype.asp
```
gijsk, line 281: Can you replace this with https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType instead, and remove the other link entirely? Thank you!

SNikhill: Upon your request, one of the links has been updated and the second one has been removed.

PR #608, opened 2020-07-26
Readability on npm

danburzo:
Here's some initial work on getting Readability ready for publishing on npm (#43).

In this version, Node.js code would require things like this:

```js
let { Readability, isProbablyReaderable } = require('@mozilla/readability');
...
```

...but `isProbablyReaderable` can just as well be added as a static property on the `Readability` object, if that style is preferable.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -875,7 +875,11 @@
 
   JSDOMParser.prototype = {
     error: function(m) {
-      dump("JSDOMParser error: " + m + "\n");
+      if (typeof dump !== "undefined") {
+        dump("JSDOMParser error: " + m + "\n");
+      } else if (typeof console !== "undefined") {
+        console.log("JSDOMParser error: " + m + "\n");
+      }
       this.errorState += m + "\n";
     },
 
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -1187,3 +1187,7 @@
   global.JSDOMParser = JSDOMParser;
 
 })(this);
+
+if (typeof module === 'object') {
+  module.exports = this.JSDOMParser;
+}
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -1188,6 +1188,6 @@
 
 })(this);
 
-if (typeof module === 'object') {
+if (typeof module === "object") {
   module.exports = this.JSDOMParser;
 }
--- a/index.js
+++ b/index.js
@@ -1,29 +1,7 @@
-var path = require("path");
-var fs = require("fs");
-var url = require("url");
-
-// We want to load Readability and JSDOMParser, which aren't set up as commonjs libraries,
-// and so we need to do some hocus-pocus with 'vm' to import them on a separate scope
-// (identical) scope context.
-var vm = require("vm");
-var readabilityPath = path.join(__dirname, "Readability.js");
-var jsdomPath = path.join(__dirname, "JSDOMParser.js");
-
-
-var scopeContext = {};
-// We generally expect dump() and console.{whatever} to work, so make these available
-// in the scope we're using:
-scopeContext.dump = console.log;
-scopeContext.console = console;
-scopeContext.URL = url.URL;
-
-// Actually load files. NB: if either of the files has parse errors,
-// node is dumb and shows you a syntax error *at this callsite* . Don't try to find
-// a syntax error on this line, there isn't one. Go look in the file it's loading instead.
-vm.runInNewContext(fs.readFileSync(jsdomPath), scopeContext, jsdomPath);
-vm.runInNewContext(fs.readFileSync(readabilityPath), scopeContext, readabilityPath);
+let Readability = require('./Readability');
+let isProbablyReaderable = require('./Readability-readerable');
 
 module.exports = {
-  Readability: scopeContext.Readability,
-  JSDOMParser: scopeContext.JSDOMParser
+	Readability: Readability,
+	isProbablyReaderable: isProbablyReaderable
 };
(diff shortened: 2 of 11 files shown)


index.js
```diff
@@ -1,29 +1,7 @@
...
-var readabilityPath = path.join(__dirname, "Readability.js");
-var jsdomPath = path.join(__dirname, "JSDOMParser.js");
-
-
-var scopeContext = {};
-// We generally expect dump() and console.{whatever} to work, so make these available
-// in the scope we're using:
-scopeContext.dump = console.log;
-scopeContext.console = console;
-scopeContext.URL = url.URL;
```
danburzo, line 18: The [URL class](https://nodejs.org/api/url.html#url_class_url) has been added to Node 10.

danburzo:
Starting with [`percollate@v.0.6.0`](https://github.com/danburzo/percollate) I've removed the patched version of Readability included with the codebase, and using it off this branch directly:

```json
{
  "dependencies": {
    "@mozilla/readability": "github:danburzo/readability#npm"
  }
}
```

```js
let Readability = require('@mozilla/readability');
```

gijsk:
Apologies for not responding here yet - I'm on PTO this week, and the folks whom I normally ask to respond to readability/reader mode things in my stead while I'm away are both on parental leave. I hope to get back to this (and your other PR) on Monday.

danburzo:
@gijsk, no worries, enjoy your time off!

test/generate-testcase.js
```diff
@@ -8,10 +8,9 @@ var http = require("http");
 var urlparse = require("url").parse;
 var htmltidy = require("htmltidy2").tidy;
 
-var readabilityCheck = require("../Readability-readerable");
-var readability = require("../index");
-var Readability = readability.Readability;
-var JSDOMParser = readability.JSDOMParser;
+var readabilityCheck = require("../index").isProbablyReaderable;
+var Readability = require("../index").Readability;
```
gijsk, line 12: Wouldn't it be preferable to `require` only once? I guess it doesn't matter much as it's just a developer script, but if we're changing it anyway... I *think* these versions of node support destructuring assignment, so maybe we can just do:

```js
var { Readability, isProbablyReaderable } = require("../index");
```
and use `isProbablyReaderable` as a name in the one place where we're using it (which you're having to update anyway, in the next hunk) ?

index.js
danburzo: Yes, the required node version should be reflected in `package.json`, if we decide that Node 10 is a good minimum version. Looking at [the releases](https://github.com/nodejs/release#release-schedule), it seems aligned with their support policy. 

Alternatively, something like:

```js
if (typeof URL === 'undefined' && typeof require === 'function') {
  URL = require('url').URL;
}
```

would allow us to use the legacy URL parsing functionality in earlier Node.js versions.

package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "readability",
+  "name": "@mozilla/readability",
```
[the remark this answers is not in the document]
danburzo: This makes it a [scoped package](https://docs.npmjs.com/misc/scope). We need to publish it as such because there's already a [`readability`](https://www.npmjs.com/package/readability) package on npm. 

Users install scoped packages like any other package, using the full name `@mozilla/readability`. On the publishing side, it only requires a few extra things:

* your npm user needs to be added to the [`mozilla`](https://www.npmjs.com/org/mozilla) organization
* your first publish command needs to be `npm publish --access public` (i.e. explicitly opt into making the package public, which is otherwise the default for unscoped packages).

test/generate-testcase.js
danburzo: Yes, seems object destructuring [has been available in Node 6+](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), so can use that. I've tried to touch the code as little as possible in the PR, but you're right this is a good time to switch.

index.js
```diff
@@ -1,29 +1,7 @@
...
-// (identical) scope context.
-var vm = require("vm");
-var readabilityPath = path.join(__dirname, "Readability.js");
-var jsdomPath = path.join(__dirname, "JSDOMParser.js");
-
-
-var scopeContext = {};
-// We generally expect dump() and console.{whatever} to work, so make these available
-// in the scope we're using:
-scopeContext.dump = console.log;
```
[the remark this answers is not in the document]
danburzo: `Readability.js` already falls back to `console.log` in the absence of `dump`, and I'd make `JSDOMParser` behave similarly. I actually wanted to ask about all the other contexts in which Readability needs to work, to see if these are covered.

test/generate-testcase.js
danburzo: This now breaks at linting stage, because eslint defaults to ES5. Since destructuring is only used in tests, I'm inclined to change it back to ES5.

JSDOMParser.js
```diff
@@ -875,7 +875,11 @@
 
   JSDOMParser.prototype = {
     error: function(m) {
-      dump("JSDOMParser error: " + m + "\n");
+      if (typeof dump !== "undefined") {
```
gijsk, line 878: My kingdom for an eslint rule that complains about the use of triple equals, but I guess I just have to accept that I have lost that fight with the internet... I won't bother you with my old-man-get-off-my-lawn-yelling, I should probably blog about it at some point.

index.js
gijsk: Making the requirement 10.0 seems fine to me. Thanks!

JSDOMParser.js
danburzo: I'm in the triples-everywhere camp because of past traumas :P But yeah, an eslint rule that points out unnecessary triples would be awesome.

danburzo:
Cool! I've just enabled ES6 support in eslint.

Issue #488, opened 2018-10-22
byline not extracted for many sites eg medium


8enmann:
https://medium.com/@pimterry/host-your-node-app-on-dokku-digitalocean-1cb97e3ab041

gijsk:
> Maybe we can [resolve a swathe of byline bugs](https://github.com/mozilla/readability/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+byline) 

We already use meta tags, and I wouldn't be surprised if a bunch of those issues are now resolved... it's been a while since anyone's had a chance to take a broom through them.

> by referring to relevant <meta> tags in the head?

In this specific case: Which meta tags? I only see `<meta property="author" ... >` which according to https://github.com/mozilla/readability/pull/478#discussion_r212388935 isn't valid. It may be necessary to re-add support if many sites rely on this, even if it's not technically in a spec somewhere. Maybe @da2x wants to weigh in.

danburzo:
> We already use meta tags, and I wouldn't be surprised if a bunch of those issues are now resolved... it's been a while since anyone's had a chance to take a broom through them.

Oops, sorry. Somehow missed that.

For the original link posted to this thread: https://medium.com/@pimterry/host-your-node-app-on-dokku-digitalocean-1cb97e3ab041 — this regressed in Firefox 63, if that gives you any clue. I coincidentally had 62 up to a few minutes ago, and I could see the byline.

gijsk:
Yes, the change in #478 removed the `<meta property=author` support because of the discussion in the comment I linked to. :-)

danburzo:
Ah, got it. As per [this spec](https://www.w3.org/TR/html5/document-metadata.html#standard-metadata-names) `author` seems to be a valid meta tag, though, so I don't see a problem with reading it. Maybe @da2x can provide some more details on the matter.

8enmann:
@danburzo I'm a little lost. Anything I can do to help? Is there a path forward?

danburzo:
My proposal, that would fix the problem with Medium articles, is to read the author from:

```html
<meta property="article:author" content="...">
```

The [`article` namespace](http://ogp.me/#type_article) corresponds to the Open Graph protocol.

danburzo:
Ugh, sorry. I was looking at the generated DOM, e.g.

```
document.querySelector('meta[property="article:author"]').content
```

It didn't even cross my mind that someone would alter these tags at run-time...

da2x:
[...]

PR #609, opened 2020-08-05
Prefer JSON-LD metadata object, when present

danburzo:
The idea was discussed in #488.

In this implementation, we're reading the JSON-LD metadata for the [Schema.org `Article` type](https://schema.org/Article) and its subtypes.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1316,55 +1316,55 @@ Readability.prototype = {
    * @return Object with any metadata that could be extracted (possibly none)
    */
   _getJSONLD: function (doc) {
-    
+
     var scripts = this._getAllNodesWithTag(doc, ["script"]);
 
     var jsonLdElement = this._findNode(scripts, function(el) {
-      return el.getAttribute('type') === 'application/ld+json';
+      return el.getAttribute("type") === "application/ld+json";
     });
 
     if (jsonLdElement) {
       try {
         // Strip CDATA markers if present
-        var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[\s*|\s*\]\]>\s*$/g, '');
+        var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[\s*|\s*\]\]>\s*$/g, "");
         var parsed = JSON.parse(content);
         var metadata = {};
         if (
-          !parsed['@context'] ||
-          !parsed['@context'].match(/^https?\:\/\/schema\.org$/)
+          !parsed["@context"] ||
+          !parsed["@context"].match(/^https?\:\/\/schema\.org$/)
         ) {
           return metadata;
         }
 
-        if (!parsed['@type'] && Array.isArray(parsed['@graph'])) {
-          parsed = parsed['@graph'].find(it =>
-            (it['@type'] || '').match(
+        if (!parsed["@type"] && Array.isArray(parsed["@graph"])) {
+          parsed = parsed["@graph"].find(function(it) {
+            return (it["@type"] || "").match(
               this.REGEXPS.jsonLdArticleTypes
-            )
-          );
+            );
+          });
         }
 
         if (
           !parsed ||
-          !parsed['@type'] ||
-          !parsed['@type'].match(this.REGEXPS.jsonLdArticleTypes)
+          !parsed["@type"] ||
+          !parsed["@type"].match(this.REGEXPS.jsonLdArticleTypes)
         ) {
           return metadata;
         }
-        if (typeof parsed.name === 'string') {
+        if (typeof parsed.name === "string") {
           metadata.title = parsed.name;
-        } else if (typeof parsed.headline === 'string') {
+        } else if (typeof parsed.headline === "string") {
           metadata.title = parsed.headline;
         }
-        if (parsed.author && typeof parsed.author.name === 'string') {
+        if (parsed.author && typeof parsed.author.name === "string") {
           metadata.byline = parsed.author.name;
         }
-        if (typeof parsed.description === 'string') {
+        if (typeof parsed.description === "string") {
           metadata.excerpt = parsed.description.trim();
         }
         if (
           parsed.publisher &&
-          typeof parsed.publisher.name === 'string'
+          typeof parsed.publisher.name === "string"
         ) {
           metadata.siteName = parsed.publisher.name;
         }
@@ -1450,7 +1450,7 @@ Readability.prototype = {
                       values["author"];
 
     // get description
-    metadata.excerpt = jsonld.excerpt || 
+    metadata.excerpt = jsonld.excerpt ||
                        values["dc:description"] ||
                        values["dcterm:description"] ||
                        values["og:description"] ||
@@ -1460,7 +1460,7 @@ Readability.prototype = {
                        values["twitter:description"];
 
     // get site name
-    metadata.siteName = jsonld.siteName || 
+    metadata.siteName = jsonld.siteName ||
                         values["og:site_name"];
 
     // in many sites the meta value is escaped with HTML entities,
--- a/Readability.js
+++ b/Readability.js
@@ -1370,8 +1370,7 @@ Readability.prototype = {
         }
         return metadata;
       } catch (err) {
-        // ignore malformed JSON-LD
-        console.error(err);
+        this.log(err.message);
       }
     }
     return {};
--- a/Readability.js
+++ b/Readability.js
@@ -53,7 +53,7 @@ function Readability(doc, options) {
   this._serializer = options.serializer || function(el) {
     return el.innerHTML;
   };
-  this._preferJSONLD = !!options.preferJSONLD || true;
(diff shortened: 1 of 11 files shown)


Readability.js
```diff
@@ -53,6 +53,7 @@ function Readability(doc, options) {
   this._serializer = options.serializer || function(el) {
     return el.innerHTML;
   };
+  this._preferJSONLD = !!options.preferJSONLD || true;
```
danburzo, line 56: Temporarily left on `true`, to gauge its effects on the testcases.

test/test-readability.js
```diff
@@ -158,23 +158,23 @@ function runTestsWithItems(label, domGenerationFn, source, expectedContent, expe
     });
 
     it("should extract expected title", function() {
-      expect(expectedMetadata.title).eql(result.title);
+      expect(result.title).eql(expectedMetadata.title);
```
danburzo, line 161: In the testsuite, the metadata checks had their expected/actual values reversed.

Readability.js
```diff
@@ -244,6 +247,21 @@ Readability.prototype = {
...
+   * the supplied test function
+   *
+   * For convenience, the current object context is applied to the provided
+   * test function.
+   *
+   * @param  NodeList nodeList The NodeList.
+   * @param  Function fn       The test function.
+   * @return void
+   */
+  _findNode: function(nodeList, fn) {
```
danburzo, line 261: Added support for yet another array method on NodeLists. If we're okay with the browser support for [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) (which I notice is already in the codebase), we could just cast all NodeLists to arrays and renounce these wrapper functions altogether.

Readability.js
```diff
@@ -1292,12 +1310,81 @@ Readability.prototype = {
...
+    var scripts = this._getAllNodesWithTag(doc, ["script"]);
+
+    var jsonLdElement = this._findNode(scripts, function(el) {
+      return el.getAttribute('type') === 'application/ld+json';
+    });
+
+    if (jsonLdElement) {
+      try {
+        // Strip CDATA markers if present
+        var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[\s*|\s*\]\]>\s*$/g, '');
```
danburzo, line 1329: When parsing the content as HTML, `jsdom` (unlike `JSDOMParser`) leaves CDATA markers as part of the content and breaks JSON parsing, so we need to strip them.

Readability.js
```diff
@@ -2029,12 +2120,15 @@ Readability.prototype = {
     // Unwrap image from noscript
     this._unwrapNoscriptImages(this._doc);
 
+    // Extract JSON-LD metadata before removing scripts
+    var jsonLd = this._preferJSONLD ? this._getJSONLD(this._doc) : {};
```
danburzo, line 2124: Security: Will `<script>` elements be available here, when Readability is integrated in Firefox?

Readability.js
```diff
@@ -1370,7 +1460,8 @@ Readability.prototype = {
                        values["twitter:description"];
 
     // get site name
-    metadata.siteName = values["og:site_name"];
+    metadata.siteName = jsonld.siteName ||
+                        values["og:site_name"];
 
     // in many sites the meta value is escaped with HTML entities,
     // so here we need to unescape it
```
danburzo, line 1467: Security: Does using the JSON-LD values directly, and/or unescaping their HMTL entities cause any security concern in the Firefox integration?

gijsk: I think right now, we expect all these values to be plaintext. Although there won't be a security issue if they include HTML, it'll look bad because we assign to `textContent`. I *think* this is correct for all the existing uses of metadata; if JSON-LD is the "odd one out", we should probably strip the markup out?

Readability.js
```diff
@@ -1292,12 +1310,81 @@ Readability.prototype = {
     });
   },
 
+  /**
+   * Try to extract metadata from JSON-LD object.
+   * For now, only Schema.org objects of type Article or its subtypes are supported.
+   * @return Object with any metadata that could be extracted (possibly none)
+   */
+  _getJSONLD: function (doc) {
+
```
gijsk, line 1319: Nit: don't start/end blocks with empty lines.

I really need to take some time and adopt most of the Firefox eslint rules for this repo...

Readability.js
gijsk: Hm, that's an interesting one - my understanding is that `Array.from` will create a copy, which IMO is only really worth it when you know you're going to modify the contents of the list (e.g. by removing elements from the DOM) which messes up live nodelist iteration otherwise. We're not doing that here, so I'd prefer to avoid the copy for now.

I do agree that it causes clutter to have so many wrappers... perhaps there's a way to clean this up in a separate PR; we could e.g. dynamically create these wrappers from a list of methods we're interested in using. Something vaguely like:

```js
for (let method of ["find", "forEach", "some", "every"]) {
  Readability.prototype["_" + method + "Node"] = function(nodelist, fn) {
    return Array.prototype[method].call(nodeList, fn, this);
  };
}
```

or something.

Readability.js
gijsk: Yes, they're only removed on the next line, when the script tags are all removed. We use JSDOMParser in a worker, and so the DOM is inert because JSDOMParser doesn't run script at all. I forget exactly how JSDOMParser treats script internals, but if the tests you have that rely on the contents of those tags pass in node.js, they should work in Firefox, too.

test/test-readability.js
```diff
@@ -279,7 +279,7 @@ describe("Test pages", function() {
 
       runTestsWithItems("jsdom", function(source) {
         var doc = new JSDOM(source, {
-          url: uri,
+          url: uri
```
gijsk, line 282: Nit: we prefer trailing commas for multiline object declarations. Again, I should really sort out or eslint setup...

test/test-readability.js
gijsk: D'oh! Thanks for fixing.

Readability.js
```diff
@@ -1292,12 +1310,81 @@ Readability.prototype = {
...
+    var scripts = this._getAllNodesWithTag(doc, ["script"]);
+
+    var jsonLdElement = this._findNode(scripts, function(el) {
+      return el.getAttribute("type") === "application/ld+json";
+    });
+
+    if (jsonLdElement) {
+      try {
+        // Strip CDATA markers if present
+        var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[\s*|\s*\]\]>\s*$/g, "");
```
gijsk, line 1329: This replaces whitespace around the CDATA bits as well. Is that necessary? I would have assumed that the JSON parsing would take care of that - or are there CDATA things *inside* the JSON? _*shudder*_

Readability.js
```diff
@@ -1292,12 +1310,81 @@ Readability.prototype = {
...
+        if (typeof parsed.name === "string") {
+          metadata.title = parsed.name;
+        } else if (typeof parsed.headline === "string") {
+          metadata.title = parsed.headline;
+        }
+        if (parsed.author && typeof parsed.author.name === "string") {
+          metadata.byline = parsed.author.name;
+        }
+        if (typeof parsed.description === "string") {
+          metadata.excerpt = parsed.description.trim();
```
gijsk, line 1363: Do we not need to trim the other metadata fields?

Readability.js
gijsk: I think it's fine for it to be an opt-out thing, but then this logic probably needs to be changed, e.g. make the option `disableJSONLD` and update the internal copy accordingly.

Readability.js
danburzo: I don't have data nor an intuition on the performance tradeoff of `Array.from` vs. `call()`-ing array methods on the NodeList, but I agree a separate effort/PR is the way to go here.

Readability.js
danburzo: Oops, good catch! It's only the trailing whitespace that needs specifying.

Readability.js
danburzo: In the testcases, the few articles that broke had trailing spaces in their description. You're right, we should trim spaces on either all or none of the fields (since it's mostly a cosmetic thing).

Readability.js
danburzo: I've changed the flag to `disableJSONLD`.

Readability.js
danburzo: Fixed!

test/test-readability.js
danburzo: Fixed!

Readability.js
danburzo: In my understanding of the threat model here, running `unescapeHTMLEntities` on the plain-text metadata can still generate HTML tags, so being added as `textContent` is the only line of defense here. Since the JSON-LD is not expected to include tags (except for malicious intent), I don't think anything else is strictly needed here, right?

gijsk:
Thanks!

PR #634, opened 2020-11-01
Configuration for isProbablyReaderable thresholds

dgellow:
This pull request makes it possible to change the score and content length limits used to determine if a content is probably readable. That allows callers to have more control over the sensitivity of the function `isProbablyReaderable `.

It would be nice to also move `isVisible` to the `options` object but that would be a breaking change.

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -37,13 +37,21 @@ function isNodeVisible(node) {
 
 /**
  * Decides whether or not the document is reader-able without parsing the whole thing.
- *
- * @return boolean Whether or not we suspect Readability.parse() will suceeed at returning an article object.
+ * @param {Object} options Configuration object.
+ * @param {number} [options.minContentLength=140] The minimum content length used to decide if the document is readerable.
+ * @param {number} [options.minScore=20] The minumum 'score' used to determine if the document is readerable.
+ * @param {Function} [options.visibilityChecker=isNodeVisible] The function used to determine if a node is visible.
+ * @return {boolean} Whether or not we suspect Readability.parse() will suceeed at returning an article object.
  */
-function isProbablyReaderable(doc, isVisible = null, options = { minScore: 20, minContentLength: 140 }) {
-  if (!isVisible) {
-    isVisible = isNodeVisible;
+function isProbablyReaderable(doc, options = {}) {
+  // For backward compatibility reasons 'options' can either be a configuration object or the function used
+  // to determine if a node is visible.
+  if (typeof options === "function") {
+    options = { visibilityChecker: options };
   }
+
+  var defaultOptions = { minScore: 20, minContentLength: 140, visibilityChecker: isNodeVisible };
+  options = Object.assign(options || {}, defaultOptions);
 
   var nodes = doc.querySelectorAll("p, pre");
 
@@ -67,7 +75,7 @@ function isProbablyReaderable(doc, isVisible = null, options = { minScore: 20, m
   // This is a little cheeky, we use the accumulator 'score' to decide what to return from
   // this callback:
   return [].some.call(nodes, function(node) {
-    if (!isVisible(node))
+    if (!options.visibilityChecker(node))
       return false;
 
     var matchString = node.className + " " + node.id;
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -46,7 +46,7 @@ function isNodeVisible(node) {
 function isProbablyReaderable(doc, options = {}) {
   // For backward compatibility reasons 'options' can either be a configuration object or the function used
   // to determine if a node is visible.
-  if (typeof options === "function") {
+  if (typeof options == "function") {
     options = { visibilityChecker: options };
   }
 
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -38,8 +38,8 @@ function isNodeVisible(node) {
 /**
  * Decides whether or not the document is reader-able without parsing the whole thing.
  * @param {Object} options Configuration object.
- * @param {number} [options.minContentLength=140] The minimum content length used to decide if the document is readerable.
- * @param {number} [options.minScore=20] The minumum 'score' used to determine if the document is readerable.
+ * @param {number} [options.minContentLength=140] The minimum node content length used to decide if the document is readerable.
+ * @param {number} [options.minScore=20] The minumum cumulated 'score' used to determine if the document is readerable.
  * @param {Function} [options.visibilityChecker=isNodeVisible] The function used to determine if a node is visible.
  * @return {boolean} Whether or not we suspect Readability.parse() will suceeed at returning an article object.
  */
@@ -51,7 +51,7 @@ function isProbablyReaderable(doc, options = {}) {
   }
 
   var defaultOptions = { minScore: 20, minContentLength: 140, visibilityChecker: isNodeVisible };
-  options = Object.assign(options || {}, defaultOptions);
+  options = Object.assign(defaultOptions, options);
 
   var nodes = doc.querySelectorAll("p, pre");
 
@@ -65,7 +65,7 @@ function isProbablyReaderable(doc, options = {}) {
   var brNodes = doc.querySelectorAll("div > br");
   if (brNodes.length) {
     var set = new Set(nodes);
-    [].forEach.call(brNodes, function(node) {
+    [].forEach.call(brNodes, function (node) {
       set.add(node.parentNode);
     });
     nodes = Array.from(set);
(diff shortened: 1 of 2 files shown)


Readability-readerable.js
```diff
@@ -40,7 +40,7 @@ function isNodeVisible(node) {
  *
  * @return boolean Whether or not we suspect Readability.parse() will suceeed at returning an article object.
  */
-function isProbablyReaderable(doc, isVisible) {
+function isProbablyReaderable(doc, isVisible = null, options = { minScore: 20, minContentLength: 140 }) {
```
gijsk, line 43: The way I understand argument default values (and it's past midnight so I'm probably missing something) this requires that a consumer who provides one also provides another. Is that right? That seems counterintuitive.

Would it make sense to instead use `{}` here but use `Object.assign` or something to ensure the defaults are respected if not passed?

gijsk (changes requested):
Thank you for the PR! I would be fine with taking this as long as there are some tests - would you mind adding some? You could cheat a little and just add another `describe` into `test-readability.js` at https://github.com/mozilla/readability/blob/master/test/test-readability.js#L218 or you could add `test-readerable.js` - I don't really mind which.

I'll also note that there has been some discussion around making the minimum content length language-dependent, so I'm a little wary of adding a language-agnostic entry for future-proof-ness, but I also don't want that to stymie development, so perhaps we should just go for it? We could always release a major version change with breaking changes if we ever grow that support in future...

Readability-readerable.js
dgellow: You're right, that's a very good point. Thanks for the review, I will do that change 👍.
Btw, where is the good place to document these defaults? If we have them in the function signature, that's already somehow documented, but if we move them away, would you prefer to have them in the doc comment, or somewhere else (such as the README file)?

dgellow: Considering the `Object.assign` makes me realize that I can merge both parameters `isVisible` and `options` into one. That way we keep backward compatibility and also avoid to have callers passing a `null` parameter if they just want to change the options.
Something like this:

```js
function isProbablyReaderable(doc, options = {}) {
  // For backward compatibility reasons 'options' can either be a configuration object or the function used
  // to determine if a node is visible.
  if (typeof options === "function") {
    options = {visibilityChecker: options}
  }

 // Here I'm also renaming `isVisible` to `visibilityChecker`. That sounds more like a function, where `isVisible` can be confused for a boolean.
  var defaultOptions = { minScore: 20, minContentLength: 140, visibilityChecker: isNodeVisible }
  options = Object.assign(options || {}, defaultOptions)

  // ...
}
```

I believe that's an improvement over my initial version, though we should be sure to document it correctly (`isVisible` wasn't really documented anywhere anyway, but still).

dgellow:
> I'll also note that there has been some discussion around making the minimum content length language-dependent, so I'm a little wary of adding a language-agnostic entry for future-proof-ness, but I also don't want that to stymie development, so perhaps we should just go for it?

Two thoughts on this:
1. In the future we could still change the default minimums to being language based, while keeping the possibility for callers to overwrite with their own values if they have a reason to do so. For example we could change the default to `-1` or `0` which would mean "I don't really care about the minimums, just do whatever heuristic you want", which can be language based, or something else, up to the library to decide. That's technically a breaking change as we are changing defaults from `score of 20` and `length of 140` to something more dynamic, but that doesn't change the function signature. So that's somehow future-proof.

2. Maybe the language based logic would make more sense as a wrapper around `isProbablyReaderable`, adjusting the `options` object as required. Let's say for French the minimum score and length would be `20` and `140`, while for Mandarin that would be `10` and `70` (I'm just giving random values here for the discussion). So, `check for language defaults => build 'options' object => call isProbablyReaderable`.

test/test-isProbablyReaderable.js
```diff
@@ -6,18 +6,106 @@ var expect = chai.expect;
...
+    var options = { minContentLength: 200, minScore: 0 };
+    expect(isProbablyReaderable(verySmallDoc, options), "very small doc").to.be.false;
+    expect(isProbablyReaderable(smallDoc, options), "small doc").to.be.false;
+    expect(isProbablyReaderable(largeDoc, options), "large doc").to.be.false;
+    expect(isProbablyReaderable(veryLargeDoc, options), "very large doc").to.be.true;
+  });
+
+  it("should declare small and large documents as readerable when lower minScore", function () {
+    var options = { minContentLength: 0, minScore: 4 };
+    expect(isProbablyReaderable(verySmallDoc, options), "very small doc").to.be.false; // score: ~3.3
```
dgellow, line 56: Tests based on score are quite fragile, but I'm not sure what would be a better idea. I would personally prefer if the function `isProbablyReaderable` would return an object with the score (and if the score is 0, a reason why), but that would be a bigger breaking change. Anyway, I think that's good enough for the current situation, but I'm open to better suggestions.

Readability-readerable.js
gijsk: Yep, this suggestion looks great, thanks! Will review the rest in a bit.

test/test-isProbablyReaderable.js
```diff
@@ -6,18 +6,106 @@ var expect = chai.expect;
...
+      }
+    };
+    expect(isProbablyReaderable(veryLargeDoc, options)).to.be.true;
+    expect(called).to.be.true;
+  });
+
+  it("should use node visibility checker provided as parameter - not visible", function () {
+    var called = false;
+    var visibilityChecker = () => {
+      called = true; return false;
```
gijsk, line 97: Nit: split this into two lines (same below) ?

gijsk (changes requested):
This looks great, thank you - esp. for the detailed documentation.

I pretty much only have styling nitpicks - I can edit those in myself or you can do it, not sure what is github "etiquette" around that at this point. I like your idea for exposing scoring details, and I made a suggestion as to how we might do that.

test/test-isProbablyReaderable.js
dgellow: Good idea. That would be quite weird to have an out parameter in a JS context, but that's a possible workaround. I will maybe look into this later today, I'm not sure yet if that's worth doing.

Readability-readerable.js
```diff
@@ -66,13 +74,14 @@ function isProbablyReaderable(doc, isVisible) {
...
-    if (!isVisible(node))
+  return [].some.call(nodes, function (node) {
+    if (!options.visibilityChecker(node)) {
       return false;
+    }
 
     var matchString = node.className + " " + node.id;
     if (REGEXPS.unlikelyCandidates.test(matchString) &&
-        !REGEXPS.okMaybeItsACandidate.test(matchString)) {
+      !REGEXPS.okMaybeItsACandidate.test(matchString)) {
```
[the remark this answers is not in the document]
dgellow: I can, but the change has been done by `npm run lint -- --fix`, which is using `eslint`.

Readability-readerable.js
```diff
@@ -37,14 +37,22 @@ function isNodeVisible(node) {
...
+ * @param {Function} [options.visibilityChecker=isNodeVisible] The function used to determine if a node is visible.
+ * @return {boolean} Whether or not we suspect Readability.parse() will suceeed at returning an article object.
  */
-function isProbablyReaderable(doc, isVisible) {
-  if (!isVisible) {
-    isVisible = isNodeVisible;
+function isProbablyReaderable(doc, options = {}) {
+  // For backward compatibility reasons 'options' can either be a configuration object or the function used
+  // to determine if a node is visible.
+  if (typeof options === "function") {
```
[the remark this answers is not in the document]
dgellow:
```suggestion
  if (typeof options == "function") {
```

test/test-isProbablyReaderable.js
dgellow:
```suggestion
      called = true;
      return false;
```

test/test-isProbablyReaderable.js
```diff
@@ -6,18 +6,106 @@ var expect = chai.expect;
...
+    };
+    expect(isProbablyReaderable(veryLargeDoc, visibilityChecker)).to.be.false;
+    expect(called).to.be.true;
+  });
+
+  it("should use node visibility checker provided as parameter - visible", function () {
+    var called = false;
+    var visibilityChecker = () => {
+      called = true; return true;
```
dgellow, line 106:
```suggestion
      called = true;
      return true;
```

Readability-readerable.js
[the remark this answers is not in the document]
dgellow:
```suggestion
        !REGEXPS.okMaybeItsACandidate.test(matchString)) {
```

dgellow:
Feel free to squash if you don't like to have 5 commits for this feature.

Readability-readerable.js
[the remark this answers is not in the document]
gijsk: Huh... that's confusing. I didn't think this repo ran `prettier` or any other formatting linter, so I'm not sure why it'd do that. Is it possible it's inheriting some other / global lint configuration?

gijsk:
Looks great, thank you!

dgellow:
Thanks a lot for your time @gijsk ❤️

PR #636, opened 2020-11-03
Run linter and tests on pull request

dgellow:
Simple GitHub actions to run the linter and tests in pull requests.
Feel free to close it if you don't want to use actions for some reasons.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,24 @@
+name: Run linter and tests
+
+on: [pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12'
+      - run: npm install
+      - run: npm run lint
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12'
+      - run: npm install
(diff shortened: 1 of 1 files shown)


gijsk:
Hm, we used to use travis for this, but I no longer see anything for that show up on github? Here's the build for your other PR, for instance: https://travis-ci.org/github/mozilla/readability - so it happened, but I don't see it in the github UI.

I'm afraid I know very little about github actions vs. travis - I didn't set up the travis system, and I haven't yet had a reason to investigate github actions (prior to this PR, that is!). Having just looked, it looks like they are globally disabled on mozilla github repos. I'll check what's up with that and/or why the travis checks aren't obviously visible on github anymore...

gijsk:
OK, so... it seems nobody has figured out how to deal with github actions and avoid Mozilla getting a giant bill if people abuse the actions, or something, which is why they're disabled (see https://bugzilla.mozilla.org/show_bug.cgi?id=1661286 ). I'm a bit out of my depth in terms of championing that (and swamped with other work), so I don't think we can take this PR right now. Sorry! :-(

It seems the easy/right thing on Mozilla's end is that we switch from travis to use taskcluster, for which the cost story has been figured out already. I requested integration access for the repo and that has been granted, so AIUI now all it takes is using https://community-tc.services.mozilla.com/quickstart to generate a yml file and replace the travis one with that? I would love a PR for that but I would totally understand if you don't have time/interest in pursuing that. Do let me know if you *want* to do it but get stuck anywhere...

dgellow:
No worries, I will create another PR :)

Issue #629, opened 2020-10-11
Question on `nextElement` implementation


hipstermojo:
This function is used to get the next possible element and may return null however the code:
```js
this.REGEXPS.whitespace.test(next.textContent)
```
allows for the possibility of text nodes also being returned. Is this the intended purpose of the function; to return both element and text nodes?

gijsk:
Yes, it looks like it can return both. The code at https://github.com/mozilla/readability/blame/d5eea06a0095b3138dbd1f6233f656d690200509/Readability.js#L580 makes this look intentional: if there's markup like:

```html
<p>
Some text
<br><br>
<br>
Some more text
</p>
```

it intends to loop over the `<br>` tags, and the whitespace inbetween, but then stop short of the "Some more text" text node (as it won't match the whitespace regex).

It'd probably be a good idea to rename this so it doesn't imply it only returns elements.

ankushduacodes:
Hi @gijsk, @hipstermojo, I would love to work on this issue, but I am very new to open source and this will only be my 3rd or 4th contribution, Is this a beginner-friendly issue, if yes that I would love to take on this issue and if not can you please recommend a good beginner-friendly issue that I can work on.
Thank you.

gijsk:
Hi @ankushduacodes , sure, you could pick this up. Fixing it would involve renaming the `nextElement` method in Readability.js to `nextNode`.

ankushduacodes:
@gijsk Could you please give a starting point from which file should start or does the order not matter?
and I will be just renaming nextElement method to nextNode right or is there something more that I should be aware of?

gijsk:
I think the method is only used from within the `Readability.js` file. And I don't think there's anything else to be aware of - once you're done, it'd be good to make sure tests still pass, you can check with `npm test`.

ankushduacodes:
For sure @gijsk,
Thank you for the tip

gijsk:
This was fixed in #637 . \o/

PR #637, opened 2020-11-04
Rename _nextElement to solve Issue#629

ankushduacodes:
Solving #629


.gitignore
```diff
@@ -8,3 +8,6 @@ node_modules
 .sw?
 *.jar
 *.xpi
+.idea/
+.vscode
```
danburzo, line 12: The files produced by your choice of editor / IDE should be excluded for all Git repositories on your machine so. [Here's how to do that](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

.gitignore
```diff
@@ -8,3 +8,6 @@ node_modules
...
+package-lock.json
```
danburzo, line 13: This is a file that actually _should_ be checked into the repository, so that everyone working on the project uses the same versions of the development dependencies.

Readability.js
```diff
@@ -552,7 +552,7 @@ Readability.prototype = {
    * whitespace in between. If the given node is an element, the same node is
    * returned.
    */
-  _nextElement: function (node) {
```
danburzo, line 555: The comment above the method definition should also reflect the change.

danburzo (changes requested):
This looks good to me, thanks for the contribution! I've added a few thoughts on minor, mostly tangential, issues that should be addressed, I think, before merging.

.gitignore
dgellow: Was there a reason to not add it to git in the first place?

.gitignore
ankushduacodes: Okay, Thank you for making me aware of that. I actually did not know if that was possible😅. I will remove them from .gitignore

.gitignore
ankushduacodes: It was not present in the repo so I thought it was intentional. Do you want me to commit the `package-lock.json` file in this pull request?

danburzo: I think @gijsk should ultimately make the call on that.

Readability.js
ankushduacodes: I have made the necessary change to the the docs. Do let me know if it looks fine or do I need to make another change.

.gitignore
gijsk: I think the only reason it's not committed right now is that there wasn't one when we first made the library node-compatible (many years ago) and we've never remedied that. I think it's a good idea to commit it, but it should be a separate commit and PR. It probably shouldn't be added to gitignore in this PR.

ankushduacodes: @gijsk I have undone the .gitignore changes, Just to make sure, Do you want me to make another issue and create a pull request adding `package-lock.json`?

gijsk:
Thanks for the PR! This all looked great apart from the whitespace-only change to gitignore, which I reverted. I'll merge this now. :-)

.gitignore
gijsk: @ankushduacodes a new PR to add `package-lock.json` would be great, yes, thank you!

PR #638, opened 2020-11-04
Run linter and tests on pull request

dgellow:
Configuration file created using https://community-tc.services.mozilla.com/quickstart.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,0 +1,26 @@
+version: 1
+policy:
+  pullRequests: public
+tasks:
+  $match:
+    '(tasks_for == "github-pull-request" && event["action"] in [pull_request.opened,pull_request.reopened,pull_request.synchronize])':
+      taskId:
+        $eval: as_slugid("pr_task")
+      provisionerId: proj-getting-started
+      workerType: tutorial
+      payload:
+        maxRunTime: 3600
+        image: node
+        command:
+          - /bin/bash
+          - '--login'
+          - '-c'
+          - >-
+            git clone {{event.head.repo.url}} repo && cd repo && git config
+            advice.detachedHead false && git checkout {{event.head.sha}} && npm
+            install . && npm test
+      metadata:
+        name: Run tests and linter
+        description: ''
+        owner: '${event.sender.login}@users.noreply.github.com'
+        source: '${event.repository.url}'
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -18,7 +18,7 @@ tasks:
           - >-
             git clone {{event.head.repo.url}} repo && cd repo && git config
             advice.detachedHead false && git checkout {{event.head.sha}} && npm
-            install . && npm test
+            install . && npm run lint && npm test
       metadata:
         name: Run tests and linter
         description: ''
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -3,7 +3,7 @@ policy:
   pullRequests: public
 tasks:
   $match:
-    '(tasks_for == "github-pull-request" && event["action"] in [pull_request.opened,pull_request.reopened,pull_request.synchronize])':
+    '(tasks_for == "github-pull-request" && event["action"] in ["opened","reopened","synchronize"])':
       taskId:
         $eval: as_slugid("pr_task")
       provisionerId: proj-getting-started
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -6,8 +6,8 @@ tasks:
     '(tasks_for == "github-pull-request" && event["action"] in ["opened","reopened","synchronize"])':
       taskId:
         $eval: as_slugid("pr_task")
-      provisionerId: proj-getting-started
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -0,0 +1,26 @@
+version: 1
+policy:
+  pullRequests: public
+tasks:
+  $match:
+    '(tasks_for == "github-pull-request" && event["action"] in [pull_request.opened,pull_request.reopened,pull_request.synchronize])':
```
dgellow: That's what the "getting started with github" page generated, but that doesn't seem to be correct based on https://community-tc.services.mozilla.com/docs/reference/integrations/github/taskcluster-yml-v1. `pull_request` shouldn't be used here, instead it should be `in ["opened", "reopened", "synchronize"]`. It seems that the generator has some issues.

dgellow:
@community-tc-integration, so, is it good now? 🤷

.taskcluster.yml
```diff
@@ -0,0 +1,26 @@
+version: 1
+policy:
+  pullRequests: public
+tasks:
+  $match:
+    '(tasks_for == "github-pull-request" && event["action"] in ["opened","reopened","synchronize"])':
+      taskId:
+        $eval: as_slugid("pr_task")
+      provisionerId: proj-getting-started
+      workerType: tutorial
```
gijsk, line 10: Hm, I imagine we need to change this too, but I'm not sure what to. I'll ping someone who should know.

.taskcluster.yml
```diff
@@ -0,0 +1,26 @@
...
+        maxRunTime: 3600
+        image: node
+        command:
+          - /bin/bash
+          - '--login'
+          - '-c'
+          - >-
+            git clone {{event.head.repo.url}} repo && cd repo && git config
+            advice.detachedHead false && git checkout {{event.head.sha}} && npm
+            install . && npm test
```
gijsk, line 21:
```suggestion
            install . && npm lint && npm test
```

maybe?

dgellow: Oh, that's funny, I thought I did that change, but I don't see it here. I guess I forgot to update my commit before pushing.

dgellow: That's the correct syntax though:

```suggestion
            install . && npm run lint && npm test
```

.taskcluster.yml
```diff
@@ -0,0 +1,26 @@
+version: 1
+policy:
+  pullRequests: public
+tasks:
+  $match:
+    '(tasks_for == "github-pull-request" && event["action"] in ["opened","reopened","synchronize"])':
+      taskId:
+        $eval: as_slugid("pr_task")
+      provisionerId: proj-getting-started
```
dgellow, line 9: Should this be changed?

gijsk:
@djmitche , would you mind taking a look? I tried to look for documentation on what `provisionerId` and `workerType` should be set to, but the search on https://docs.taskcluster.net/ seems nonfunctional (ie if I put in `provisionerId`, it comes up with "no options" and pressing enter/return does not appear to do anything). I used web search for "taskcluster provisionerId" instead and found https://docs.taskcluster.net/docs/manual/task-execution/provisioning but I can't work out if there's a limited set of known things we should use or if this is basically arbitrary labeling or...

Any guidance would be appreciated!

(I'd add you as a reviewer but github isn't letting me...)

.taskcluster.yml
```diff
@@ -0,0 +1,26 @@
...
+      workerType: tutorial
+      payload:
+        maxRunTime: 3600
```
gijsk, line 12: We could probably adjust this down a bit, but it's hard to know how much until it initially works.

gijsk:
Thanks very much for pursuing this! I'm sure we can get there... but I feel like the blind leading the blind atm. 😅

.taskcluster.yml
djmitche, line 10: You'll want
```suggestion
      provisionerId: proj-misc
      workerType: ci
```

The config for this deployment is in https://github.com/mozilla/community-tc-config/, which [allows all mozilla/* repos to use this worker pool](https://github.com/mozilla/community-tc-config/blob/24423f6b33f7988c6c06db5b861d3facedfbabd4/config/projects/misc.yml#L59-L63).  We happily take pull requests there to modify the config, but for the moment that shouldn't be necessary.

I know it's difficult to discover these things, and partly that's because it's specific to the deployment (this worker pool wouldn't work in the Firefox-CI Taskcluster deployment) and partly it's specific to the repo (e.g., `gijsk/readability` does not have permission to use this worker pool).

djmitche (commented):
This looks good otherwise!

dgellow:
@djmitche Is there a way to test if the config file is correct?

djmitche:
Easiest is just to land it!  It's not running on this PR because there's no `policy` in place on the main branch right now.

gijsk:
🤞

dgellow:
Thanks to both of you for the review and quick feedbacks!

Issue #625, opened 2020-10-01
Question on `hasChildBlockElement`


hipstermojo:
I noticed that the function checks whether any child node has a tag name matching those in the `DIV_TO_P_ELEMS` array. One of these elements is the `<a>` tag which is actually an inline element. Was this intentional?

gijsk:
Well, the code has been like that since 6 years ago when it was imported to github, so probably since the original arc90 version of this. So it's hard to say what the "intent" of other people, all those years ago, was - the comments don't really tell us. :-(

@hipstermojo Out of interest, how did you run into this? It's easier to see we should change it if it's breaking something in particular. :-)

gijsk:
(It looks like dropping "A" from the list has 0 impact across all of the testcases.)

hipstermojo:
I ran into this "realization" while porting this library to Rust. I haven't finished porting it so I don't how how differently the code will run once I write it slightly differently from what is in the "canonical" implementation in this repository.

I opted to be more formal and just compare elements with the full list of block elements since I saw that even `<img>` and `<select>` are not block elements.

hipstermojo:
I've tested it out and it does break quite a lot of tests lol. I will try investigating why. The time taken to run tests is much longer, probably because searching for a tag name in the array runs at n time and using a Set would be more efficient. In any case, would it mean the test suite should be changed?

gijsk:
I'm a little confused - what specifically breaks tests? Swapping the list out for a list of actual block elements per the spec? Because just removing `"A"` didn't affect test output...

As for the time taken to run tests being longer, that could also be a result of different outcomes meaning we spend more time in other parts of the code...

Using a `Set` is a good idea though I don't know if those are guaranteed to be available in the version of node the library requires...

hipstermojo:
I meant swapping it for a list of actual block elements as per the spec.

> Using a Set is a good idea though I don't know if those are guaranteed to be available in the version of node the library requires...
I feared as much :(

hipstermojo:
Would it be okay to open other issues on some of the things that can also be safely removed from the codebase for easier readability? (excuse the pun)

gijsk:
Sure, I'd be more than happy with some issues and/or PRs around things that, on closer inspection, look useless / broken to you. Thank you for sharing that knowledge!

hipstermojo:
Alright. Sounds great :smile: . I'll just close this then and hopefully get a PR out when I'm done porting as much as I can

PR #645, opened 2020-11-07
Code readability fixes

hipstermojo:
This PR makes a few fixes to the _readability_ of the code in the repo. This includes [removing redundancies](https://github.com/mozilla/readability/commit/d138978bf1691433dc79de1bafc068940cf3fac7) and fixing a JSDoc comment.

The PR also addresses the issue discussed in #625

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -230,8 +230,7 @@ Readability.prototype = {
     if (this._docJSDOMParser && nodeList._isLiveNodeList) {
       throw new Error("Do not pass live node lists to _replaceNodeTags");
     }
-    for (var i = 0; i < nodeList.length; i++) {
-      var node = nodeList[i];
+    for (const node of nodeList) {
       this._setNodeTag(node, newTagName);
     }
   },
@@ -1446,13 +1445,11 @@ Readability.prototype = {
       if (elementProperty) {
         matches = elementProperty.match(propertyPattern);
         if (matches) {
-          for (var i = 0; i < matches.length; i++) {
-            // Convert to lowercase, and remove any whitespace
-            // so we can match below.
-            name = matches[i].toLowerCase().replace(/\s/g, "");
-            // multiple authors
-            values[name] = content.trim();
-          }
+          // Convert to lowercase, and remove any whitespace
+          // so we can match below.
+          name = matches[0].toLowerCase().replace(/\s/g, "");
+          // multiple authors
+          values[name] = content.trim();
         }
       }
       if (!matches && elementName && namePattern.test(elementName)) {
@@ -2036,7 +2033,9 @@ Readability.prototype = {
 
       this.log("Cleaning Conditionally", node);
 
-      if (weight < 0) {
+      var contentScore = 0;
+
+      if (weight + contentScore < 0) {
         return true;
       }
 
--- a/Readability.js
+++ b/Readability.js
@@ -143,7 +143,7 @@ Readability.prototype = {
 
   UNLIKELY_ROLES: [ "menu", "menubar", "complementary", "navigation", "alert", "alertdialog", "dialog" ],
 
-  DIV_TO_P_ELEMS: [ "A", "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL", "SELECT" ],
+  DIV_TO_P_ELEMS: [ "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL" ],
 
   ALTER_TO_DIV_EXCEPTIONS: ["DIV", "ARTICLE", "SECTION", "P"],
 
--- a/Readability.js
+++ b/Readability.js
@@ -230,7 +230,7 @@ Readability.prototype = {
     if (this._docJSDOMParser && nodeList._isLiveNodeList) {
       throw new Error("Do not pass live node lists to _replaceNodeTags");
     }
-    for (var i = nodeList.length - 1; i >= 0; i--) {
+    for (var i = 0; i < nodeList.length; i++) {
       var node = nodeList[i];
       this._setNodeTag(node, newTagName);
     }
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -230,7 +230,7 @@ Readability.prototype = {
     if (this._docJSDOMParser && nodeList._isLiveNodeList) {
       throw new Error("Do not pass live node lists to _replaceNodeTags");
     }
-    for (var i = nodeList.length - 1; i >= 0; i--) {
+    for (var i = 0; i < nodeList.length; i++) {
```
dgellow, line 233: The index doesn't seem to be used for anything else than the iteration itself, you should be able to use `for (const i of nodeList) {/* ... */}` here.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of.

dgellow, line 233: The `i` index doesn't seem to be used for something other than the iteration itself here. You should be able to use the form `for (const node of nodeList) { /* ... */ }` instead.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of.

Readability.js
```diff
@@ -2033,11 +2033,10 @@ Readability.prototype = {
       }
 
       var weight = this._getClassWeight(node);
-      var contentScore = 0;
 
       this.log("Cleaning Conditionally", node);
 
-      if (weight + contentScore < 0) {
```
dgellow, line 2040: I see that `contentScore` is never set to anything other than `0`, but was it supposed to be set to something? Or is the `weight` the same as the "content score"?

danburzo: I would assume this comment `// TODO: Consider taking into account original contentScore here.` means `contentScore` could/should potentially be based on `node.readability.contentScore`.

Readability.js
hipstermojo: A lot of code patterns here could be updated to ES6 syntax but it will mean needing a polyfill for older Node versions. This is why I have not changed to `for...of`

Readability.js
hipstermojo: Using the node's contentScore breaks a lot of tests so there's probably something that was to be done there. Should I revert this change?

Readability.js
gijsk: The reverse iteration here (pre-patch) was deliberate - if the nodelist is live, iterating forward means you miss elements, because the ones you remove from the DOM get removed from the node list and so the indices will be wrong.

By now, we throw here if we operate on a live list, and there's only 1 callsite which either passes the result of `querySelectorAll` (which is a static list) or the concatenation of arrays (also not live), so I guess it's OK... but worth bearing in mind if you find more of these loops.

Other than that, AFAICT the package requires node.js >= 10.0.0, which should definitely support for...of loops, and wouldn't need a polyfill? If we change these lines we might as well update to modern JS in one go... :-)

Readability.js
```diff
@@ -1446,7 +1446,7 @@ Readability.prototype = {
       if (elementProperty) {
         matches = elementProperty.match(propertyPattern);
         if (matches) {
-          for (var i = matches.length - 1; i >= 0; i--) {
+          for (var i = 0; i < matches.length; i++) {
```
gijsk, line 1449: This alters the outcome, not just readability, right? `propertyPattern`'s `match` will have generated an array, and the loop overwrites `name` and `values[name]` with the last item in that array, instead of the first. Is that an intentional change? Should we just get the first item out of the list and then stop iterating?

Readability.js
gijsk: I'd leave it like this for now, yes please. It might be worth a separate investigation as to why tests break and if we should either remove both this code and the TODO comment, and add an explanation of why we can't/shouldn't use the `contentScore`, or fix it so we *do* use it.

Readability.js
```diff
@@ -143,7 +143,7 @@ Readability.prototype = {
 
   UNLIKELY_ROLES: [ "menu", "menubar", "complementary", "navigation", "alert", "alertdialog", "dialog" ],
 
-  DIV_TO_P_ELEMS: [ "A", "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL", "SELECT" ],
+  DIV_TO_P_ELEMS: [ "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL" ],
```
gijsk, line 146: It looks like the addition of `select` was from https://bugzilla.mozilla.org/show_bug.cgi?id=785343 - I bet all the pages cited there have changed and this was before we added unit-y tests for this kind of thing (in fact, it was before I was involved with this code!) so hard to know what to do about it now. Do we perhaps scrap `<option>` elements elsewhere, to avoid counting them as 'text' for the scoring algorithm?

gijsk (changes requested):
Thanks for picking this up! I left some feedback here and there.

Readability.js
hipstermojo: I guess I can just keep it as it was before, looping in reverse, just in case it causes a regression later.

Readability.js
hipstermojo: >  Do we perhaps scrap <option> elements elsewhere, to avoid counting them as 'text' for the scoring algorithm?
I will add this in the next commit. I'd also like to change `DIV_TO_P_ELEMS` to a Set since it is only used to test for membership and I think a Set is better for this. Is that okay?

gijsk: Yep, using a `Set` sounds great, thank you!

Readability.js
hipstermojo: Yes, it actually changes the outcome. I seemed to have mindlessly changed any loop that was in reverse to loop in order. Getting the first element, which is the whole match is actually better.

hipstermojo:
The new changes are:
- `DIV_TO_P_ELEMS` is changed from an array to a Set since it only ever checks for membership
- The `contentScore` variable is back
- Use `for...of` to traverse the NodeList
- Replaced looping the `matches` with using the first element in the matches which is the whole match

gijsk (approved):
Makes sense, thank you!

PR #647, opened 2020-11-17
Skip removing nodes in <code>

jakubriedl:
Basically all syntax highlighters wrap comments in code blocks in `<span class="comment">// some comment</span>`.
And because word `comment` is correctly in `REGEXPS.unlikelyCandidates` and `REGEXPS.negative` they get removed. But those should be kept as it is valid content.

This PR fixes the issue by disabling removing of nodes in `<code>`.

test url: https://v8.dev/blog/emscripten-standalone-wasm


Readability.js
```diff
@@ -2029,6 +2030,10 @@ Readability.prototype = {
         return false;
       }
 
+      if (this._hasAncestorTag(node, "code", -2)) {
```
gijsk, line 2033: Why -2 ? In the other change you didn't pass a `maxDepth` argument at all. It also looks like passing `0` or `-2` is functionally equivalent, so this is a bit confusing. Can you clarify, or maybe change to not pass this if it doesn't make a difference? Thanks!

gijsk (changes requested):
This looks good in principle, just one nit around the use of `_hasAncestorTag`. Thanks!

(And apologies for the slowness, things are rather... hectic... both at work and in my own life.)

jakubriedl:
nw with response time. Thank you for doing open-source projects.

I've removed the `maxDepth` param. There isn't actually any need to limit that. Was more copy/paste thing from line above.

Issue #615, opened 2020-08-20
Headings and dds missing from Google SRE book website


williamsmj:
e.g. https://landing.google.com/sre/sre-book/chapters/monitoring-distributed-systems/

In reader mode the section headings and dd/dt elements are missing. Tables and lists are shown fine.

PalmerAL:
For the headers, it looks like readability [removes all h1 tags](https://github.com/mozilla/readability/blob/master/Readability.js#L678). That doesn't seem correct to me; I think the correct solution would be to only remove the element selected by `_getArticleTitle`.

gijsk:
> For the headers, it looks like readability [removes all h1 tags](https://github.com/mozilla/readability/blob/master/Readability.js#L678). That doesn't seem correct to me; I think the correct solution would be to only remove the element selected by `_getArticleTitle`.

Yeah, the headers part is #464 .

Issue #594, opened 2020-05-17
h2 headings not appearing in reader mode for nymag.com


yoasif:
1. Visit https://nymag.com/intelligencer/2020/05/polarization-republicans-democracy-ezra-klein-book-review.html?utm_source=undefined&utm_medium=undefined&utm_campaign=feed-part
2. Enable reader mode in Firefox

What happens:

The headings: 

*  Humanity had a traumatic childhood. 
*  The golden era of bipartisanship was (not coincidentally) an age of authoritarian white supremacy. 
*  The left is all right. 
*  Why the right’s gone wrong. 

etc. don't appear in reader mode.

PR #650, opened 2020-11-22
Fix headers being removed

### LINKED ISSUE #464: H1 tags within the article's body are removed
The target page I am trying to use Readability on is: https://topicseed.com/content-depth-for-seo/

The article's body is correctly identified, yet, all of these H1 tags are not appearing in the parsed content:
- Assess How Deep Is Your Content
- Rewrite With Content Depth In Mind
- Yes, Content Depth and Breadth Overlap
- Depth of Content = Quality + Frequency

The paragraphs before and after these are correctly present in the parsed content. No idea what could be causing this as it is nothing extraordinary?

Config:
- nodejs 9 + jsdom
- package.json imports `"readability": "github:mozilla/readability"`

jakubriedl:
Fixes #464 (✅) #594 (possibly, cannot be tested as there is some issue with `JSDOMParser`) #615 (headings part ✅)

As described in mentioned issues headings are removed from content where they shouldn't. This PR fixes that by modifying the logic that removes them and adds new checks.

There is one big question though. **Should we remove headings in case, where it is first element, but it is completely different to title.** eg. `this._articleTitle = "Dublin Core property title"` (from sample `003-metadata-preferred`)
```html
    <article>
        <h2>Test document title</h2>
        <p> ... </p>
        <p> ... </p>
    </article>
```
I personally think it **shouldn't** be removed in that case as it isn't duplicit. But this PR as it is now removes it to reduce number of test cases that are changed (19 cases). If you want to keep this case I will just remove the [condition](https://github.com/lr-nr/readability/blob/4e6ae75ecf2040afb0840202e0cb9567defc00b8/Readability.js#L2127-L2134) 

Effects of this PR on tests:
- ⚠️ `buzzfeed-1` it removes title that should stay because it is first element. It comes to the question above if we want to keep this or not.
- ✅ `citylab-1`
- ✅ `dropbox-blog` (added as new test for this PR https://dropbox.tech/infrastructure/asynchronous-task-scheduling-at-dropbox)
- ✅ `ehow-1`
- ✅ `embeded-videos` (removes h2 on top of article that is similar to title)
- ✅ `keep-tabular-data` (removes h2 on top of article that is similar to title)
- ✅ `lazy-image-1`
- ✅ `medium-1` (removes h2 on top of article that is similar to title)
- ✅ `tmz-1` (removes h2 on top of article that is similar to title)
- ✅ 🤔 `yahoo-2` correctly keeps heading in the content, but in the original source the text of that is duplicated in following `<p>` so it looks less correct. But it is what the underlying html tells us to show.


test/test-pages/buzzfeed-1/expected.html
```diff
@@ -1,9 +1,5 @@
 <div id="readability-page-1" class="page">
     <div id="buzz_sub_buzz">
-        <div id="superlist_3758406_5547137" rel:buzz_num="1">
-            <h2>The mother of a woman who took suspected diet pills bought online has described how her daughter was “literally burning up from within” moments before her death.</h2>
-            <p> <span>West Merica Police</span></p>
-        </div>
```
gijsk, line 6: Hm, shouldn't this be staying here?

Readability.js
```diff
@@ -831,6 +815,23 @@ Readability.prototype = {
...
+    var tokensA = textA.toLowerCase().split(this.REGEXPS.tokenize).filter(Boolean);
+    var tokensB = textB.toLowerCase().split(this.REGEXPS.tokenize).filter(Boolean);
+    if (tokensA.length === 0 || tokensB.length === 0) {
+      return 0;
+    }
+    var uniqTokensA = tokensA.filter(token => !tokensB.includes(token));
+    var uniqTokensB = tokensB.filter(token => !tokensA.includes(token));
+    var distanceA = uniqTokensA.join(" ").length / tokensA.join(" ").length;
+    var distanceB = uniqTokensB.join(" ").length / tokensB.join(" ").length;
+    return 1 - Math.min(distanceA, distanceB);
```
gijsk, line 832: So, as I understand it, if the detected article title is "Good food", and the header is "Good food: everything you need to know to cook like a pro", or whatever, this algorithm returns 1, because all of "Good food" occurs in the longer title, so one of `distanceA` and `distanceB` is 0, and so we return 1. And that would cause us to drop the h1. Correct? If so, I don't think that's quite optimal - can we avoid doing this if the detected title is the longer version of the same?

Readability.js
```diff
@@ -831,6 +815,23 @@ Readability.prototype = {
...
   },
 
+  // text similarity comparison
+  // 1 = same text, 0 = completely different text
+  // works the way that it splits both texts into words and then finds words that are unique in one or the other
+  // the result is given by the lower length of unique parts
+  _textSimilarity: function(textA, textB) {
+    var tokensA = textA.toLowerCase().split(this.REGEXPS.tokenize).filter(Boolean);
+    var tokensB = textB.toLowerCase().split(this.REGEXPS.tokenize).filter(Boolean);
+    if (tokensA.length === 0 || tokensB.length === 0) {
```
gijsk, line 825: Nit: `!tokensA.length || !tokensB.length`

Readability.js
```diff
@@ -2104,14 +2117,28 @@ Readability.prototype = {
...
-      return this._getClassWeight(header) < 0;
+    this._removeNodes(this._getAllNodesWithTag(e, ["h1", "h2"]), function (node) {
+      // remove heading if it is first element
+      let prev = node.previousSibling || (node.parentNode || {}).previousSibling;
+      while (prev && this._isWhitespace(prev)) {
+        prev = prev.previousSibling || (prev.parentNode || {}).previousSibling;
+      }
+      if (!prev) {
+        return true;
+      }
```
gijsk, line 2134: Hm, did we do this before? I'm worried about this especially because some high-profile sites (CNN, if I remember correctly) like to use `h2` for their introductory paragraphs, and obviously we would be better off not cutting those entirely...

gijsk (changes requested):
Thanks for this! It looks pretty interesting, but I have a few questions/concerns. And yes:

> I personally think it **shouldn't** be removed in that case as it isn't duplicate. But this PR as it is now removes it to reduce number of test cases that are changed (19 cases). If you want to keep this case I will just remove the condition

Yeah, I think we should keep it... if we improve the testcases we shouldn't be worried about that. :-)

test/test-pages/buzzfeed-1/expected.html
jakubriedl: yes, that will stay in moment when I keep first headers.

Readability.js
jakubriedl: yes that is exactly how it works now. I was inspired by the original H2 removal. I also tried using levensthein distance and that struggles in cases where it has the same words just in different order.

What if we do here `1 - Math.max(distanceA, distanceB)`. In that case the longer text will define the final number and in case as you described the text similarity will be low (~0.2) and so it will be kept.

Readability.js
jakubriedl: it wasn't specified like this but practically we did. (because 19 test change when we don't do it). I will remove this block as you mentioned above that you want to keep those headers.

jakubriedl:
@gijsk it is updated and keeps the leading headings if they are different, also considers the heading different if it is superset of title words. I also did one change and that is to remove max one heading. Sometimes there is match lower in the text which should be kept.

All changed tests look correct to me.

gijsk:
(Just a note to say I've not forgotten this but I've been rather pre-occupied having just moved house with my family. I'll get back to this as soon as I can but it's likely to be next week - mostly because there's no fixed internet here yet, and doing reviews over a tethered mobile connection is... not easy. Apologies!)

test/test-pages/ehow-1/expected.html
```diff
@@ -7,6 +7,11 @@
             <figure> <img src="http://img-aws.ehowcdn.com/640/cme/photography.prod.demandstudios.com/16149374-814f-40bc-baf3-ca20f149f0ba.jpg" alt="Glass cloche terrariums" title="Glass cloche terrariums" data-credit="Lucy Akins " longdesc="http://s3.amazonaws.com/photography.prod.demandstudios.com/16149374-814f-40bc-baf3-ca20f149f0ba.jpg" /> </figure>
             <figcaption class="caption"> Glass cloche terrariums (Lucy Akins) </figcaption>
         </div>
+        <div id="relatedContentUpper" data-module="rcp_top">
+            <header>
+                <h3>Other People Are Reading</h3>
+            </header>
+        </div>
```
gijsk, line 14: Hm, I guess we should file a separate issue to filter these out...

gijsk (approved):
This looks great! Apologies again for my slowness... but your patches are very much appreciated!

I think I should probably cut a 0.4 release soon, unless you had other things you were hoping to fix before then?

jakubriedl:
I don't have any fix in plan soon as we got our project ([lrnr.dev](lrnr.dev)) to MVP phase and focus goes on different parts now. But we plan to come back to it later do more fixes and open-source our analyzer which depends on Readability.

dgellow:
Merging this PR broke tests on master branch.

image

gijsk:
@dgellow, thanks for the heads-up. I dunno why that wasn't flagged up by taskcluster on the PR... will try and address this "soon" but I'm having to keep a few too many plates spinning at the moment, and so I can't promise when I'll be able to do so.

dgellow:
Sorry, I didn't want to add even more pressure to you @gijsk. I do not know you outside of our short interactions on this repository bug tracker, but I wish you the best in your current situation ❤️. Managing your stress level should have priority over failing tests.

All the best!

PR #663, opened 2020-12-22
Fix #659 by splitting out removing headers because they're similar to titles and removing those with low scores

### LINKED ISSUE #659: Automated tests are failing since #650 (removing an `h2` that wasn't previously removed)
As noted in https://github.com/mozilla/readability/pull/650#issuecomment-744545189, the `v8-blog` test is failing:

```
      -h2#using-standalone-mode-in-emscripten
      +p
```

Re-running `generate-testcase.js` produces this diff:

```diff
diff --git a/test/test-pages/v8-blog/expected.html b/test/test-pages/v8-blog/expected.html
index f95d96e..85664fb 100644
--- a/test/test-pages/v8-blog/expected.html
+++ b/test/test-pages/v8-blog/expected.html
@@ -1,8 +1,6 @@
 <div id="readability-page-1" class="page">
     <div itemprop="articleBody">
         <p> Emscripten has always focused first and foremost on compiling to the Web and other JavaScript environments like Node.js. But as WebAssembly starts to be used <em>without</em> JavaScript, new use cases are appearing, and so we've been working on support for emitting <a href="https://github.com/emscripten-core/emscripten/wiki/WebAssembly-Standalone"><strong>standalone Wasm</strong></a> files from Emscripten, that do not depend on the Emscripten JS runtime! This post explains why that's interesting. </p>
-        <h2 id="using-standalone-mode-in-emscripten"> Using standalone mode in Emscripten <a href="#using-standalone-mode-in-emscripten">#</a>
-        </h2>
         <p> First, let's see what you can do with this new feature! Similar to <a href="https://hacks.mozilla.org/2018/01/shrinking-webassembly-and-javascript-code-sizes-in-emscripten/">this post</a> let's start with a "hello world" type program that exports a single function that adds two numbers: </p>
         <pre><code><span>// add.c</span><br/><span><span>#</span><span>include</span> <span>&lt;emscripten.h&gt;</span></span><p>EMSCRIPTEN_KEEPALIVE<br/><span>int</span> <span>add</span><span>(</span><span>int</span> x<span>,</span> <span>int</span> y<span>)</span> <span>{</span><br/>  <span>return</span> x <span>+</span> y<span>;</span><br/><span>}</span></p></code></pre>
         <p> We'd normally build this with something like <code>emcc -O3 add.c -o add.js</code> which would emit <code>add.js</code> and <code>add.wasm</code>. Instead, let's ask <code>emcc</code> to only emit Wasm: </p>
```

which would fix the test, but is a little unfortunate - the main title is in an `h1` in the source and has already been elided; we shouldn't *also* elide this `h2`, I think.

@jakubriedl , should we maybe only drop the *first* header that seems to match the title? Or did you run into any reason that would be a bad idea?

gijsk:
This fixes the test failures and improves some logging changes (I split out the two parts for clarity) that I ran into while trying to debug some of this.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -893,6 +893,8 @@ Readability.prototype = {
       var elementsToScore = [];
       var node = this._doc.documentElement;
 
+      let shouldRemoveTitleHeader = true;
+
       while (node) {
         var matchString = node.className + " " + node.id;
 
@@ -904,6 +906,13 @@ Readability.prototype = {
 
         // Check to see if this node is a byline, and remove it if it is.
         if (this._checkByline(node, matchString)) {
+          node = this._removeAndGetNext(node);
+          continue;
+        }
+
+        if (shouldRemoveTitleHeader && this._headerDuplicatesTitle(node)) {
+          this.log("Removing header: ", node.textContent.trim(), this._articleTitle.trim());
+          shouldRemoveTitleHeader = false;
           node = this._removeAndGetNext(node);
           continue;
         }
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -60,31 +60,32 @@ function Readability(doc, options) {
...
 
   // Control whether log messages are sent to the console
   if (this._debug) {
-    logEl = function(e) {
-      var rv = e.nodeName + " ";
-      if (e.nodeType == e.TEXT_NODE) {
-        return rv + '("' + e.textContent + '")';
+    let logNode = function(node) {
+      if (node.nodeType == node.TEXT_NODE) {
+        return node.nodeName + ' ("' + node.textContent + '")';
```
msujaws, line 68: Can you use a template literal string here?

Readability.js
```diff
@@ -60,31 +60,32 @@ function Readability(doc, options) {
...
-      var elDesc = "";
-      if (e.id)
-        elDesc = "(#" + e.id + classDesc + ")";
-      else if (classDesc)
-        elDesc = "(" + classDesc + ")";
-      return rv + elDesc;
+      return "<" + node.localName + " " + (
+        Array.from(node.attributes, function(attr) {
+          return attr.name + "='" + attr.value + "'";
+        }).join(" ") + ">");
```
msujaws, line 73:
```
let attrPairs = Array.from(node.attributes, function(attr) {
  return `${attr.name}="${attr.value}"`;
}).join(" ");
return `<${node.localName} ${attrPairs}>`;
```

msujaws (approved):

Issue #605, opened 2020-07-24
Send custom serializer through options object to enable XHTML output


danburzo:
I'm trying to bundle some web pages filtered through Readability as EPUB, and the format requires the content to be XHTML. However, an element's `.innerHTML` returns a HTML, not XHTML serialization.

I would like to propose a `serializer` option, that defaults to `serializer: el => el.innerHTML` but which can be swapped to `serializer: (new window.XMLSerializer()).serializeToString` when returning data here:

https://github.com/mozilla/readability/blob/52ab9b5c8916c306a47b2119270dcdabebf9d203/Readability.js#L2056-L2065

gijsk:
I don't really understand - Readability doesn't control the DOM's `innerHTML` output itself. This depends on the DOM implementation used for the DOM you pass readability, right?

Unfortunately readability has to be able to work with a simulated DOM so we cannot assume `window` or `XMLSerializer` exists.

gijsk:
Put differently: can you solve your problem by making sure the DOM you pass readability is an XHTML DOM (as implemented by `jsdom` or similar) ?

danburzo:
Unfortunately, as far as I can tell, `jsdom` can only decide at instantiation time whether a document should be treated as HTML or XML/XHTML via the `contentType` option. However, loading a normal HTML as XML/XHTML understandably fails due to syntax mismatch. Therefore it's only at serialization-time that we can create the XHTML representation.

For a HTML document, `.innerHTML` returns the content formatted as HTML (which is invalid XHTML, and causes the EPUB to stop displaying content). A solution to this problem is to allow Readability to get a custom serializer for the content:

```js
return { 
   title: this._articleTitle, 
   byline: metadata.byline || this._articleByline, 
   dir: this._articleDir, 
   content: options.serializer ? options.serializer(articleContent) : articleContent.innerHTML, 
   textContent: textContent, 
   length: textContent.length, 
   excerpt: metadata.excerpt, 
   siteName: metadata.siteName || this._articleSiteName 
}
```

And call Readability with the appropriate option:

```js
let dom = new JSDOM('content');
new Readability(dom.window.documentElement, {
  serializer: (new dom.window.XMLSerializer()).serializeToString
});
```

There _may_ be a way to hijack the `innerHTML` implementation to point to XMLSerializer, but I'm not sure what effect it has on Readability's function, since it's used throughout the code.

danburzo:
[...]

PR #667, opened 2021-01-01
Add typescript definition

mathieudutour:
This PR adds the type definitions of the library.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,49 @@
+/**
+ * Decides whether or not the document is reader-able without parsing the whole thing.
+ * @return {boolean} Whether or not we suspect Readability.parse() will suceeed at returning an article object.
+ */
+export function isProbablyReaderable(
+  document: Document,
+  options?: {
+    /** The minimum node content length used to decide if the document is readerable. */
+    minContentLength?: number;
+    /** The minumum cumulated 'score' used to determine if the document is readerable. */
+    minScore?: number;
+    /** The function used to determine if a node is visible. */
+    visibilityChecker?: (node: Node) => boolean;
+  }
+): boolean;
+
+export class Readability {
+  constructor(
+    document: Document,
+    options?: {
+      debug?: boolean;
+      maxElemsToParse?: number;
+      nbTopCandidates?: number;
+      charThreshold?: number;
+      classesToPreserve?: string[];
+      keepClasses?: boolean;
+      serializer?: (node: Node) => string;
+      disableJSONLD?: boolean;
+    }
+  );
+
+  parse(): null | {
+    /** article title */
+    title: string;
+    /** author metadata */
+    byline: string;
+    /** content direction */
+    dir: string;
+    /** HTML string of processed article content */
+    content: string;
+    /** text content of the article (all HTML removed) */
+    textContent: string;
+    /** length of an article, in characters */
+    length: number;
+    /** article description, or short excerpt from the content */
+    excerpt: string;
+    siteName: string;
+  };
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,7 +14,7 @@ export function isProbablyReaderable(
   }
 ): boolean;
 
-export class Readability {
+export class Readability<T = string> {
   constructor(
     document: Document,
     options?: {
@@ -24,7 +24,7 @@ export class Readability {
       charThreshold?: number;
       classesToPreserve?: string[];
       keepClasses?: boolean;
-      serializer?: (node: Node) => string;
+      serializer?: (node: Node) => T;
       disableJSONLD?: boolean;
     }
   );
@@ -36,8 +36,8 @@ export class Readability {
     byline: string;
     /** content direction */
     dir: string;
-    /** HTML string of processed article content */
-    content: string;
+    /** HTML of processed article content */
+    content: T;
     /** text content of the article (all HTML removed) */
     textContent: string;
     /** length of an article, in characters */
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.4.0",
   "description": "A standalone version of the readability library used for Firefox Reader View.",
   "main": "index.js",
+  "types": "index.d.ts",
   "scripts": {
     "lint": "eslint .",
     "test": "mocha test/test-*.js",


index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(
+    document: Document,
+    options?: {
+      debug?: boolean;
+      maxElemsToParse?: number;
+      nbTopCandidates?: number;
+      charThreshold?: number;
+      classesToPreserve?: string[];
+      keepClasses?: boolean;
+      serializer?: (node: Node) => string;
```
danburzo, line 27: The `serializer` option is used exclusively to obtain the `content` property in the `parse()` call. It's not necessary for it to produce a string. Whatever you return from the serializer function will be reflected in `content`, but I'm not sure how to model that in TypeScript.

mathieudutour: Oh right, thanks. https://github.com/mozilla/readability/pull/667/commits/44a9c32bc407e2688cb521568163169a72ccba62 should do it

gijsk:
Thanks for the PR. Can you clarify what this does? That is, what is the point of the PR? You left the initial comment here empty, I don't know very much about typescript, and this repository doesn't use it.

It seems you're making some kind of type description of the API surface, but it also seems like we're duplicating comments/descriptions of all those API arguments etc. - so my main concerns would be, can we avoid that duplication, and how do we make sure that this stays up-to-date if/when we make API changes.

mathieudutour:
This PR adds the type definitions of the library. It is indeed a description of the API surface that's used by typescript. 

The source code doesn't have to use TS to expose types (as this PR shows) but in order to avoid duplication, it would have to be build with it (or use another build step to somehow generate the types from the comments).

index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+    /** The minimum node content length used to decide if the document is readerable. */
+    minContentLength?: number;
+    /** The minumum cumulated 'score' used to determine if the document is readerable. */
+    minScore?: number;
+    /** The function used to determine if a node is visible. */
+    visibilityChecker?: (node: Node) => boolean;
+  }
+): boolean;
+
+export class Readability<T = string> {
```
gijsk, line 17: I'm a little confused - what other T is supported here? Why not just not have this template-like construction and use `string` literally below?

gijsk (changes requested):
It looks like typescript supports generating API docs from JSDoc comments - https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html . Could that be used here? I also see https://www.npmjs.com/package/tsd-jsdoc and I don't know if that's different or not...

I understand that getting automation to work reliably is perhaps more work than just writing this definition file once, but ISTM we will have a better chance keeping this up to date if there are straightforward steps to update it if/when the API surface changes -- and improved JSDoc comments are probably useful for other things in addition to typescript support.

OTOH, if the automation support doesn't work well or at all for the readability setup, I'm happy to just merge this without adding support for updating it. But I'd like to understand the difficulty with that better before taking on having to learn typescript definition formats and "manually" updating this every time.

index.d.ts
mathieudutour: I'm not sure what you mean by "literally below". This makes sure that the type of `content` will be `T` (which is the return type of `serializer` and defaults to `string` if there is no serializer)

gijsk: Ah, hm. That's... interesting. @danburzo is there a particular reason we shouldn't require it to be a string (even if it can be a non-string right now)? Feels odd to have something called "serializer" that then does other transforms and produces something other than a string.

danburzo: > is there a particular reason we shouldn't require it to be a string (even if it can be a non-string right now)? 

We introduced `serializer` to support other types of DOM serialization (e.g. XHTML/XML for EPUB, #605). In my particular case, where Readability is part of a larger processing pipeline, I found it's actually more efficient to skip an entire serialization/parse pair altogether:

```js
const R = new Readability(dom.window.document, {
        // ...
	/*
		Change Readability's serialization to return 
		a DOM element (instead of a HTML string) 
		as the `.content` property returned from `.parse()`

		This makes it easier for us to run subsequent
		transformations (sanitization, hyphenation, etc.)
		without having to parse/serialize the HTML repeatedly.
	 */
	serializer: el => el
});
```

In my previous comment, I had suggested the change in the TypeScript definition for the sake of accuracy, but if this detail of the `serializer` option obscures the class as a whole, I think we can safely pretend it always returns a string here?

gijsk: Hm. Sorry for not doing a good job at reading before. No-op seems more in keeping (like saying somehow "don't serialize this at all"). We could make it return `Node | string`, assuming the typescript definition file allows for that? Then it won't require parametrization/templatization of the class. We should probably clarify this in the jsdoc and/or docs though.

mathieudutour: So it's a bit hard to to have the return type of `parse` depends on the return type of `serializer` without having it in a template. IMO it's more important to have an exact return type for `parse` rather than for `serializer`.

The template is transparent for the user, they won't ever need to use it so I don't think it's obscuring the class too much?

mathieudutour:
So I had a go at the generation of the types from the JS Doc and it requires a lot of changes to the source code (only cosmetic of course) and the template for the `serializer` isn't working (because of https://github.com/microsoft/TypeScript/issues/29401).

I push it here: https://github.com/mozilla/readability/commit/aec77ec40cbbefcd8968a41f70212fc196a18903. The types are generating automatically before each release (before running `npm run release`) and can be manually generated with `npm run generate-typescript-definition`.

Let me know what you prefer.

gijsk:
> So I had a go at the generation of the types from the JS Doc and it requires a lot of changes to the source code (only cosmetic of course) and the template for the `serializer` isn't working (because of [microsoft/TypeScript#29401](https://github.com/microsoft/TypeScript/issues/29401)).

Oof. Yeah, that's unfortunate. Let's go with this version for now then, and we can file a follow-up issue to consider if we can do this in an automated fashion. I'd want to avoid all the additional `@private` annotations (feels like there should be an opt-in way specifying what methods are (supposed to be) exposed, instead of having to opt-out nearly the entire class...), and ideally all the whitespace/cosmetic changes. Perhaps switching to `class Readability` instead of the `Readability.prototype.foo` bit would go better, wrt to that part? Even so, I don't really understand why typescript's generator requires the use of individual property assignments instead of being able to parse `N.prototype = { ... }`, but that's not a battle we should fight here.

Apologies for all the back and forth on this.

I'm going to accept this PR as-is, but I'd like to understand what to put in the `CHANGELOG` file and how this will affect versioning. For typescript users, is this an API-surface-breaking change, or not? Specifically, I guess that if you're a typescript user of this library, and now the library comes with a type specification, and your use does not match the type specification, that'd break typescript compilation, right? If my understanding is correct, per semver I guess this requires a minor (rather than patch) rev to the version because we're pre-1.0. OTOH, in theory, the new definition should be compatible with "correct" existing usages (but it's hard to imagine all of those usages, and perhaps people use the not-intentionally-exposed methods on the class from typescript, which will break after this PR is merged, maybe) ? @mathieudutour do you have thoughts on this?

mathieudutour:
> Perhaps switching to class Readability instead of the Readability.prototype.foo bit would go better, wrt to that part?

It would for sure, but would drop support for some old version of Nodejs and some browsers.

>  I don't really understand why typescript's generator requires the use of individual property assignments

Me neither, I got a really cryptic error without them ¯\_(ツ)_/¯

> I guess that if you're a typescript user of this library, and now the library comes with a type specification, and your use does not match the type specification, that'd break typescript compilation, right?

You're right but I tend to think that TS users would have added local types to cover the library (that's what we did) and that use case won't break (the local types will take precedence). I usually release types under a patch version, especially pre 1.0, but I don't have a strong opinion

gijsk:
(I ended up settling on a patch release as suggested by Mathieu, so this is now in 0.4.1)

mathieudutour:
Awesome, could you publish the version on npm? Thanks

gijsk:
> ? Publish @mozilla/readability to npm? Yes
> ? Commit (Release 0.4.1)? Yes
> ? Tag (0.4.1)? Yes
> ? Push? Yes
> 🔗 https://www.npmjs.com/package/@mozilla/readability
> 🏁 Done (in 23s.)

🤔 

(it's not actually there, you're absolutely right - but I didn't notice.)

gijsk:
I re-ran `npm publish` manually from the 0.4.1 tag, and it looks like that did the trick. Apologies for missing that.

mathieudutour:
No worries, thanks a lot!

PR #668, opened 2021-02-02
Force LF eol for windows os

cckang:
Force JS file in Windows uses LF eol to match the repository style guidelines.


gijsk:
This is great, thank you!

PR #673, opened 2021-02-22
Reorganize and expand README

### LINKED ISSUE #672: Reorganize README
I'd like to attempt reorganizing the README and include more details on the following:

* The constructor options
* The shape of the object returned by `Readability.parse()`
* The `Readability.isReaderable()` function
* Usage in the browser vs. Node.js 

I thought I'd open an issue to get confirmation before starting a PR.

danburzo:
Fixes #672. I'll add some comments inline to clarify some decisions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,24 @@ var article = new Readability(documentClone).parse();
 
 A quick-and-dirty way of figuring out if it's plausible that the contents of a given document are suitable for processing with Readability. It is likely to produce both false positives and false negatives. The reason it exists is to avoid bogging down a time-sensitive process (like loading and showing the user a webpage) with the complex logic in the core of Readability. Improvements to its logic (while not deteriorating its performance) are very welcome.
 
+The `options` object accepts a number of properties, all optional:
+
+* `minContentLength` (number, default `140`): the minimum node content length used to decide if the document is readerable;
+* `minScore` (number, default `20`): the minumum cumulated 'score' used to determine if the document is readerable;
+* `visibilityChecker` (function, default `isNodeVisible`): the function used to determine if a node is visible;
+
+The function returns a boolean corresponding to whether or not we suspect `Readability.parse()` will suceeed at returning an article object. Here's an example:
+
+```js
+/* 
+    Only instantiate Readability  if we suspect 
+    the `parse()` method will produce a meaningful result.
+*/
+if (isProbablyReaderable(document)) {
+    let article = new Readability(document).parse();
+}
+```
+
 ## Node.js usage
 
 Readability is available on npm:
--- a/README.md
+++ b/README.md
@@ -2,38 +2,56 @@
 
 A standalone version of the readability library used for Firefox Reader View.
 
-## Usage on the web
+## Basic usage
 
-To parse a document, you must create a new `Readability` object from a DOM document object, and then call `parse()`. Here's an example:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,86 +2,85 @@
...
-
-### Example:
+Since Node.js does not come with its own DOM implementation, we rely on external libraries like [jsdom](https://github.com/jsdom/jsdom). Here's an example using `jsdom` to obtain a DOM document object:
 
 ```js
 var { Readability } = require('@mozilla/readability');
-var JSDOM = require('jsdom').JSDOM;
-var doc = new JSDOM("<body>Here's a bunch of text</body>", {
+var { JSDOM } = require('jsdom');
+var doc = new JSDOM("<body>Look at this cat: <img src='./cat.jpg'></body>", {
```
danburzo, line 67: Added an `` with a relative `src` to tie to the next paragraph, explaining the purpose of the `url` option in JSDOM.

README.md
```diff
@@ -2,86 +2,85 @@
...
+```js
 var documentClone = document.cloneNode(true); 
 var article = new Readability(documentClone).parse();
 ```
 
-## Usage from Node.js
+### `isProbablyReaderable(document, options)`
+
+A quick-and-dirty way of figuring out if it's plausible that the contents of a given document are suitable for processing with Readability. It is likely to produce both false positives and false negatives. The reason it exists is to avoid bogging down a time-sensitive process (like loading and showing the user a webpage) with the complex logic in the core of Readability. Improvements to its logic (while not deteriorating its performance) are very welcome.
+
```
danburzo, line 53: TODO: add an example
TODO: document options.

README.md
```diff
@@ -2,86 +2,103 @@
...
 Readability is available on npm:
 
 ```bash
 npm install @mozilla/readability
 ```
 
-In Node.js, you won't generally have a DOM document object. To obtain one, you can use external
-libraries like [jsdom](https://github.com/jsdom/jsdom). While this repository contains a parser of
-its own (`JSDOMParser`), that is restricted to reading XML-compatible markup and therefore we do
-not recommend it for general use.
```
danburzo, line 47: I've removed the mention of `JSDOMParser` — if I understand correctly, people don't actually need to know that it exists. It may be useful, however, to add a recipe for parsing XHTML with `jsdom`? I don't remember exactly, but by default it might choke on the weirder things in XHTML (`CDATA` sections, etc.)

README.md
```diff
@@ -2,86 +2,103 @@
...
+    let article = new Readability(document).parse();
+}
+```
+
+## Node.js usage
 
 Readability is available on npm:
 
 ```bash
 npm install @mozilla/readability
```
danburzo, line 77: I think currently the npm package is a bit buried under the fold, and people may be inclined to grab [the _other_ `readability`](https://npmjs.com/package/readability) before reading. Maybe start with an _Installation_ section, or work it into the _Basic usage_ section?

danburzo:
Alright, this would be my first pass. Looking for feedback, and otherwise feel free to just pull this in and modify as needed, if it saves time.

README.md
gijsk: Either of those options sound good to me - I definitely agree moving this higher up is a good idea. I'll give this a shot after merging.

gijsk:
Awesome, thanks!

PR #684, opened 2021-04-03
Add iframe support for cdn.embedly.com in REGEXPS.videos

cayolblake:
This pull request adds support for extracting iframes with src from cdn.embedly.com which is common on many websites.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -130,7 +130,7 @@ Readability.prototype = {
     byline: /byline|author|dateline|writtenby|p-author/i,
     replaceFonts: /<(\/?)font[^>]*>/gi,
     normalize: /\s{2,}/g,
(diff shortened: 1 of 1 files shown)


gijsk:
Thank you for this PR! Do you think it'd be possible to add a testcase to verify this works?

cayolblake:
Most welcomed. Planning to do some more soon.

About test cases, not sure how to do that, I can try if there're some sort of guide or a tutorial on doing so.

gijsk:
You can run the [`generate-testcase.js`](https://github.com/mozilla/readability/blob/master/test/generate-testcase.js) script with `node` and pass it a slug (probably `embedly` would be a good one here) and a URL (to a test page) and it will fetch that page's HTML, run readability on it, and create all the files (though you'll need to `git add` them to a commit of course).

Niek (approved):

cgorrieri:
I see this was opened for a while and more the reason to make this video regex configurable with a default value.
https://github.com/mozilla/readability/issues/787

Issue #678, opened 2021-03-07
Bug in extracting images from NewYorkTimes website


cayolblake:
Hello,

I'm using this library as standalone, links like https://www.nytimes.com/2021/02/04/nyregion/malliotakis-trump.html shows only first two images in article.

I verified same behavior using `Firefox Nightly/Developer Edition` ends with the same result in Reader Mode.

goleary:
I see all 3, but they are all duplicated
![image](https://user-images.githubusercontent.com/16123225/120676692-278bd980-c44b-11eb-8acf-9e56ad5ac297.png)

PR #693, opened 2021-05-12
Improve images detection

thomasdao:
Hi, thanks for the great library. I notice that Readability sometimes remove images from website, like in this ticket #678 and a few websites that I have tested like https://www.news.com.au/sport/american-sports/nba/michael-jordans-reveals-final-texts-to-kobe-bryant/news-story/43297745901f8e10a60cda2ad4ce2e8f (and all articles on that site).

This commit improves the algorithm for cleaning up elements by:
- If the image size is big, increase its weight
- When clean up an element, first calculate total weight of the element from its own weight and its children weight

Please help review, thank you!

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1807,30 +1807,14 @@ Readability.prototype = {
     return weight;
   },
 
-  _contains (parent, child) {
-    if (parent.contains) {
-      return parent.contains(child);
-    }
-
-    // For JSDOM, which does not have contains function
-    let p = child.parentNode;
-    while (p) {
-      if (p === parent) {
-        return true;
-      }
-      p = p.parentNode;
-    }
-
-    return false;
-  },
-
   _getTotalClassWeight: function(node) {
     let weight = this._getClassWeight(node);
-    let nextNode = this._getNextNode(node, false);
-    while (nextNode && this._contains(node, nextNode)) {
-      weight += this._getClassWeight(nextNode);
-      nextNode = this._getNextNode(nextNode, false);
+
+    let children = node.getElementsByTagName('*')
+    for (const child of children) {
+      weight += this._getClassWeight(child);
     }
+
     return weight;
   },
 
--- a/Readability.js
+++ b/Readability.js
@@ -1778,9 +1778,9 @@ Readability.prototype = {
     var weight = 0;
 
     // If the element is an image and size is big enough, increase weight
-    if (e.tagName === 'IMG') {
+    if (e.tagName === "IMG") {
       if (e.width > 400 && e.height > 400) {
-        weight += 50
+        weight += 50;
       } else if (e.width > 200 && e.height > 200) {
         weight += 25;
       }
--- a/Readability.js
+++ b/Readability.js
@@ -1777,15 +1777,6 @@ Readability.prototype = {
 
     var weight = 0;
 
-    // If the element is an image and size is big enough, increase weight
-    if (e.tagName === "IMG") {
-      if (e.width > 400 && e.height > 400) {
-        weight += 50;
-      } else if (e.width > 200 && e.height > 200) {
-        weight += 25;
-      }
-    }
-
     // Look for a special classname
     if (typeof(e.className) === "string" && e.className !== "") {
       if (this.REGEXPS.negative.test(e.className))
@@ -1802,17 +1793,6 @@ Readability.prototype = {
 
       if (this.REGEXPS.positive.test(e.id))
         weight += 25;
-    }
-
-    return weight;
-  },
-
-  _getTotalClassWeight: function(node) {
-    let weight = this._getClassWeight(node);
-
-    let children = node.getElementsByTagName("*");
-    for (const child of children) {
-      weight += this._getClassWeight(child);
     }
 
     return weight;
@@ -2083,9 +2063,23 @@ Readability.prototype = {
         return false;
       }
 
-      // Calculate the weight of a node based on its own weight
-      // plus its children weight
-      var weight = this._getTotalClassWeight(node);
+      var hasImage = false;
+      if (node.querySelector) {
+        if (node.querySelector("img")) {
+          hasImage = true;
+        }
+      } else {
+        var images = node.getElementsByTagName("img");
+        if (images && images.length > 0) {
+          hasImage = true;
+        }
+      }
+
+      if (hasImage) {
+        return false;
+      }
+
+      var weight = this._getClassWeight(node);
 
       this.log("Cleaning Conditionally", node);
 
--- a/Readability.js
+++ b/Readability.js
@@ -1810,7 +1810,7 @@ Readability.prototype = {
   _getTotalClassWeight: function(node) {
     let weight = this._getClassWeight(node);
 
-    let children = node.getElementsByTagName('*');
+    let children = node.getElementsByTagName("*");
     for (const child of children) {
       weight += this._getClassWeight(child);
     }
--- a/Readability.js
+++ b/Readability.js
@@ -1777,6 +1777,15 @@ Readability.prototype = {
 
     var weight = 0;
 
+    // If the element is an image and size is big enough, increase weight
+    if (e.tagName === 'IMG') {
+      if (e.width > 400 && e.height > 400) {
+        weight += 50
+      } else if (e.width > 200 && e.height > 200) {
+        weight += 25;
+      }
+    }
+
     // Look for a special classname
     if (typeof(e.className) === "string" && e.className !== "") {
       if (this.REGEXPS.negative.test(e.className))
@@ -1795,6 +1804,16 @@ Readability.prototype = {
         weight += 25;
     }
 
+    return weight;
+  },
+
+  _getTotalClassWeight: function(node) {
+    let weight = this._getClassWeight(node);
+    let nextNode = this._getNextNode(node, false);
+    while (nextNode && node.contains(nextNode)) {
+      weight += this._getClassWeight(nextNode);
+      nextNode = this._getNextNode(nextNode, false);
+    }
     return weight;
   },
 
@@ -2063,7 +2082,9 @@ Readability.prototype = {
         return false;
       }
 
-      var weight = this._getClassWeight(node);
+      // Calculate the weight of a node based on its own weight
+      // plus its children weight
+      var weight = this._getTotalClassWeight(node);
 
       this.log("Cleaning Conditionally", node);
 
--- a/Readability.js
+++ b/Readability.js
@@ -1807,10 +1807,27 @@ Readability.prototype = {
     return weight;
   },
 
+  _contains (parent, child) {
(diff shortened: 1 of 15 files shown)


gijsk (changes requested):
Thanks for the PR, and sorry for the slow review.

A number of tests are failing with this change. In some cases this is because we're now including images that we weren't before, which might be a good thing - in other cases, it's less clear. You can re-generate the expected markup using `node test/generate-testcase.js wapo-1` (for instance - it'll work with any existing test subdirectoory), and if you then commit it and make it part of this PR, the test should pass and we can see what changes this change to the algorithm causes, and evaluate whether those make sense.

In terms of the implementation, it also appears you've introduced a performance issue, so I don't think we can merge this as-is.

I'm guessing (but can't easily tell from just the code) that your aim is to score images and their ancestors higher to avoid them getting dropped, but the way you're increasing the ancestor scores is causing some performance issues.

Instead of looking for image scoring nodes in all descendants of every node using `node.getElementsByTagName("*")`, and potentially doing that repeatedly, perhaps in the pre-processing step we could look just for all the image tags in the document, find the big ones, and then tag all the ancestors with an image bonus property, that the scoring algorithm reads? That should resolve the performance issue, and will hopefully make the code easier to follow, and reduce side effects (right now, class positive/negative matches of descendants are impacting ancestors to an increased degree, which I think is also causing some test issues). :-)

thomasdao:
Hi @gijsk, thanks for your review. My observation on site like [news.com.au](https://www.news.com.au/sport/american-sports/nba/michael-jordans-reveals-final-texts-to-kobe-bryant/news-story/43297745901f8e10a60cda2ad4ce2e8f) has the image structure like below:

```html
<div class="w650 h487 image media kskd-media">
  <div class="image-wrapper">
    <img width="650" height="487" src="xxx" class=" image-loaded">
  </div>
  <p class="caption">
    Jordan got emotional at Kobe’s funeral. 
  </p>
</div>
```

Readability calls `_cleanConditionally` on this `div`, and it looks at the class of the `div` and found `media`, which is in the `negative` regex and remove it and its children. I think to determine if the div should be removed, we should look at it and its descendants (because the `img` is not the direct children of `div`). For the structure above, while parent `div` has negative weight, the descendants add weight to the parent so we can have better estimate whether the parent div should be removed.

Is that a reasonable approach? I can improve this further if you think the direction is ok, thank you :)

Readability.js
```diff
@@ -1798,6 +1807,17 @@ Readability.prototype = {
     return weight;
   },
 
+  _getTotalClassWeight: function(node) {
+    let weight = this._getClassWeight(node);
+
+    let children = node.getElementsByTagName("*");
+    for (const child of children) {
```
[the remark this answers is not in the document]
thomasdao: @gijsk I'm not sure why we compute the weight of innermost child 5 times:
```js
for (const child of children) {
  weight += this._getClassWeight(child);
}
```
I'm not calling recursive here, the `_getClassWeight` function is still the same and only calculate the weight for that specific node and not its children or descendants. Can you help clarify?

thomasdao:
I am thinking of a less intrusive change that if a div contains a `img` descendant (maybe with certain width or height), then it should not be removed even if it contains negative class. What do you think?

Readability.js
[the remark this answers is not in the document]
gijsk: @thomasdao so every time that code runs for a given node, it evaluates the class weight of that node and all its descendants,  adding all of them together.

But `_cleanConditionally` gets called a number of times, with different tag names. In the example with the nested divs you gave in the main comment, the call at https://github.com/mozilla/readability/blob/e1972362816cd47e679d1a4bbbf923a58b8a88d5/Readability.js#L706 will run this code for every `div`, so once for the parent div, then again for the child div -- but if there are more `div` elements higher up in the document tree (which is likely!), it'll run this for all of those as well.

So the fact that the code you're adding runs `getElementsByTagName("*")`, collecting all the descendants, for every node on which the inner callback of `_cleanConditionally` is called, is likely what causes us to evaluate nodes more than once.

To fix this, I would suggest that rather than putting this in `_cleanConditionally`, instead collect all `img` elements in the document, loop over those and for all the large ones, walk up their parent tree and store the "image bonus" on the ancestors directly. You could add another "private" (`_`-prefixed) method for this and call it after we unwrap no script images, here: https://github.com/mozilla/readability/blob/e1972362816cd47e679d1a4bbbf923a58b8a88d5/Readability.js#L2212

This also avoids the other issue here, which is how you weight the class weight of the descendants when applying it to ancestors (for which we otherwise use a multiplier factor, https://github.com/mozilla/readability/blob/e1972362816cd47e679d1a4bbbf923a58b8a88d5/Readability.js#L1036-L1046 ).

gijsk:
> I am thinking of a less intrusive change that if a div contains a `img` descendant (maybe with certain width or height), then it should not be removed even if it contains negative class. What do you think?

Right, that seems OK to me - you could do this using my suggestion in https://github.com/mozilla/readability/pull/693#discussion_r634299129 - effectively "tagging" all the ancestors of each `img` that is big enough with some kind of `_dontClean` property, and referring to that in the `_cleanConditionally` code.

thomasdao:
Hi @gijsk, sorry I forgot updating this PR. After further testing, I update the algorithm to simply keep the element if it contains an image. (In the last comment, I was thinking the image should be kept if it is large enough, however many sites did not include width and height attributes in the markup, and to know the width or height might require the page to be fully rendered, so it is not very flexible).

I also re-generate all test cases and go through the changes, I think many of the changes are quite reasonable. Please let me know if you have any feedback, thanks :)

gijsk:
(apologies for the slowness here, I am and have been out on holiday and will look at this when I get back, next week)

thomasdao:
Sure no hurry, please take your time and enjoy your holiday :)

test/test-pages/nytimes-3/expected.html
```diff
@@ -160,6 +160,17 @@ <h2 id="link-13d0a396"> So, where do people learn how to fix some of these issue
...
+                <a href="https://www.nytimes.com/interactive/2016/08/18/nyregion/new-york-101-streets-repair-and-maintenance.html?action=click&amp;module=RelatedLinks&amp;pgtype=Article">
+                    <div>
+                        <p> Why Are New York City’s Streets Always Under Construction? </p>
+                        <p><time datetime="2016-08-18">Aug. 18, 2016</time>
+                        </p>
+                    </div>
+                    <p><img src="https://static01.nyt.com/images/2016/08/12/nyregion/nyc101-01/nyc101-01-videoLarge-v7.jpg" />
+                    </p>
+                </a>
+            </div>
```
gijsk, line 173: I'm not convinced this should stay.

test/test-pages/pixnet/expected.html
```diff
@@ -184,5 +184,10 @@
         <p><span><span><span>內文有不定期的更新旅遊、露營圖文訊息 </span></span><span><span>謝謝!</span></span>
             </span>
         </p>
+        <div>
+            <p><img src="http://panel.pixfs.net/images/icons/tag_blue.gif" original="http://panel.pixfs.net/images/icons/tag_blue.gif" />
+                <a href="http://tags.pixnet.net/blog/user/stevenhgm/%E7%8E%89%E5%B3%B0%E9%81%93%E8%B7%AF" rel="tag" target="_blank">玉峰道路</a>, <a href="http://tags.pixnet.net/blog/user/stevenhgm/%E4%B8%8A%E5%AE%87%E8%80%81%E8%A7%80%E6%99%AF%E5%8F%B0" rel="tag" target="_blank">上宇老觀景台</a>, <a href="http://tags.pixnet.net/blog/user/stevenhgm/%E7%BE%8E%E6%A8%B9%E7%87%9F%E5%9C%B0" rel="tag" target="_blank">美樹營地</a>, <a href="http://tags.pixnet.net/blog/user/stevenhgm/%E6%8E%A7%E6%BA%AA%E5%90%8A%E6%A9%8B" rel="tag" target="_blank">控溪吊橋</a>, <a href="http://tags.pixnet.net/blog/user/stevenhgm/%E7%9F%B3%E7%A3%8A%E9%83%A8%E8%90%BD" rel="tag" target="_blank">石磊部落</a>
```
gijsk, line 189: Is this helpful? It's hard for me to be sure as the image is gone and I do not speak the language... but it looks like a series of links, which we normally omit in reader mode because they are not usually part of the core article (there's logic elsewhere that evaluates link density, ie links vs. non-link paragraph text and so on).

test/test-pages/wapo-1/expected.html
```diff
@@ -1,5 +1,37 @@
...
+                    <p>Security forces in Tunisia stormed the country’s most prominent museum, freeing hostages and ending a siege by gunmen.</p>
+                    <p><span>March 18, 2015</span> <span>Tunisian security forces secure the area after gunmen attacked the famed Bardo National Museum in Tunis, the capital.</span>
+                        <span>Fethi Belaid/AFP/Getty Images</span>
+                    </p>
+                    <p><a href="#" target="_blank">Buy Photo</a>
+                    </p>
+                </div>
+            </div>
+            <div>
+                <p>Wait 1 second to continue.</p>
```
gijsk, line 32: Including text like "wait 1 second to continue" and "full screen", "autoplay", "view photos" etc. (but without working links) isn't great. The carousel stuff here and in some of the other pages really looks like it would need more work to make it appear in a reasonable way in reader mode / readability.

test/test-pages/hukumusume/expected.html
```diff
@@ -67,23 +56,6 @@
...
-                            <audio src="http://ohanashi2.up.seesaa.net/mp3/ae_0101.mp3" controls=""></audio>
-                        </td>
-                    </tr>
-                    <tr>
-                        <td>
-                            <a href="http://www.voiceblog.jp/onokuboaki/"><span size="-1">亜姫の朗読☆　イソップ童話より</span></a>
-                        </td>
-                    </tr>
-                </tbody>
-            </table>
```
gijsk, line 86: Hm, why do we now delete this when we did not before? Is this a result of your patch?

gijsk (changes requested):
I'm really sorry but to me, the changes to the tests show that always including images is likely not a good path forward. Although it might work for wikipedia and maybe some of the new york times articles, in a bunch of the other cases, we now end up including extraneous content, broken bits of text, or "see also" or "other article" links that were correctly filtered out before.

I think it'd be worth taking a step back and looking at some examples of cases that motivated this patch, where it's clear that the image should be included, and working out what classes or tags are in use that we could use to ensure those images stay, rather than always including any nodes that contain images (which also ends up including content that has tracking gifs or similar elements).

gijsk:
Perhaps we could look for things labeled "caption" and find the associated image in the document, or something like that?

thomasdao:
Thanks for the feedback, let me think about it more before sending new PR.

PR #694, opened 2021-05-16
Avoid live collection

minas90:
Avoid using childNodes live collection logic and replace with `firstChild` logic.
This small change fixes compatibility with [linkedom](https://github.com/WebReflection/linkedom)

It's already partly fixed with #677 , but some pages still cause infinite loop, like this one: [ansa.it](https://www.ansa.it/canale_saluteebenessere/notizie/sanita/2021/05/16/covid-5.753-positivi-93-vittime.-e-il-dato-piu-basso-da-sette-mesi_4e28d22f-3ded-418c-bab1-2ad7e9a6f9be.html)

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -390,8 +390,8 @@ Readability.prototype = {
           } else {
             // if the link has multiple children, they should all be preserved
             var container = this._doc.createElement("span");
-            while (link.childNodes.length > 0) {
-              container.appendChild(link.childNodes[0]);
+            while (link.firstChild) {
+              container.appendChild(link.firstChild);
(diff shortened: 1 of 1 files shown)


gijsk:
Hm. As far as I can tell there are still more accesses that are likely to be problematic:

https://github.com/mozilla/readability/blob/e1972362816cd47e679d1a4bbbf923a58b8a88d5/Readability.js#L1175-L1177

(we decrement if `append` ends up true, and revisit that item in the list)

gijsk (approved):
Thanks!

PR #695, opened 2021-05-23
Fix doc re: disableJSONLD

01abhishekjain:
Since Readability parses Schema.org fields by default, one needs to set `disableJSONLD` to `true`, not `false`, to skip JSON-LD parsing.


gijsk:
Thanks, and apologies for the oversight...

PR #706, opened 2021-09-06
Fix text element grouping for FONT tag when cleaning up redundant BR nodes

mpazik:
### Issue
Text element grouping after `BR` cleanup, was broken when there was a `FONT` element inside.

The problem is better visible on a page that actually was broken while I used the readability. 
You can see it on [a test branch](https://github.com/mpazik/readability/commit/95afe0de8300eda87bfca2c6d41da6ad7051a905#diff-20fcb902cc028c15c21d866c52af8f15e8ca51a4ba1c151fe5b22f99add2a629L89) when I regenerate the test case after this fix.
Notice broken `<p> [</p>`. The paragraph didn't include other text elements as in the source they contained `FONT` element. Source of that fragment: `[<a name="f1n" id="f1n"><font color="#000000">1</font></a>]`

### Cause
Inside the `_prepDocument,` `_replaceBrs` was executed before `FONT` elements were replaces.

### Solution
As the `FONT` element is not part of `PHRASING_ELEMS` it was not considered a text element, resulting in a broken grouping of elements inside a create wrapping paragraph. 

To fix the problem, we could either add the `FONT` tag to `PHRASING_ELEMS` or make sure that all of the invocations of `_isPhrasingContent` are only invoked on cleaned up dom.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -542,11 +542,11 @@ Readability.prototype = {
     // Remove all style tags in head
     this._removeNodes(this._getAllNodesWithTag(doc, ["style"]));
 
+    this._replaceNodeTags(this._getAllNodesWithTag(doc, ["font"]), "SPAN");
+
     if (doc.body) {
       this._replaceBrs(doc.body);
     }
-
-    this._replaceNodeTags(this._getAllNodesWithTag(doc, ["font"]), "SPAN");
   },
 
   /**
--- a/test/test-pages/hukumusume/expected.html
+++ b/test/test-pages/hukumusume/expected.html
@@ -110,44 +110,67 @@
             <table>
                 <tbody>
                     <tr>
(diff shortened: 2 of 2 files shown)


mpazik:
Closing the PR as there issues with the test :)

mpazik:
The test is failing due to compatibility issues between `JSDOM` and `JSDOMParser` parser. The new resulting HTML is read differently by these two libraries. 

`JSDOM` read fragment
```
<p>
    <span size="-2"><u>
            <p> 366日への旅</p>
        </u></span>
</p>
```
as
```
<p><u> 366日への旅</u></p>
```

To fix that either one of the libraries would need to be adjusted or ideally `readabilty` would simplify the HTML by removing unnecessary wrappers.

PR #712, opened 2021-11-01
fixes a problem on aktualne.cz which puts article title into "headline" in JSON-LD and uses "name" for site name

michaldvorak79:
I encountered an issue on Aktuálně.cz (a major Czech news site) where Readability detects each article title as "Aktuálně.cz" instead of the actual article title. This is caused by the fact that the web site puts its own name into the "name" element of JSON-LD and only puts the article title into the "headline" element. Readability prefers the "name" element in this case, which leads to incorrect results here.

Now, I understand this is probably the website's fault and that the "name" element should contain the article title. But the users probably don't care about whose fault it is, they just see Readability (and Firefox) to present the wrong title. So I'm proposing a solution: when a page's JSON-LD data contains both the "name" and the "headline" elements, Readability will check whether one of them is similar to the HTML title of the page. If so, it will use that one. If not (i.e. if either both match the title, or neither does), it will use the "name" element as usual.

I hope I didn't make any mistake; Javascript is not my forte. But all the unit tests are still passing, including the new one for aktualne.cz which I created, and the lint check doesn't complain, so I hope everything's as it should be.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1389,7 +1389,22 @@ Readability.prototype = {
         ) {
           return metadata;
         }
-        if (typeof parsed.name === "string") {
+        if (typeof parsed.name === "string" && typeof parsed.headline === "string" && parsed.name !== parsed.headline) {
+          // we have both name and headline element in the JSON-LD. They should both be the same but some websites like aktualne.cz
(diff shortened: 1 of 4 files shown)


gijsk:
This looks great, thanks very much for the patch!

PR #713, opened 2021-11-02
parse other JSON-LD elements if the first one is not of a recognized type

michaldvorak79:
One more proposed change regarding JSON-LD. I noticed that the "engadget" test case is not working optimally. The expected-metadata.json file expectes byline to be null. But the source.html actually contains byline information in a JSON-LD element: "Devindra Hardawar". 

The problem is that there are two JSON-LD elements. The first one has type "Review" which Readability doesn't recognize. The second one is "Article", which it recognizes. However, Readability only checks the first JSON-LD element "Review" and when it sees unsupported type, it gives up. 

My proposal is to keep going through the JSON-LD elements until we find one we support. I implemented it in this change request and I changed the byline in engadget/expected-metadata.json to contain the actual author's name. All unit tests are passing and lint doesn't complain, so I hope I did it correctly.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1357,72 +1357,74 @@ Readability.prototype = {
   _getJSONLD: function (doc) {
     var scripts = this._getAllNodesWithTag(doc, ["script"]);
 
-    var jsonLdElement = this._findNode(scripts, function(el) {
-      return el.getAttribute("type") === "application/ld+json";
-    });
+    var metadata;
 
-    if (jsonLdElement) {
-      try {
-        // Strip CDATA markers if present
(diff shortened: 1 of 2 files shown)


gijsk:
This makes sense to me and looks good - thanks for the patch! There were some conflicts with your other patch, which I've attempted to resolve. Just checking that tests pass (ie that I didn't screw it up) and then I'll merge this in.

michaldvorak79:
> There were some conflicts with your other patch, which I've attempted to resolve.

Sorry about that, I was worried that would be the case. The changes were very close to each other. But I figured I'd still better make two pull requests instead of one in case you liked one but not the other.

gijsk:
No worries at all, you did the right thing! Thanks again for both the patches.

PR #715, opened 2021-11-07
Improve isProbablyReaderable()

mevers303:
* Add <div>s that only contain inline text.
* Add parents of <divs> that have a small amount of text and no child elements

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -106,7 +106,7 @@ function isProbablyReaderable(doc, options = {}) {
       return false;
     }
 
-    var textContentLength = node.textContent.trim().replace(/\s+/g, ' ').length;
+    var textContentLength = node.textContent.trim().replace(/\s+/g, " ").length;
     if (textContentLength < options.minContentLength) {
       return false;
     }
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -55,6 +55,21 @@ function isProbablyReaderable(doc, options = {}) {
 
   var nodes = doc.querySelectorAll("p, pre");
 
+  // Get <div> nodes which meets certain criteria them into the `nodes` variable.
+  var set = new Set(nodes);
+  var divs = doc.querySelectorAll("div");
+  if (divs.length) {
+    [].forEach.call(divs, function (node) {
+      // add the parent node of divs that have no child elements and contain a small amount of text
+      if (!node.children.length && node.textContent.length < options.minContentLength) {
+        set.add(node.parentNode);
+      // divs that only contain inline text
+      } else if (![].some.call(node.children, child => !["A", "BR", "SPAN", "PRE"].includes(child.nodeName))) {
+        set.add(node);
+      }
+    });
+  }
+
   // Get <div> nodes which have <br> node(s) and append them into the `nodes` variable.
   // Some articles' DOM structures might look like
   // <div>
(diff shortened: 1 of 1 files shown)


Readability-readerable.js
```diff
@@ -55,6 +55,21 @@ function isProbablyReaderable(doc, options = {}) {
 
   var nodes = doc.querySelectorAll("p, pre");
 
+  // Get <div> nodes which meets certain criteria them into the `nodes` variable.
+  var set = new Set(nodes);
+  var divs = doc.querySelectorAll("div");
+  if (divs.length) {
+    [].forEach.call(divs, function (node) {
+      // add the parent node of divs that have no child elements and contain a small amount of text
```
gijsk, line 63: Why the parent node instead of the node itself?

Readability-readerable.js
```diff
@@ -55,6 +55,21 @@ function isProbablyReaderable(doc, options = {}) {
...
+      if (!node.children.length && node.textContent.length < options.minContentLength) {
+        set.add(node.parentNode);
+      // divs that only contain inline text
+      } else if (![].some.call(node.children, child => !["A", "BR", "SPAN", "PRE"].includes(child.nodeName))) {
```
gijsk, line 67: This would include divs that contain only links, and would also include (all) divs that have no children at all, or no text at all (though they're excluded from the previous `if` condition). I don't think that's a good idea, because it's likely to include items that are tables of content or "see also" references or such. Article text shouldn't (ideally) be full of links. We could extend this check to ensure that there is some other, non-link, non-empty text nodes, perhaps? Instead of only looping over the children, we could add a helper function that loops over `childNodes`, returns `false` as soon as it finds a non-inline child, and returns `true` as soon as it's seen enough inline text to qualify the node? For `a`, `span` and `font` tags we could determine the `textContent` length of the inner child. I'm not sure why `pre` is in the list here...

A nitpick on the actual code: Given how hot this code is, putting the set of inline nodes into a set and using `inlineSet.has(child.nodeName)` seems like a worthwhile optimization.

gijsk (changes requested):
Thanks for the PR.

What are the performance implications of this change? We run this functionality in Firefox after pageload, and due to its nature it has to run on the main thread (unlike the full readability algorithm, which only runs in response to the user asking for the "reader mode" version). Running code on pageload on the main thread is obviously time-sensitive, so we need at least some idea of the performance impact of this change.

Off-hand, I would anticipate that for a lot of pages, also iterating over a bunch of divs is going to have a significant negative impact on performance. We already try to cater to pages that use "bad" HTML (divs) to represent paragraph text, by checking for `div > br`. This isn't sufficient for some pages, which is a trade-off we have so far accepted, to avoid the impact on pageload otherwise... so if we want to change this, we need to be quite careful not to regress performance.

As an aside, to merge this we'd need to update the metadata expectations for a number of the automated tests here to now expect that the `isProbablyReaderable` function returns `true`.

Readability-readerable.js
```diff
@@ -89,7 +106,7 @@ function isProbablyReaderable(doc, options = {}) {
       return false;
     }
 
-    var textContentLength = node.textContent.trim().length;
+    var textContentLength = node.textContent.trim().replace(/\s+/g, " ").length;
```
gijsk, line 109: What's the point of the added `replace` here?

aehlke: to count contiguous whitespace as one character, to not inflate text length of sparse content

Issue #662, opened 2020-12-21
isProbablyReaderable should include `article` tags


gijsk:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1683445 

STR:

1. open e.g. https://blog.nightly.mozilla.org/2020/12/18/these-weeks-in-firefox-issue-85/ in Firefox

ER:
reader mode icon

AR:
no reader mode icon

This is because `isProbablyReaderable` discounts all the paragraphs as they're invisible or too short, and doesn't include the `article` content. The main content is all in `li` tags which are also ignored (and we probably don't want to change that).

gijsk:
@ritik307 this may be a more suitable issue to look at? Basically, we need to add `article` to https://github.com/mozilla/readability/blob/2d358266560e669f13932d6197c6cff958bc3df6/Readability-readerable.js#L56 and add a test using the URL from my initial comment. Let me know if this works for you and/or if you have any questions about this. Thanks for wanting to contribute!

tigleym:
We have an MSU student who will be helping us improve Reader Mode this semester. But I can't seem to assign them. @gijsk Would you be able to do that? Their username is: https://github.com/tthhtao

gijsk:
@tigleym I think they need to comment here first, cf https://github.blog/2019-06-25-assign-issues-to-issue-commenters/

tthhtao:
Hi, this is Jintao. I am interested in working on this bug.

tigleym:
Fixed by #724

PR #720, opened 2022-01-15
make isProbablyReaderable include article tags/content

tthhtao:


tigleym:
I feel like the failing `isProbablyReaderable` tests here should be considered "readerable" now that we're allowing content for `<article>` tags. But perhaps I'm missing something here. 

I'm proposing we set the `readerable` properties here to true? Although these are generated files so we'd need the original URL to update these tests...

tigleym:
@tthhtao Could you check the failing tests are the ones that contain <article> tags with content inside of them? You can find the failure logs here https://community-tc.services.mozilla.com/tasks/WalqZxjeSJ6ys0mRgafRjg/runs/0/logs/public/logs/live.log

tigleym (changes requested):
We still have some tests that fail as a result of this patch. @tthhtao could you look into this?

tigleym:
#662

PR #721, opened 2022-01-16
Readability.parse should expose lang attribute of original document.

micha154:


Readability.js
```diff
@@ -896,6 +896,14 @@ Readability.prototype = {
       let shouldRemoveTitleHeader = true;
 
       while (node) {
+
+        if(node.tagName === "HTML") {
```
tigleym, line 900: This should have a whitespace between the `if` and opening  bracket. This is also causing the tests to fail: https://community-tc.services.mozilla.com/tasks/groups/Lqf0BNjvQoSSmbvA9CS_AA 

Once that's addressed you should run `npm run lint && npm test` to make sure we don't have any other linting/test failures. See contributing guide [here](https://github.com/mozilla/readability/blob/master/CONTRIBUTING.md#tests) for more info.

tigleym (changes requested):

Readability.js
```diff
@@ -896,6 +896,14 @@ Readability.prototype = {
...
+          var articleLang = node.getAttribute("lang");
+          if(articleLang) {
+            this._articleLang = articleLang;
```
tigleym, line 903: I don't think it's necessary for us to have this check here, actually. We could just get the "lang" attribute  and directly assign it.

```
this._articleLang = node.getAttribute("lang");
```

micha154: if (node.tagName === "HTML") {
          var articleLang = node.getAttribute("lang");
          this._articleLang = articleLang;
        }
I just want to make sure I understand, like this? And then would I run "npm run lint && npm test" in readability not mozilla-unified, correct?

tigleym: Yeah! Although I think we can (and this is more of a nitpick)  reduce this by one more line:

```
    if (node.tagName === "HTML") {
      this._articleLang = node.getAttribute("lang");
    }
```

And you're correct. You should run `npm run lint && npm test` from the readability repo.

Readability.js
```diff
@@ -2263,6 +2268,7 @@ Readability.prototype = {
       title: this._articleTitle,
       byline: metadata.byline || this._articleByline,
       dir: this._articleDir,
+      lang: this._articleLang,
```
tigleym, line 2271: Could we make sure we have a test for this? These are the steps you should take for this:

1. Assert that Readability.parse() extracts the lang attribute. You should add this after this line https://github.com/mozilla/readability/blob/master/test/test-readability.js#L178. 
2. Generate a test case where the original document has a "lang" attribute on the root. You can generate this test case using the following:

```
node test/generate-testcase.js nytimes-5 https://www.nytimes.com/es/
```

Once you run that command you can manually check the expected-metadata.json file under the generated directory. If you used the above example then the "lang" property will have an "es" value.

3. Make sure to run `npm test && npm run lint`

tigleym (changes requested):
Awesome! We're really close. I think once we have a test case for this we can get this merged :) 

I added some steps you can take to get this test. You can also refer to [this](https://github.com/mozilla/readability/blob/master/CONTRIBUTING.md#add-new-tests) for additional documentation.

test/test-readability.js
```diff
@@ -176,6 +176,10 @@ function runTestsWithItems(label, domGenerationFn, source, expectedContent, expe
     expectedMetadata.dir && it("should extract expected direction", function() {
       expect(result.dir).eql(expectedMetadata.dir);
     });
+
+    expectedMetadata.lang && it("should extract expected direction", function() {
```
tigleym, line 180: We should update the message here to "should extract expected lang"

tigleym (approved):
Awesome! This looks good. R+ from me after addressing the assertion message

test/test-readability.js
micha154: Sorry, I don't know how I missed that, should be good now! :)

PR #724, opened 2022-01-24
Is probably readable fix for article tags

tthhtao:
Tried to fix the failing tests


tigleym (approved):
Thank you! This looks good to me.

PR #725, opened 2022-01-26
Fix nytimes-5 test case

tigleym:


niklasbaumgardner (approved):
Fixes the test failure. Thanks!

PR #737, opened 2022-02-28
Updated Readability Title Parsing

tkabaker:
This change updated Readability.js's title parsing to match that of Firefox's standard browser view. The tests files have been updated to reflect this change. 

Related to: https://bugzilla.mozilla.org/show_bug.cgi?id=1134363 and https://phabricator.services.mozilla.com/D139614

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1443,6 +1443,9 @@ Readability.prototype = {
         }
       }
     });
+    if (metadata && doc.title) {
+      metadata.title = doc.title;
+    }
     return metadata ? metadata : {};
   },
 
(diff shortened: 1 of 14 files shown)


niklasbaumgardner (changes requested):
I left a comment. I'm curious to what you think about always returning the title? Currently it only returns the title if `metadata` exists.

Readability.js
```diff
@@ -1443,6 +1443,9 @@ Readability.prototype = {
         }
       }
     });
+    if (metadata && doc.title) {
+      metadata.title = doc.title;
+    }
```
niklasbaumgardner, line 1448: Should we return the `doc.title` even if `metadate` is null?

Something like
```
if (doc.title) {
  if (!metadata) {
    metadata = {};
  }
  metadata.title = doc.title;
}
```

Issue #697, opened 2021-06-01
Lists of images not showing


seebz:
I noticed that lists of image are not showing (same for `<figure>`with image).

---

Works :  
```html
<p><img src="pic.jpg"></p>
```

Not working :
```html
<ul>
  <li><img src="pic.jpg"></li>
  <li><img src="pic.jpg"></li>
  <li><img src="pic.jpg"></li>
</ul>
```

---

A test page is available at https://dev418.fr/readability/

Tuxy222:
I'd like to be assigned to this bug

tigleym:
Thanks @Tuxy222 ! Assigned :)

PR #738, opened 2022-03-09
Fixed Lists of Images Not Showing

Tuxy222:
This change allows for lists of images within the content of the page to be present in reader mode. 

Related to: https://github.com/mozilla/readability/issues/697

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -2129,7 +2129,7 @@ Readability.prototype = {
         var contentLength = this._getInnerText(node).length;
 
         var haveToRemove =
-          (img > 1 && p / img < 0.5 && !this._hasAncestorTag(node, "figure")) ||
+          (!isList && img > 1 && p / img < 0.5 && !this._hasAncestorTag(node, "figure")) ||
           (!isList && li > p) ||
           (input > Math.floor(p/3)) ||
           (!isList && headingDensity < 0.9 && contentLength < 25 && (img === 0 || img > 2) && !this._hasAncestorTag(node, "figure")) ||
--- a/Readability.js
+++ b/Readability.js
@@ -2137,6 +2137,24 @@ Readability.prototype = {
           (weight >= 25 && linkDensity > 0.5) ||
           ((embedCount === 1 && contentLength < 75) || embedCount > 1);
 
+        if (haveToRemove) {
+          // Check for lists to allow lists of images to remain in the page
+          if (isList) {
+            for (var x = 0; x < node.children.length; x++) {
+              if (node.children[x].localName == "li") {
+                if (node.children[x].children.length > 1) {
+                  return true;
+                }
+              } else {
+                return true;
+              }
+            }
+            li_count = node.getElementsByTagName("li").length;
+            if (img > 1 && img <= (li_count)) {
+              return false;
+            }
+          }
+        }
         return haveToRemove;
       }
       return false;
--- a/Readability.js
+++ b/Readability.js
@@ -2129,7 +2129,7 @@ Readability.prototype = {
         var contentLength = this._getInnerText(node).length;
 
         var haveToRemove =
-          (!isList && img > 1 && p / img < 0.5 && !this._hasAncestorTag(node, "figure")) ||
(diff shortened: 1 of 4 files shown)


Readability.js
```diff
@@ -2109,6 +2109,7 @@ Readability.prototype = {
         var embedCount = 0;
         var embeds = this._getAllNodesWithTag(node, ["object", "embed", "iframe"]);
 
+
```
niklasbaumgardner, line 2112: nit: We can remove this empty line

Readability.js
```diff
@@ -2136,6 +2137,29 @@ Readability.prototype = {
...
+                if (node.children[x].children.length > 1) {
+                  return true;
+                }
+              } else {
+                // Expected li tag -- remove this element
+                return true;
+              }
+            }
+            // Allow lists of images to remain, so long as img count doesn't exceed li count
+            if (img > 1 && img <= (li+100)) {
```
niklasbaumgardner, line 2158: Why do we add 100 to `li` here?

niklasbaumgardner, line 2158: Why do we add 100 to `li` here?

Readability.js
```diff
@@ -2136,6 +2137,29 @@ Readability.prototype = {
           (!isList && weight < 25 && linkDensity > 0.2) ||
           (weight >= 25 && linkDensity > 0.5) ||
           ((embedCount === 1 && contentLength < 75) || embedCount > 1);
+
+        if (haveToRemove) {
+          // Check for lists of images
```
niklasbaumgardner, line 2142: Instead of having comments on each line, could we have one comment at the beginning explaining why we are doing this check?

niklasbaumgardner, line 2142: Instead of having comments on each line, could we have one comment at the beginning explaining why we are doing this check?

niklasbaumgardner (changes requested):
I just have a few things to address. Thanks for working on this!

Readability.js
```diff
@@ -2136,6 +2136,25 @@ Readability.prototype = {
...
+                  return true;
+                }
+              } else {
+                return true;
+              }
+            }
+            li_count = node.getElementsByTagName("li").length;
+            if (img > 1 && img <= (li_count)) {
+              return false;
+            }
```
tigleym, line 2155:
```suggestion
              let child = node.children[x];
              if (child.localName == "li" && child.children.length) {
                  return true;
              }
            li_count = node.getElementsByTagName("li").length;
            if (img > 1 && img <= (li_count)) {
              return false;
            }
```

tigleym: I think it would be a good idea to document here why we remove some images for this specific case. It's not immediately clear to me here.

tigleym (changes requested):

tigleym:
It looks like we have one failing test here. Could you investigate it?

Readability.js
```diff
@@ -2136,6 +2136,21 @@ Readability.prototype = {
           (!isList && weight < 25 && linkDensity > 0.2) ||
           (weight >= 25 && linkDensity > 0.5) ||
           ((embedCount === 1 && contentLength < 75) || embedCount > 1);
+        // Allow simple lists of images to remain in pages
+        if (isList && haveToRemove) {
+          for (var x = 0; x < node.children.length; x++) {
+            let child = node.children[x];
+            // Don't filter in lists with lis that contain more than one child
```
niklasbaumgardner, line 2143:
```suggestion
            // Don't filter in lists with li's that contain more than one child
```

niklasbaumgardner: just a nit

niklasbaumgardner (changes requested):
I just have a nit to fix. Otherwise, I think this is good

tigleym (approved):
Thank you!

niklasbaumgardner (approved):

tholu:
Will this be merged? Or why not?

gijsk:
I think this just got lost. I just merged it.

gijsk:
(Thanks for the ping!)

PR #746, opened 2022-04-15
Fix articles showing cookie information in reader mode

tthhtao:
Applying the `aria-modal` property to an element with `role="dialog"` replaces the technique of using `aria-hidden` in the background. So I think this kind of element should also be considered "invisible" to the reader.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1357,87 +1357,88 @@ Readability.prototype = {
   _getJSONLD: function (doc) {
     var scripts = this._getAllNodesWithTag(doc, ["script"]);
 
-    var jsonLdElement = this._findNode(scripts, function(el) {
-      return el.getAttribute("type") === "application/ld+json";
-    });
+    var metadata;
 
-    if (jsonLdElement) {
-      try {
-        // Strip CDATA markers if present
-        var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[|\]\]>\s*$/g, "");
-        var parsed = JSON.parse(content);
-        var metadata = {};
-        if (
-          !parsed["@context"] ||
-          !parsed["@context"].match(/^https?\:\/\/schema\.org$/)
-        ) {
-          return metadata;
-        }
-
-        if (!parsed["@type"] && Array.isArray(parsed["@graph"])) {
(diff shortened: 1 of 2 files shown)


Readability.js
```diff
@@ -904,6 +904,12 @@ Readability.prototype = {
           continue;
         }
 
+        // User is not able to see elements applied with both "aria-modla = true" and "role = dialog"
+        if (node.getAttribute("aria-modal") == "true" && node.getAttribute("role") == "dialog") {
+          node = this._removeAndGetNext(node);
+          continue;
+        }
```
tigleym, line 911: This should be added to the check in [isProbablyVisible](https://github.com/tthhtao/readability/blob/2edee0ee6224b9c0ed94f81832cb5e2678d0f0d4/Readability.js#L2206).

tigleym (changes requested):
Thanks @tthhtao ! This works nicely, but I think we should address the comment I have here.

Readability.js
```diff
@@ -2202,7 +2208,9 @@ Readability.prototype = {
     return (!node.style || node.style.display != "none")
       && !node.hasAttribute("hidden")
       //check for "fallback-image" so that wikimedia math images are displayed
-      && (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") != "true" || (node.className && node.className.indexOf && node.className.indexOf("fallback-image") !== -1));
+      && (!node.hasAttribute("aria-hidden") || node.getAttribute("aria-hidden") != "true" || (node.className && node.className.indexOf && node.className.indexOf("fallback-image") !== -1))
+      // user can not see element applied with both attribute "aria-modla = true" and "role = dialog"
```
tigleym, line 2212: "arial-modal"

Readability.js
tigleym: We should remove this check since it should be handled in the one above now

tigleym (changes requested):

gijsk:
I think all the review comments were addressed so let's get this merged.

gijsk:
@tthhtao sorry this sat for such a long time!

PR #750, opened 2022-05-03
Optimize scanning process and prevent HTML element from being stripped

foxdavidj:
This PR changes two things:

1. When a website explicitly declares where the main content is via `<main>` or the aria role `role=main` we should respect this and skip the scanner to start at this element. This ensures we always skip over repetitive header information.
2. I've had a few websites that have the keyword "header" in the classname of the HTML element, breaking this entire process. For example they will have the classname "is-logged-in has-header-button"

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -891,7 +891,12 @@ Readability.prototype = {
       // class name "comment", etc), and turn divs into P tags where they have been
       // used inappropriately (as in, where they contain no other block level elements.)
       var elementsToScore = [];
-      var node = this._doc.documentElement;
+      
+      // Start scanning at the element declared to contain the main content if it exists
+      var node = this._doc.querySelector('main, [role=main]');
+      if (!node) {
+        node = this._doc.documentElement;
+      }
 
       let shouldRemoveTitleHeader = true;
 
@@ -929,6 +934,7 @@ Readability.prototype = {
               !this._hasAncestorTag(node, "table") &&
               !this._hasAncestorTag(node, "code") &&
               node.tagName !== "BODY" &&
+              node.tagName !== "HTML" &&
               node.tagName !== "A") {
             this.log("Removing unlikely candidate - " + matchString);
             node = this._removeAndGetNext(node);


gijsk (changes requested):
Thanks for the patch!

Code in readability needs to work with the `JSDOMParser.js` library included in it, and your patch relies on `querySelector` with elaborate selector support, which isn't present there, so unfortunately it won't work as-is.

Additionally, it would be good to have some testcases for where/when this solves a concrete issue you're running into?

Finally, it seems there are some linting failures in the patch, which is also causing the tests not to run (so we can't see what the effect on the existing testcases is - that would probably at least show the `querySelector` issue).

foxdavidj:
Thanks for the reply! Do you allow any kind of progressive enhancement? Love having everything functional via the libraries bundled within, but being able to use libraries like querySelector and getComputedStyle when available would help increase efficiency and accuracy.

gijsk:
> Thanks for the reply! Do you allow any kind of progressive enhancement? Love having everything functional via the libraries bundled within, but being able to use libraries like querySelector and getComputedStyle when available would help increase efficiency and accuracy.

Well, checking for a `main` element doesn't require use of `querySelector` (`getElementsByTagName` works fine). The `role` thing you could also implement without relying on `querySelector`, though it would be a bit more work (effectively, in the main parsing algorithm, you could empty the candidates array and replace the cached document innerHTML when reaching a node that has `role=main` - though as noted, I'd be interested to see testcases where this is necessary).

As for `getComputedStyle`, I don't think it's used here, right? (You can use `element.style` in checks, fwiw, which will work for inline styles - but in Firefox, we can't rely on other styles being there as the entire document is checked in a worker, where the original DOM is gone and we don't have/apply any of the stylesheets.)

We don't currently have test infrastructure that allows for files that work with `jsdom` but fail with `JSDOMParser` in the way you suggest (ie right now, testcases are just input + output, and both parsers + readability should produce the same output), though I could be persuaded that should be added, it'd be a bit more work.

gijsk:
I'm going to close this out as it's been 18 months. Happy to reopen / consider a new PR that doesn't require `querySelector`.

PR #762, opened 2022-08-30
Simplify script tags removal

willrnch:


gijsk (approved):
The `nodeValue` reset and src removal has been here since the dawn of time. I guess we're hoping it won't break anything to not do that... Let's find out!

PR #770, opened 2022-11-18
Update and pin dependencies

tholu:
I have updated the dependencies to the latest available version and pinned them, which I believe to be a best practice in dependency management.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,15 +20,14 @@
         "sinon": "14.0.2"
       },
       "engines": {
-        "node": ">=16.0.0"
+        "node": ">=14.0.0"
       }
     },
     "node_modules/@babel/code-frame": {
       "version": "7.18.6",
-      "resolved": "https://nexus.symptoma.com/repository/npm-group/@babel/code-frame/-/code-frame-7.18.6.tgz",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
       "integrity": "sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==",
       "dev": true,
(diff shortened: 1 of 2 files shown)


tholu:
Not sure why it failed. There is a 401 error in the Task log. Tests run fine locally.

gijsk:
I'm asking around about the task log failure. I do have a separate curiosity question: why increase the nodejs version requirement, which presumably stops people still using older versions using an up-to-date version of this library? From a quick web search it seems Node 14 is still supported as LTS (though 12 isn't anymore).

tholu:
@gijsk Thanks!

You're right, [14 is still maintained](https://github.com/nodejs/release#release-schedule). I wasn't sure if it works with updated dependencies under 10 anymore and wanted to push people in the direction of a current LTS (which is 18 now). Happy to change it to 14 or keep it at 10 if there are no side-effects!

gijsk:
Matrix for taskcluster suggests that the reason the PR fails tests is:

> gijs: the package-lock.json in the PR references https://nexus.symptoma.com/repository/npm-group/@babel/code-frame/-/code-frame-7.18.6.tgz which gives me 401

@tholu Perhaps this is an internal thing setup in your NPM setup that isn't externally accessible?

gijsk:
Looks like I can't push to your PR but I pastebin'd a diff with just deleting package-lock.json and regenerating it, which produced this: https://paste.mozilla.org/VyeUu6vT

I think just applying this should make things work? But my npm-fu is pretty weak, and I'm traveling so it's possible I'm missing some git/github options that would have made this easier.

gijsk:
(Oh, and I have no idea why that removes the SRI bits and some license bits from package-lock - maybe because I used `--package-lock-only`? It's not clear.)

tholu:
@gijsk That makes sense, the registry in my local config was not the official one, so I fixed that now. Unclear to me why it strips the license bits - my guess is that newer npm versions strip it by default.

gijsk (approved):
Thanks!

PR #778, opened 2023-01-06
Separate regexes

Pandapip1:
This avoids some code duplication and allows for runtime modification of the regexes.


gijsk (changes requested):
Thanks for the PR, and apologies that nobody responded for a few weeks - this got slightly lost over the Christmas / new year break.

I have a few concerns, unfortunately.

First of all, we use this code in Firefox, and have to jump through some hoops to load it because we don't use commonJS. So the additional files would need changes in mozilla-central (the Firefox main repo) to avoid breaking the code entirely. As you can see, neither `Readability.js` nor `Readability-readerable.js` currently have _any_ dependencies (ie there are no `require` calls at all).

I'd be willing to make those but I also confess I don't really understand the value of the change here. My confusion is twofold.

WRT duplication, there are only 2 regexes that are duplicated right now, which I don't think justifies the overhead of another file (which also makes it harder to work out what the code is doing, as you can't look at the regex in-context).

Second, the 1-line comment around why you'd like to see this change mentions changing the regexes. I can see the value of that for the, let's say, algorithmic regexes. But for things like `normalize` or `tokenize` or `whitespace`, I don't think there's much point? It also implicitly significantly expands the API surface and what we'd need to support. Would there be an expectation that readability warns you if you do things to the regexes that break readability (e.g. adding `g` flags to them) ?


If the main motivation of the change is to allow configuring the algorithm regexes, I think I'd prefer a PR that added the specific regexes as options to the constructor, allowing overriding them that way, and clearly documenting how they are used (the documentation you added is a good start but is a little vague, and talks about them impacting how readability evalutes "the webpage", but in most cases really it's the individual elements, not the page as a whole...).

Again, apologies this took such a long time and hopefully this is still helpful and/or something you're interested in.

Pandapip1:
How about I move the regexes int Readability, and have Readability-readerable use the regexes in Readability?

gijsk:
> How about I move the regexes int Readability, and have Readability-readerable use the regexes in Readability?

That'd still have a `require` call in the `readerable` file right?

The reason they're split is because a boolean yes/no decision can be a lot more lightweight (from a performance perspective) than running all of readability. Firefox runs the former on ~every page load to assess whether to show the reader mode icon. The full readability only runs when the user actually opens reader mode.

I guess because Firefox doesn't support `require` anyway, we could try to work around this on our end. But it does feel fundamentally a bit odd...

I'm still curious about your motivation as noted in my previous comment - what's prompted you to suggest this change?

Pandapip1:
> I'm still curious about your motivation as noted in my previous comment - what's prompted you to suggest this change?

I'd really like to be able to tweak the settings without having to fork readability - in particular, disabling the next / previous article links.

gijsk:
(Apologies, my notification email for this got lost)

> I'd really like to be able to tweak the settings without having to fork readability - in particular, disabling the next / previous article links.

OK. For that purpose, I'd really suggest just extending the options object to take replacement regexes that readability then uses instead of the default ones. I'd be happy to take a patch for that. AIUI that won't need any changes to readability-readerable or factoring out of the regexes.

Pandapip1:
I'll make a new PR for that, then.

PR #785, opened 2023-02-23
added an official link about Reader View for context

todrobbins:
Hi there! I thought this might be helpful for future readers of the `README`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Readability.js
 
-A standalone version of the readability library used for Firefox Reader View.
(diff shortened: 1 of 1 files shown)


gijsk (approved):
Thanks! This somehow got lost for a bit. I'm not really sure why. Anyway, I'm taking the patch, but dropped the `en-US` part of the link so people will get the support.m.o page in their own language. :-)

todrobbins:
Ah! Good catch. I'll note that for the future.

PR #794, opened 2023-04-15
fix: delete needless this._debug

sdta25196:
this.log is not executed when _debug is false, so this.log does not need to use this

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1242,12 +1242,10 @@ Readability.prototype = {
         }
       }
 
-      if (this._debug)
-        this.log("Article content pre-prep: " + articleContent.innerHTML);
+      this.log("Article content pre-prep: " + articleContent.innerHTML);
       // So we have all of the content that we need. Now we clean it up for presentation.
       this._prepArticle(articleContent);
-      if (this._debug)
-        this.log("Article content post-prep: " + articleContent.innerHTML);
+      this.log("Article content post-prep: " + articleContent.innerHTML);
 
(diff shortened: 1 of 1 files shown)


gijsk:
On most DOM implementations, asking for `.innerHTML` on a node with a lot of content in it is expensive, so these explicit additional guards are there so we don't waste time doing that if we're not actually going to use the output. It makes running the code faster on complicated DOMs. So unfortunately, I don't think we want to remove this.

sdta25196:
Ok, you're right. Also, I notice that only English commas are used in the scoring of nodes. 
So, should I change it to more punctuation?  After all, punctuation is not used anywhere but in the text.

gijsk:
> Also, I notice that only English commas are used in the scoring of nodes. So, should I change it to more punctuation? After all, punctuation is not used anywhere but in the text.

Yes! I'd be happy to take a separate/new pull request to change https://github.com/mozilla/readability/blob/830d8ab21e7489f4dff44df285f6c16cb6c5db8c/Readability.js#L1033 to use a regular expression that understands more types of commas. It looks like the [wikipedia article for comma has a decent list of which characters we should consider commas](https://en.wikipedia.org/wiki/Comma#Comma_variants).

sdta25196:
I submitted a new PR ([Added comma regular expressions](https://github.com/mozilla/readability/pull/796)), Added commas for German, Hawaiian, Chinese and other languages and for vertical writing.

PR #796, opened 2023-04-24
feat: Added comma regular expressions

sdta25196:
Added commas for German, Hawaiian, Chinese and other languages and for vertical writing


gijsk (approved):
Thanks! I've fixed the linting issue (missing space) and dropped the ʻOkina character which isn't really a comma. I'll update the wikipedia page. I think there are more commas we could add support for, but this is an improvement so let's start here.

gijsk:
Looks like this still breaks the 'qq' test. :-(

I'll try to find some time to work out why and/or if the result is better/worse.

sdta25196:
Well, look forward to your merger.

PR #812, opened 2023-10-01
Update lang field in test expected metadata files

migueldoblado:
This PR adds the lang field to all the `test/test-pages/*/expected-metadata.json` files test cases.

This is done in order to avoid mixing fields in the regeneration of test-cases for future features.


gijsk (approved):
Thanks!

PR #833, opened 2024-01-12
fix: jsonld parse is ignored when context url include the trailing slash

tianzhich:
Some articles use the jsonld in which their `@context` is `http://schema.org/`. It will be ignored by the current regular expression test. Then we can't get the correct metadata.

This PR fix this issue.


.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules
 .sw?
 *.jar
 *.xpi
+.vscode
```
fchasen, line 11:
```suggestion
```
I would recommend adding this to your `~/.gitignore` instead.

CHANGELOG.md
```diff
@@ -50,9 +54,7 @@ reasonable).
 
 ### Added
 
-- `isProbablyReaderable` [can now take an optional options object](https://github.com/mozilla/readability/pull/634) to configure it,
-allowing you to specify the minimum content length, minimum score, and how to
-check if nodes are visible.
+- `isProbablyReaderable` [can now take an optional options object](https://github.com/mozilla/readability/pull/634) to configure it, allowing you to specify the minimum content length, minimum score, and how to check if nodes are visible.
```
fchasen, line 57: Can you revert this? We'd prefer the line to wrap.

fchasen (changes requested):
Thanks for these changes and the test, just some small nits to fix.

tianzhich:
> Thanks for these changes and the test, just some small nits to fix.

Hi Fred, Thanks for your time. Now it would be ok.

fchasen (approved):

PR #857, opened 2024-04-04
Update taskcluster badge path to the correct branch

lotas:
Due to a bug in taskcluster code, those requests were failing instead of returning some valid error response.

Updated branch name


fchasen (approved):

fchasen:
Thanks for the fix!

Issue #775, opened 2022-12-19
`<table>` tag can be mistakenly trimmed


mnaoumov:
Consider https://en.wikipedia.org/wiki/List_of_films_featuring_time_loops

Running `Readability` captures `<thead>` but skips `<table>` making the markup invalid

pdeblanc:
Not sure if this helps, but I encountered a similar issue viewing some pages on ctext.org, and I trimmed one of them down to a "minimal" example:

```
import { Readability } from "@mozilla/readability"
import { JSDOM } from "jsdom"

const url = "https://ctext.org/hongloumeng/ch1/zh"
const html = `
<html lang="zh-TW">
  <body>
    <table>
      <tbody>
        <tr>
          <td><a href="hongloumeng/ch1/zh#n103347">1</a></td>
          <td>
            此開卷第一回也。作者自云曾歷過一番夢幻之後，故將真事隱去，而借「通靈」說此《石頭記》一書也，故曰「甄士隱」云云。但書中所記何事何人？自己又云：今風塵碌碌，一事無成，忽念及當日所有之女子，一一細考較去，覺其行止見識皆出我之上，我堂堂鬚眉，誠不若彼裙釵。我實愧則有餘，悔又無益，大無可如何之日也！當此日，欲將已往所賴天恩祖德錦衣紈袴之時，飫甘饜肥之日，背父兄教育之恩，負師友規訓之德，以致今日一技無成，半生潦倒之罪，編述一集，以告天下。知我之負罪固多，然閨閣中歷歷有人，萬不可因我之不肖自護己短，一并使其泯滅也。所以蓬牖茅椽，繩床瓦灶，並不足妨我襟懷。況那晨風夕月，階柳庭花，更覺得潤人筆墨。我雖不學無文，又何妨用假語村言敷衍出來，亦可使閨閣昭傳，復可破一時之悶，醒同人之目，不亦宜乎？故曰「賈雨村」云云。更於篇中間用「夢」「幻」等字，卻是此書本旨，兼寓提醒閱者之意。
          </td>
        </tr>
        <tr>
          <td><a href="hongloumeng/ch1/zh#n103348">2</a></td>
          <td>看官！你道此書從何而起？說來雖近荒唐，細玩頗有趣味。</td>
        </tr>
        <tr>
          <td><a href="hongloumeng/ch1/zh#n103349">3</a></td>
          <td>
[...]

PR #858, opened 2024-04-05
make sure elements are not deleted if they contain a data table

WetHat:
This addresses issue #775 where a data table got removed because the owning element was considered removable.


gijsk (approved):
Thanks, this looks great! Apologies it took a long time to get to the review.

PR #866, opened 2024-05-04
chore: update expected metadata with null published times

inhumantsar:
While working on a separate issue, I ran `node generate-testcase.js all` to make debugging a little easier and noticed many testcases hadn't been updated to include `publishedTime` when it would result in a `null` value.

This PR adds those `null` values and apparently removes a newline from the `visibility-hidden` test case.


fchasen (approved):
Thanks so much for adding that to so many tests.

PR #867, opened 2024-05-04
fix: ensure short strings of legitimate content are not excluded

### LINKED ISSUE #861: Short sentence standalone paragraphs missing from reader view of https://www.royalroad.com/fiction/63759/super-supportive/chapter/1449598/one-hundred-two-what-kind-of-wordchain
As described in https://bugzilla.mozilla.org/show_bug.cgi?id=1890310, this page contains a number of very short paragraphs (one or a few words) that are lost in reader view, even though most of the content is correctly displayed.

It's worth noting that this website's design is to wrap each paragraph's content in `<div><div><span>`, with inline styles applied to all three elements, rather than just putting the content in sibling `<p>` elements. So the page structure looks like a few hundred blocks like this:

```
<div style="margin: 1em 0; caret-color: #000000">
  <div style="direction: ltr">
    <span style="font-family: arial, sans-serif">Some content here</span>
  </div>
</div>
```

Some examples of missed content:

* Alden stared at it.
* “Let me think about it.”
* “Gotcha.”

Slightly longer sentences not dropped from reader view:

* “I was thinking the same thing.”
* “That’s very efficient,” said Alden.

inhumantsar:
This PR changes one of the `haveToRemove` checks to allow for short paragraphs, such as the written dialog in the linked issue, by adding `linkDensity` to the primary "short content" check. The rationale being that short strings which don't contain any links are likely to be text the user would want to read, provided that the initial preprocessing has already removed the bulk of the page elements.

I regenerated all test cases to check for regressions and added a couple of new checks to deal with those:

* `adWords` and `loadingWords` regexes to help identify ad blocks and loading indicators.
* `textDensity` + image count to exclude elements without useful content.

Test case changes:

* `citylab-1` now includes the published time in the Readability content.
* `ehow-1` lost an extraneous "Other People Are Reading" header but gained an extraneous "Found This Helpful" header
* `ehow-2` lost its "Other People Are Reading" header and gained the word "Save" previously used for a button
* `engadget` now shows the base price and score originally present in the product review
* `firefox-nightly-blog` no longer shows "Leave a Reply"
* `mercurial` now correctly shows a previously excluded code snippets and commands
* `qq` now shows the page header with published date
* `toc-missing` now *incorrectly* shows "Interactive Editor"
* `wikipedia` now shows image captions wrapped in `<div><p>...</p></div>`

The changes which are definitely regressions seem like an acceptable trade for the improvements gained. I would like some input on whether the `adWords` and `loadingWords` regexes are acceptable though. They are scoped so that they will only produce a match against a node's entire `innerText` string, so it's unlikely to impact real content, but it still feels like a slippery slope.

Closes #861

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -160,7 +160,7 @@ Readability.prototype = {
   PRESENTATIONAL_ATTRIBUTES: [ "align", "background", "bgcolor", "border", "cellpadding", "cellspacing", "frame", "hspace", "rules", "style", "valign", "vspace" ],
 
   DEPRECATED_SIZE_ATTRIBUTE_ELEMS: [ "TABLE", "TH", "TD", "HR", "PRE" ],
-  
+
   // The commented out elements qualify as phrasing content but tend to be
   // removed by readability when put into paragraphs, so we ignore them here.
   PHRASING_ELEMS: [
--- a/Readability.js
+++ b/Readability.js
@@ -148,18 +148,8 @@ Readability.prototype = {
     jsonLdArticleTypes: /^Article|AdvertiserContentArticle|NewsArticle|AnalysisNewsArticle|AskPublicNewsArticle|BackgroundNewsArticle|OpinionNewsArticle|ReportageNewsArticle|ReviewNewsArticle|Report|SatiricalArticle|ScholarlyArticle|MedicalScholarlyArticle|SocialMediaPosting|BlogPosting|LiveBlogPosting|DiscussionForumPosting|TechArticle|APIReference$/,
     // used to see if a node's content matches words commonly used for ad blocks or loading indicators
     adWords: /^(ad(vertising|vertisement)?|pub(licité)?|werb(ung)?|广告|Реклама|Anuncio)$/iu,
-    loadingWords: /^((loading|正在加载|Загрузка|chargement|cargando)(\.\.\.)?)$/iu
-  },
-
-  UNLIKELY_ROLES: [ "menu", "menubar", "complementary", "navigation", "alert", "alertdialog", "dialog" ],
-
-  DIV_TO_P_ELEMS: new Set([ "BLOCKQUOTE", "DL", "DIV", "IMG", "OL", "P", "PRE", "TABLE", "UL" ]),
-
-  ALTER_TO_DIV_EXCEPTIONS: ["DIV", "ARTICLE", "SECTION", "P"],
-
-  PRESENTATIONAL_ATTRIBUTES: [ "align", "background", "bgcolor", "border", "cellpadding", "cellspacing", "frame", "hspace", "rules", "style", "valign", "vspace" ],
-
-  DEPRECATED_SIZE_ATTRIBUTE_ELEMS: [ "TABLE", "TH", "TD", "HR", "PRE" ],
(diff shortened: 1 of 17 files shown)


Readability.js
```diff
@@ -145,7 +145,10 @@ Readability.prototype = {
     // see: https://en.wikipedia.org/wiki/Comma#Comma_variants
     commas: /\u002C|\u060C|\uFE50|\uFE10|\uFE11|\u2E41|\u2E34|\u2E32|\uFF0C/g,
     // See: https://schema.org/Article
-    jsonLdArticleTypes: /^Article|AdvertiserContentArticle|NewsArticle|AnalysisNewsArticle|AskPublicNewsArticle|BackgroundNewsArticle|OpinionNewsArticle|ReportageNewsArticle|ReviewNewsArticle|Report|SatiricalArticle|ScholarlyArticle|MedicalScholarlyArticle|SocialMediaPosting|BlogPosting|LiveBlogPosting|DiscussionForumPosting|TechArticle|APIReference$/
+    jsonLdArticleTypes: /^Article|AdvertiserContentArticle|NewsArticle|AnalysisNewsArticle|AskPublicNewsArticle|BackgroundNewsArticle|OpinionNewsArticle|ReportageNewsArticle|ReviewNewsArticle|Report|SatiricalArticle|ScholarlyArticle|MedicalScholarlyArticle|SocialMediaPosting|BlogPosting|LiveBlogPosting|DiscussionForumPosting|TechArticle|APIReference$/,
+    // used to see if a node's content matches words commonly used for ad blocks or loading indicators
+    adWords: /^(ad(vertising|vertisement)?|pub(licité)?|werb(ung)?|广告|Реклама|Anuncio)$/iu,
+    loadingWords: /^((loading|正在加载|Загрузка|chargement|cargando)(\.\.\.)?)$/iu
```
gijsk, line 151: Just one small nitpick here: Could you add the unicode ellipsis to the `(\.\.\.)` bit, and a trailing comma to the line so the next person won't need to update it as you had the `jsonLdArticleTypes`? I should really configure the eslintrc to enforce that...

Readability.js
```diff
@@ -2138,17 +2141,57 @@ Readability.prototype = {
...
-          (!isList && headingDensity < 0.9 && contentLength < 25 && (img === 0 || img > 2) && !this._hasAncestorTag(node, "figure")) ||
-          (!isList && weight < 25 && linkDensity > 0.2) ||
-          (weight >= 25 && linkDensity > 0.5) ||
-          ((embedCount === 1 && contentLength < 75) || embedCount > 1);
+        var textishTags = ["SPAN", "LI", "TD"].concat(Array.from(this.DIV_TO_P_ELEMS));
+        var textDensity = this._getTextDensity(node, textishTags);
+        var isFigureChild = this._hasAncestorTag(node, "figure");
+
+        // apply shadiness checks, then check for exceptions
+        const shouldRemoveNode = () => {
```
gijsk, line 2158: Thanks for refactoring this and adding some error logging to it!

gijsk (changes requested):
Thanks! Just one nit, but I think this otherwise looks good. We really appreciated the thorough evaluation. Like you, we feel it's a balancing act in terms of the impact, but we think it's generally a positive change so let's roll with it. :-)

gijsk (approved):
Thank you!

PR #871, opened 2024-05-20
WIP: add citation, prism, and dc metadata

inhumantsar:


AramZS:
`dc:` is Dublin Core and it is very widely used in academic circles. I think this would be a great addition!

PR #874, opened 2024-05-30
Allow link density value to be modified

vankov1:
A link density modifier is added, this is useful for extracting content with high link density.


fchasen:
Thanks @vankov1, looks great but could you add a bit of documentation about this new option to the [Readme API Reference](https://github.com/mozilla/readability/blob/5abeedd6f54170d94322f06581acdd6260f29568/README.md?plain=1#L29)?

vankov1:
@fchasen Absolutely, I'll update the PR. Thank you for your time checking this.

gijsk (approved):

PR #876, opened 2024-06-06
Add S and U phrasing elements

vankov1:
**S** and **U** should be part of the phrasing elements list:
https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content

Leaving them out causes `<p>` elements to be placed within `<u>` or `<s>` like in the _hukumusume_ test case, which is invalid.


inhumantsar:
@vankov1 you can run `node generate-testcase.js hukumusume` to regenerate `expected.html` and `expected-metadata.json`

PR #883, opened 2024-06-26
Drop polyfill.io reference out of abundance of caution (fixes #882)

### LINKED ISSUE #882: Remove polyfill.io from test code
Due to https://sansec.io/research/polyfill-supply-chain-attack

gijsk:


tantek (changes requested):
I think it's better to remove it entirely, to both avoid leaving behind a breadcrumb that may lead someone to trying to retrieve the "original" at polyfill io without knowing the context of the compromise, and avoid a synthetic snippet of code that wasn't present in the original. Didn't see this PR before I filed a "simple" removal PR #884

cmkm (approved):
Thanks!

PR #885, opened 2024-06-26
Fix missing header from GitLab blog page

revolter:


fchasen:
Thanks for submitting the test case and fix, the simpler fix for this seem to be to just remove `tool` from the regex as that doesn't appear to break any other test cases.

Readability.js
```diff
@@ -127,7 +127,7 @@ Readability.prototype = {
     okMaybeItsACandidate: /and|article|body|column|content|main|shadow/i,
 
     positive: /article|body|content|entry|hentry|h-entry|main|page|pagination|post|text|blog|story/i,
-    negative: /-ad-|hidden|^hid$| hid$| hid |^hid |banner|combx|comment|com-|contact|foot|footer|footnote|gdpr|masthead|media|meta|outbrain|promo|related|scroll|share|shoutbox|sidebar|skyscraper|sponsor|shopping|tags|tool|widget/i,
+    negative: /-ad-|hidden|^hid$| hid$| hid |^hid |banner|combx|comment|com-|contact|foot|footer|footnote|gdpr|masthead|media|meta|outbrain|promo|related|scroll|share|shoutbox|sidebar|skyscraper|sponsor|shopping|tags|tool(?!chain)|widget/i,
```
fchasen, line 130:
```suggestion
    negative: /-ad-|hidden|^hid$| hid$| hid |^hid |banner|combx|comment|com-|contact|foot|footer|footnote|gdpr|masthead|media|meta|outbrain|promo|related|scroll|share|shoutbox|sidebar|skyscraper|sponsor|shopping|tags|widget/i,
```

revolter:
Wasn't it added there with a reason? Maybe there is a test case missing for that part of the regex? 🤔

revolter:
Oh, it's there since [the initial commit](https://github.com/mozilla/readability/commit/55587d91ac5b333ca6bf968e5504f3083c59cf6c#diff-2e165ea528afeb3f92b2bed19888c901d730c559afbbcd821902506eb7b7f9a2R76).

So, should I remove `tool` completely?

fchasen:
Yep, seems to be fine to remove completely.

fchasen (changes requested):

fchasen (approved):

fchasen:
Thanks again for the text case that has most of the text of the header in the id, that pattern doesn't work great with regex matching (for instance if the header was about `Meta`). We may look into this more soon depending on how common that is.

revolter:
Good job with the contribution instructions 👏🏻 Was thinking about creating an issue at first 😅

PR #889, opened 2024-07-01
Fix cloning attributes to use setAttributeNode to fix #859

### LINKED ISSUE #859: Problem handling invalid HTML attributes
I recently started using Readability.js for a reader view that I work on. One user reported an error with [this website](https://www.spiegel.de/kultur/tv/gangster-serie-crooks-auf-netflix-mit-frederick-lau-belmondo-mit-schmaeh-a-def5d211-addc-4082-b439-74eb4224cdf8?giftToken=bfd96d6a-2b01-413d-915f-d03497e53d50) (note: it's paywalled so only a limited number of people can view it). The error shown in the console is:

```
Error handling response: Error: Failed to execute 'setAttribute' on 'Element': '@click' is not a valid attribute name.
    at Readability._simplifyNestedElements
```

This points to this part of Readability.js:

```js
} else if (
  this._hasSingleTagInsideElement(node, "DIV") ||
  this._hasSingleTagInsideElement(node, "SECTION")
) {
  var child = node.children[0];
    for (var i = 0; i < node.attributes.length; i++) {
      child.setAttribute(
        node.attributes[i].name,
        node.attributes[i].value
      );
    }
```

I think `@click` as an attribute [is a Vue.js thing](https://stackoverflow.com/q/45369553/2065702).

Regardless, can Readability.js be updated to properly handle (probably just ignore) invalid HTML attributes instead of breaking? I can see validation of attribute names as a way forward or try/catching it and ignoring bits that error.

gijsk:


test/test-pages/virginia-edu-broken-attributes/expected.html
```diff
@@ -0,0 +1,12 @@
+<div id="readability-page-1" class="page">
```
gijsk, line 1: This isn't actually very good, but it addresses the invalid markup breaking us entirely.

gijsk:
I don't love this - https://github.com/mozilla/readability/pull/509 fixed a similar issue by just using a try...catch; The other code that is now hitting the same problem doesn't have a try...catch. I could be convinced that just adding that try...catch is better than using `setAttributeNode`, especially given the context from #509 and the responses that those tickets on `jsdom` and the DOM spec have received ("don't do that"), and the fact that the attributes aren't actually used so we might as well lose them...

fchasen:
@gijsk This has some conflicts now, but otherwise seems ready to merge.

gijsk:
Let's close this in favour of #918

PR #890, opened 2024-07-01
Eslint/prettier improvements

gijsk:
This switches to the mozilla recommended eslint settings and prettier output.

Hopefully this makes the code style bits a bit easier to deal with.

I took the opportunity to switch to flat config as well.


cmkm (approved):
thank you!

PR #892, opened 2024-07-05
Avoid unnecessary copying while unwrapping single-image noscripts

JKingweb:
This patch reorders two operations in the `_unwrapNoscriptImages` function to avoid making a clone of the `<noscript>` if it is not a single-image wrapper. I expect the difference is minuscule, but it is the order in which things should be done, as far as I can tell.


gijsk (approved):
Thank you!

PR #893, opened 2024-07-05
Make isSingleImage function a loop rather than recursive

JKingweb:
My motivation for this change was to avoid any potential for a stack overflow (though I'm not sure how likely this is in practice or if it's even possible in JavaScript), but it turns out to reduce the time the test suite takes to run on my computer from 20s to 18s, so there does seem to be a practical benefit.

The return at the bottom was added to appease eslint; the conditions in the loop body make it unreachable.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1867,7 +1867,8 @@ Readability.prototype = {
     while (node) {
       if (node.tagName === "IMG") {
         return true;
-      } else if (node.children.length !== 1 || node.textContent.trim() !== "") {
+      }
+      if (node.children.length !== 1 || node.textContent.trim() !== "") {
         return false;
       }
       node = node.children[0];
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -1864,15 +1864,15 @@ Readability.prototype = {
...
-    if (node.tagName === "IMG") {
-      return true;
-    }
-
-    if (node.children.length !== 1 || node.textContent.trim() !== "") {
-      return false;
+    while (node) {
+      if (node.tagName === "IMG") {
+        return true;
+      } else if (node.children.length !== 1 || node.textContent.trim() !== "") {
```
gijsk, line 1870: I'd prefer no else after a return (as it was before), if that's OK? :-)

gijsk (commented):
Thanks for this simplification - would you mind keeping the "no else after return" practice that was there before?

gijsk (approved):
Err, I guess I can just edit that in - thanks!

PR #894, opened 2024-07-06
Remove _concatNodeLists function

JKingweb:
The `_getAllNodesWithTag` function performs the task just as well, and is used throughout Readability, whereas `_concatNodeLists` is only used once.


gijsk (approved):
Thank you!

PR #896, opened 2024-07-16
Improve character reference unescaping

JKingweb:
This improves the entity unescaping such that:

1. Null characters are no longer emitted
2. Characters beyond four digits are handled properly

This is only an improvement, not a complete fix. Two exceptions to correctness remain:
1. Most named entities are not converted (a documented limitation)
2. Character references to C1 controls are not mapped to windows-1252 chracters

The best way to really do this correctly is to write to a temporary element's `innerHTML` (after escaping any `<`) and then read back its `textContent`. However this would require enhancing JSDOMParser such that it handles all entities and character references correctly, so that all tests may pass consistently. If there's any appetite for this I'd be happy to do the work.

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -60,26 +60,23 @@
       .replace(/&(quot|amp|apos|lt|gt);/g, function (match, tag) {
         return entityTable[tag];
       })
-      .replace(
-        /&#(?:x([0-9a-f]+)|([0-9]+));/gi,
-        function (_, hex, numStr) {
-          var num = parseInt(hex || numStr, hex ? 16 : 10);
-          
-          // these character references are replaced by a conforming HTML parser
-          if (num == 0 || num > 0x10FFFF || (num >= 0xD800 && num <= 0xDFFF)) {
-            num = 0xFFFD;
-          }
-          
-          // code points beyond the BMP must be converted to a surrogate pair when using String.fromCharCode
-          if (num > 0xFFFF) {
-            var high = 0xD800 + (((num - 0x10000) & 0xFFC00) >> 10);
-            var low = 0xDC00 + ((num - 0x10000) & 0x3FF);
-            return String.fromCharCode(high) + String.fromCharCode(low);
-          }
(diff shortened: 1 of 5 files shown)


gijsk:
Thank you! Before I dive into code, I'm curious, do you have an example of website brokenness (and/or existing issues in this repo) that this addresses?

JKingweb:
I'm porting Readability to PHP as an exercise, so I'm just looking at everything as I go. The lack of handling for named entities does affect one existing test (`msn`) where the substring `&nbsp;` is output as part of the excerpt, but this patch as it stands just corrects edge cases in the abstract. It's conceivable that emoji are being mangled with the old logic, but I've not used Readability much yet, so I can't point to a concrete example.

JSDOMParser.js
```diff
@@ -60,13 +60,23 @@
...
+        if (num == 0 || num > 0x10ffff || (num >= 0xd800 && num <= 0xdfff)) {
+          num = 0xfffd;
+        }
+
+        // code points beyond the BMP must be converted to a surrogate pair when using String.fromCharCode
+        if (num > 0xffff) {
+          var high = 0xd800 + (((num - 0x10000) & 0xffc00) >> 10);
+          var low = 0xdc00 + ((num - 0x10000) & 0x3ff);
+          return String.fromCharCode(high) + String.fromCharCode(low);
         }
```
gijsk, line 76: Could we use `String.fromCodePoint` directly here? It looks like that is supported from NodeJS 4.0 and we require a higher nodeJS version so we should be OK. That would save me (and any future readers) from having to reason about the hex code. 😅

JKingweb: Yes, `fromCodePoint` would obviate a nice chunk of code. I was unaware of the baseline and a lot of code presumably predates the policy, so I jump through the hoops. 

I'll make the change when I have a minute.

gijsk: Ah, great - I may also be able to do it directly - I didn't want to do so without asking just in case there was a reason not to that I was missing.

JKingweb: Done.

gijsk (approved):
Awesome, thank you!

PR #899, opened 2024-07-28
Avoid repetition when saving attempts during article grabbing

JKingweb:
All four branches of a conditional section (including the `else` branch) invoke `this._attempts.push` with the same argument. This patch simply avoids the repetition by performing the push to the array before the branching.


fchasen (approved):
Thanks, seems like a nice clean up!

PR #902, opened 2024-08-13
Fix `this` context issue to resolve undefined `this.REGEXPS`

dispatchqueue:


gijsk (approved):
Thanks! It would be nice to have a testcase that exercised this code (so we don't break it again) but either way this is an improvement. :-)

PR #905, opened 2024-08-22
Clean up byline validation

JKingweb:
The `_checkByline` function is a little weird in a few ways:

- Despite its name and return type it does more than check
- It relies on external information to short-curcuit itself
- It calls another function to do more checks
- It checks whether the node is an element (which will always be true) before getting attributes, but unconditionally relies on variables it has set conditionally

It gets the job done, but I found it confusing to read and make sense of considering its modest length. I have consequently cleaned it up and re-organized the logic surrounding byline harvesting in general.

I included some basic documentation, but since I've never written a JavaScript doc-string before and I'm uncertain what the `matchString` argument is actually for, suggestions for improvement there are welcome.

For reference, one can assume that `node` is an element becaause the `_grabArticle` loop begins at `documentElement`, and the `_getNextNode` function which drives the loop always returns an element if it returns a node.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -988,14 +988,14 @@ Readability.prototype = {
   _isValidByline(node, matchString) {
     var rel = node.getAttribute("rel");
     var itemprop = node.getAttribute("itemprop");
-    var byline = node.textContent.trim();
+    var bylineLength = node.textContent.trim().length;
 
     return (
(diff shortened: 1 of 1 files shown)


gijsk (approved):
Thanks and sorry for the wait!

PR #906, opened 2024-08-23
Use array reduction when cleaning paragraphs

JKingweb:
The current code for cleaning non-content paragraphs is straightforward, but brittle against change: one must remember to modifiy both the list of element counts as well as `totalCount`. This change removes the need to modify `totalCount` (or indeed to set any more variables) by instead having a list of tag names `reduce()`d into `totalCount` in one operation. It's slightly less readable, but it seemed worth it to me.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -832,15 +832,18 @@ Readability.prototype = {
     );
 
     // Remove extra paragraphs
+    // At this point, nasty iframes have been removed; only embedded video
+    // ones remain.
+    var contentTags = ["img", "embed", "object", "iframe"];
     this._removeNodes(
       this._getAllNodesWithTag(articleContent, ["p"]),
       function (paragraph) {
-        var imgCount = paragraph.getElementsByTagName("img").length;
-        var embedCount = paragraph.getElementsByTagName("embed").length;
-        var objectCount = paragraph.getElementsByTagName("object").length;
-        // At this point, nasty iframes have been removed, only remain embedded video ones.
-        var iframeCount = paragraph.getElementsByTagName("iframe").length;
-        var totalCount = imgCount + embedCount + objectCount + iframeCount;
+        // We use an array reduction here so the counts of elements are summed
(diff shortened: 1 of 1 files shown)


Readability.js
```diff
@@ -832,15 +832,18 @@ Readability.prototype = {
...
-        var objectCount = paragraph.getElementsByTagName("object").length;
-        // At this point, nasty iframes have been removed, only remain embedded video ones.
-        var iframeCount = paragraph.getElementsByTagName("iframe").length;
-        var totalCount = imgCount + embedCount + objectCount + iframeCount;
+        // We use an array reduction here so the counts of elements are summed
+        // without anyone having to make further code edits if the list of
+        // content tags is changed.
+        var totalCount = contentTags.reduce(function (total, tag) {
+          return total + paragraph.getElementsByTagName(tag).length;
+        }, 0);
```
gijsk, line 846: So I'm a little lost. Any reason not to use `this._getAllNodesWithTag(paragraph, contentTags).length` ? That would seem easier than hand-rolling the reduce.

The other option, for reasons of speed, would be to iterate over `contentTags` (without a nested function) and to return `false` as soon as we find any of the content tags (as the `return` below will be `false` as soon as we find any of the elements in question). It might also make sense to put the `_getInnerText` call first in that case.

Both of those would still have the same benefit in terms of not requiring manual copy-pasting for each kind of inner tag that we're looking for.

gijsk (changes requested):
It's a good thing to identify that this code is a little repetitive. However, I think we can juice this particular fruit a little more - I left a comment inline.

Readability.js
JKingweb: I see what you mean: I fixed one problem of weird methodology in the original, but not the other. I'll iterate further when I find some time.

gijsk: @JKingweb any chance you had a moment to look at this again? :-)

gijsk (approved):
Thanks!

Issue #913, opened 2024-10-11
<ol> tag is dropped from an ordered list


beala:
Sometimes the `<ol>` element of an ordered list is dropped from the `content` output.

Here is a repro:

```js
test('JSDOM and Readability', () => {
    const html = `
<html>
    <body>
        <ol>
            <li><p>AI hasn’t meaningfully changed anything in cybersecurity so far. Deep fake phishing is still rare, L</p></li>
        </ol>
    </body>
</html>`
    const dom: JSDOM = new JSDOM(html)
    const extracted = new Readability(dom.window.document, {
        debug: true
    }).parse()
    
    expect(extracted?.content).toContain("<ol>")
})
```

Here is the failure:
```
  ● JSDOM and Readability

    expect(received).toContain(expected) // indexOf

    Expected substring: "<ol>"
    Received string:    "<div id=\"readability-page-1\" class=\"page\"><div>
                <li><p>AI hasn’t meaningfully changed anything in cybersecurity so far. Deep fake phishing is still rare, L</p></li>
            </div></div>"

      82 |     }).parse()
      83 |     
    > 84 |     expect(extracted?.content).toContain("<ol>")
         |                                ^
      85 | })

      at Object.<anonymous> (__tests__/email/emailParser.test.ts:84:32)
```

This is using
```
"dependencies": {
    "jsdom": "^24.0.0",
    "@mozilla/readability": "^0.5.0",
}
```

Here is the debug output:
```
  console.log
    Reader: (Readability) **** grabArticle ****

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Starting grabArticle loop

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Candidate: <li > with score 1

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Candidate: <ol > with score -1

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Candidate: <body > with score 0.6666666666666666

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Looking at sibling node: <ol > with score -1

      at Readability.log (node_modules/@mozilla/readability/Readability.js:84:21)

  console.log
    Reader: (Readability) Sibling has score -1

[...]

PR #914, opened 2024-10-11
Fix "<ol> tag is dropped from an ordered list" #913

beala:
Fixes https://github.com/mozilla/readability/issues/913


gijsk (approved):
Awesome, thank you!

PR #915, opened 2024-10-11
Use better test for start of base64 payload in _fixLazyImages

JKingweb:
The prior method didn't take into account the whitespace it matches when adding seven to the offset, nor did it consider that the type might itself end in "base64"; looking for the comma which is necessarily the delimiter between metadata and data is both simpler and more accurate.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -2315,10 +2315,10 @@ Readability.prototype = {
             }
           }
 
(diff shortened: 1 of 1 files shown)


JKingweb:
It occurred to me belatedly that the `parts[0]` variable already contains the data URI metadata, so we don't need to examine the attribute value a second time in the first place.

gijsk (approved):
Thank you!

PR #918, opened 2024-10-19
Fix DomException when simplifying nested elements caused by invalid attribute names

dbowring:
Copies the same fix for copying invalid property names (such as `"`) used in `_setNodeTag` to `_simplifyNestedElements`.

You can reproduce this error running the following code in node (with `jsdom` installed):

```javascript
const { JSDOM } = require("jsdom");
const Readability = require("./Readability.js");

const sampleHTML = '<div "=""><div><div>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</div></div></div>';
const dom = new JSDOM(sampleHTML, { contentType: "text/html" });
const readability = new Readability(dom.window.document);
const parsed = readability.parse();
console.log(parsed.textContent);
```

Without the fix:

```console
$ node demo.js  

./readability/node_modules/jsdom/lib/jsdom/living/helpers/validate-names.js:10
    throw DOMException.create(globalObject, [`"${name}" did not match the Name production`, "InvalidCharacterError"]);
    ^
[DOMException [InvalidCharacterError]: """ did not match the Name production]

Node.js v20.18.0
```

With the fix:

```console
$ node demo.js
Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```

### DIFF
--- a/JSDOMParser.js
+++ b/JSDOMParser.js
@@ -542,6 +542,10 @@
     getEncodedValue() {
       return encodeHTML(this._value);
     },
+    // Cheat horribly. This is fine for our usecases.
+    cloneNode() {
+      return this;
+    },
   };
 
   var Comment = function () {
@@ -821,6 +825,10 @@
         }
       }
       this.attributes.push(new Attribute(name, value));
+    },
+
+    setAttributeNode(node) {
+      this.setAttribute(node.name, node.value);
     },
 
     removeAttribute(name) {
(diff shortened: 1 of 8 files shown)


dbowring:
Added the demo from the original comment as a test (402e5de606312e1c8aeeef0710546acf5527e458), but it also requires changes the way attributes are compared for the tests (to ignore invalid attributes, if they are there they are fine, but if we need to remove them that should also be fine), so you may wish to drop that commit if the test changes are not acceptable.

gijsk (approved):
I had put up #889 for this but then neglected to update it when it became outdated/conflicted. I've unfortunately not been as active as I'd like due to some health issues so I'm sorry it's taken so long. Anyway, I've sort of merged the two pulls and will approve and merge this into the repo and then close my own older PR.

PR #940, opened 2025-01-01
Handle schema.org context objects. Fixes #934

### LINKED ISSUE #934: `_getJSONLD` doesn't handle context objects
For example this json-ld from https://www.nbcnews.com/news/world/south-korea-martial-law-president-yoon-suk-yeol-impeachment-vote-rcna183138 fails with `Reader: (Readability) parsed.@context.match is not a function`:

```
{
  '@context': {
    '@vocab': 'http://schema.org',
    articleId: { '@id': 'Text', '@type': '@id' },
    topics: { '@id': 'Text', '@type': '@id' }
  },
  '@type': 'NewsArticle',
  headline: 'South Korean president’s impeachment fails after ruling party lawmakers walk out'
  ...
}
```

The problematic code is this, which assumes `parsed["@context"]` will always be a string, never an object:

```javascript
          if (
            !parsed["@context"] ||
            !parsed["@context"].match(/^https?\:\/\/schema\.org$/)
          ) {
            return;
          }
```

The quick fix to at least avoid the error would be:

```diff
          if (
            !parsed["@context"] ||
+           !(typeof parsed["@context"] === "string") ||
            !parsed["@context"].match(/^https?\:\/\/schema\.org$/)
          ) {
            return;
          }
```

To support objects in addition to strings:

```diff
           var content = jsonLdElement.textContent.replace(/^\s*<!\[CDATA\[|\]\]>\s*$/g, "");
           var parsed = JSON.parse(content);
-          if (
-            !parsed["@context"] ||
-            !parsed["@context"].match(/^https?\:\/\/schema\.org$/)
-          ) {
+
+          const schemaDotOrgRegex = /^https?\:\/\/schema\.org$/;
+          const matches =
+            (typeof parsed["@context"] === "string" && parsed["@context"].match(schemaDotOrgRegex)) ||
+            (typeof parsed["@context"] === "object" && typeof parsed["@context"]["@vocab"] == "string" && parsed["@context"]["@vocab"].match(schemaDotOrgRegex))
+
+          if (!matches) {
             return;
           }
```

danielnixon:
Fixes https://github.com/mozilla/readability/issues/934


test/test-pages/schema-org-context-object/expected-metadata.json
```diff
@@ -0,0 +1,10 @@
+{
+  "title": "South Korean president’s impeachment fails after ruling party lawmakers walk out",
+  "byline": "Stella Kim, Jennifer Jett",
```
danielnixon, line 3: It turns out this change doesn't affect any existing test cases, so I added one. This is the key line here. Prior to this change, this would contain data taken from the DOM, starting with "By...".

gijsk (approved):
Thanks, this looks great!

PR #944, opened 2025-01-03
Add null | undefined to parse results. Fixes #937

### LINKED ISSUE #937: TypeScript type returned by `parse` is misleading
The return type in https://github.com/mozilla/readability/blob/main/index.d.ts is misleading. It doesn't include null/undefined in the types of values that can definitely be null or undefined. [`byline`](https://github.com/mozilla/readability/blob/main/index.d.ts#L50), for example, can commonly be null/undefined.

This has bitten me a few times while calling Readability from a TypeScript context.

It might be nice if we could update index.d.ts to reflect the types more accurately.

danielnixon:
Fixes https://github.com/mozilla/readability/issues/937.

This PR _might_ be too much of a swing in the opposite direction. The reason I've included null and undefined in all props is because it's difficult to trace exactly which props are always assigned a value versus which might be null versus which might be null _or_ undefined.

I've got a branch locally where I'm slowly annotating everything with types, but that is slow going. In the meantime, erring this way is arguably more appropriate.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,33 +32,33 @@ export class Readability<T = string> {
 
   parse(): null | {
     /** article title */
-    title: string;
+    title: string | null | undefined;
 
     /** HTML string of processed article content */
-    content: T;
+    content: T | null | undefined;
 
     /** text content of the article, with all the HTML tags removed */
-    textContent: string;
+    textContent: string | null | undefined;
 
     /** length of an article, in characters */
-    length: number;
+    length: number | null | undefined;
 
     /** article description, or short excerpt from the content */
-    excerpt: string;
+    excerpt: string | null | undefined;
 
     /** author metadata */
-    byline: string;
+    byline: string | null | undefined;
 
     /** content direction */
-    dir: string;
+    dir: string | null | undefined;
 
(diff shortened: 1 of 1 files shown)


danielnixon:
An alternative option here (slightly more invasive) could be to coalesce everything in the object returned from `parse` with either null _or_ undefined, then only include one or the other in the type definition. Something like:

```ts
return {
      title: this._articleTitle ?? undefined,
      byline: metadata.byline || this._articleByline ?? undefined,
      dir: this._articleDir ?? undefined,
      lang: this._articleLang ?? undefined,
      content: this._serializer(articleContent) ?? undefined,
      textContent: textContent  ?? undefined,
      length: textContent.length  ?? undefined,
      excerpt: metadata.excerpt  ?? undefined,
      siteName: metadata.siteName || this._articleSiteName  ?? undefined,
      publishedTime: metadata.publishedTime  ?? undefined,
    };
```

(or null instead of undefined if that is the preference).

cmkm (approved):
Thank you! If this doesn't work for yourself or other Typescript users in the future we can revisit this with your alternate solution.

PR #952, opened 2025-01-23
feat: Preserve custom element tags

### LINKED ISSUE #951: Preserve mathjax when processing content with readability
The Readability parser currently does not preserve custom element tags such as `<my-custom-tag>` during the parsing process. These custom tags are stripped out, which can result in the loss of important semantic information and custom function

## Desired Feature

Add an option to specify tag names which should be preserved during the parsing.

```js
var article = new Readability(document, {
  tagsToPreserve: ["my-custom-tag", "another-tag"],
}).parse();
```

ras0q:
close #951

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -64,6 +64,7 @@ function Readability(doc, options) {
   this._disableJSONLD = !!options.disableJSONLD;
   this._allowedVideoRegex = options.allowedVideoRegex || this.REGEXPS.videos;
   this._linkDensityModifier = options.linkDensityModifier || 0;
+  this._tagsToPreserve = options.tagsToPreserve || [];
 
   // Start with all flags set
   this._flags =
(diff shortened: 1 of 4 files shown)


ras0q:
@gijsk Could you please review this PR?

Readability.js
```diff
@@ -1156,7 +1157,11 @@ Readability.prototype = {
           continue;
         }
 
-        if (this.DEFAULT_TAGS_TO_SCORE.includes(node.tagName)) {
+        const tagsToScore = [
+          ...this.DEFAULT_TAGS_TO_SCORE,
+          ...this._tagsToPreserve,
+        ];
+        if (tagsToScore.includes(node.tagName)) {
```
gijsk, line 1164: The naming of this is modeled on `classesToPreserve`, which functions as an allowlist for classes that are then not removed from the output (otherwise, all class information is removed).

From that, and from the documentation, and from the ticket you filed ("which can result in the loss of important semantic information and custom function") I would expect this new property to ensure that the content in these custom tags is _always_ preserved. But the patch does not do that, it only scores those elements and they may yet get rejected if their score is too low.

So I think either the patch behaviour needs to be changed (so the tags always stay part of the output, which I suspect is non-trivial to organize), or the documentation/explanation/labeling of the property needs to be changed so that it matches the behaviour. Maybe calling it `additionalTagsToScore` would be better?

Finally, if you keep the current implementation (but perhaps rename the property), if we need to create a list of scorable tag names let's not do it on every iteration inside this loop, as that is expensive and this loop is hot (performance-sensitive) code.

gijsk:
Let's close this in favour of #958.

PR #958, opened 2025-03-02
Preserve MathJax tags

AlexTate:
Related: #951

**Revised 3/4**: Attempt to preserve rendered MathJax. In this PR, `<mjx-container>` tags are no longer identified as unlikely candidates.  This allows child `<math>` tags to be preserved. These `<math>` tags take the place of `<mjx-math>` tags as the primary container of rendered glyphs.


AlexTate:
@gijsk No, that was my mistake. `<mjx-math>` tags are a product of client-side rendering rather than part of the static HTML provided by the server, so they weren't captured during test generation and I didn't check for them in the testcase. However, I misdiagnosed the issue and it really doesn't matter if the `<mjx-container>` and `<mjx-math>` tags are present in Readability outputs (see my updated comment on #951). I'll update the PR description shortly.

I've updated the testcase to include all rendered MathJax tags. To assist with future testcases involving dynamic content, I added support for `file://` URLs in generate-testcase.js. One can save the DOM tree to disk and pass in the resulting file so that it receives the standard treatments. Unfortunately in this testcase, the page was determined to be "unreaderable" after it was processed by `htmltidy2` (though the original is "readerable"), and if `htmltidy2` is skipped then `JSDOMParser` produces a slew of errors. Nonetheless, the expected jsdom* tests run and pass.

gijsk (approved):
Thanks, also for the test helper. I just noticed one small issue in that which I'll fix.

PR #960, opened 2025-03-04
Handle en and em dashes in titles

### LINKED ISSUE #954: Handle en dashes (–) and em dashes (—) in titles
To determine an article's title[^1], the first thing that Readability does is to attempt to split the document's `<title>` by a separator and keep the first part:

https://github.com/mozilla/readability/blob/118f01538e167218bd86ffd493bd3466aec4870a/Readability.js#L595-L598

This is a good heuristic and seems to extract the right title for a variety of documents. However the separators only cover the following regex: `/ [\|\-\\\/>»] /`. This means that the following separator patterns are covered:

- `Title | Site name`
- `Title - Site name`
- `Title \ Site name`
- `Title > Site name`
- `Title » Site name`

## Goal

The heuristic described above leaves out some other relatively common separators. The ones I'm seeing most often are en and em dashes:

- `Title – Site name`
- `Title — Site name`

I'd like to add support for these two separators. This would be a two-line change and wouldn't change any of the existing logic.

Here are example titles that this would fix:[^2]

- `Your weekly Web Directions reading – Web Directions` ([article](https://webdirections.org/blog/your-weekly-web-directions-reading/))
- `Partisan Politics and the Road to Plutocracy – Economics from the Top Down` ([article](https://economicsfromthetopdown.com/2025/01/31/partisan-politics-and-the-road-to-plutocracy/))
- `Pluralistic: Premature Internet Activists (13 Feb 2025) – Pluralistic: Daily links from Cory Doctorow` ([article](https://pluralistic.net/2025/02/13/digital-rights/))[^3]

Can I open a PR?

## Non-goals (unless you want them to be)

- Handling cases where em dashes are used as separators without spaces: `Title—Site name`. This is relatively common ([example 1](https://asteriskmag.com/issues/09/can-we-build-a-five-gigawatt-data-center), [example 2](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs)), but it would require treating an em dash differently from all other separators (a hyphen without spaces around cannot be treated as a separator, it would split hyphenated words). If you don't mind the extra complexity, I'm happy to cover this as well.
- Handling less common separators. For example, I've seen various types of interpuncts being used (`·`, `•`, `∙`, `◦`, etc.), but I don't expect Readability to cover all of these since the list could become endless (though it could consider covering the most common, perhaps `&middot;` and `&bullet;`—I leave this up to you).

[^1]: Unless there is explicit title metadata in the document: https://github.com/mozilla/readability/blob/118f01538e167218bd86ffd493bd3466aec4870a/Readability.js#L1813-L1815
[^2]: These are all using en dashes but I think including em dashes makes sense as well. There may not be many English-language documents using em dashes with spaces around them, but that is a common thing in French for example.
[^3]: Readability currently thinks that the title here is `Daily links from Cory Doctorow`, which is the publication name. This happens because of the semicolons: if Readability doesn't find a separator but finds a semicolon instead, it takes whatever is _after the last_ semicolon.

robinmetral:
Fixes #954.

I'm not sure whether this warrants new test cases. Happy to add some if necessary (with directions).

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -593,23 +593,18 @@ Readability.prototype = {
     }
 
     // If there's a separator in the title, first remove the final part
-    const titleSeparators = /\|\-–—\\\/>»/;
-    if (new RegExp(`\\s[${titleSeparators.source}]\\s`).test(curTitle)) {
+    const titleSeparators = /\|\-–—\\\/>»/.source;
+    if (new RegExp(`\\s[${titleSeparators}]\\s`).test(curTitle)) {
       titleHadHierarchicalSeparators = /\s[\\\/>»]\s/.test(curTitle);
       let allSeparators = Array.from(
-        origTitle.matchAll(
-          new RegExp(`\\s[${titleSeparators.source}]\\s`, "gi")
-        )
+        origTitle.matchAll(new RegExp(`\\s[${titleSeparators}]\\s`, "gi"))
       );
       curTitle = origTitle.substring(0, allSeparators.pop().index);
 
       // If the resulting title is too short, remove the first part instead:
       if (wordCount(curTitle) < 3) {
         curTitle = origTitle.replace(
-          new RegExp(
-            `^[^${titleSeparators.source}]*[${titleSeparators.source}]`,
-            "gi"
-          ),
+          new RegExp(`^[^${titleSeparators}]*[${titleSeparators}]`, "gi"),
           ""
         );
       }
@@ -654,10 +649,7 @@ Readability.prototype = {
       (!titleHadHierarchicalSeparators ||
         curTitleWordCount !=
           wordCount(
-            origTitle.replace(
-              new RegExp(`\\s[${titleSeparators.source}]\\s`, "g"),
-              ""
-            )
+            origTitle.replace(new RegExp(`\\s[${titleSeparators}]\\s`, "g"), "")
           ) -
             1)
     ) {
--- a/Readability.js
+++ b/Readability.js
@@ -593,14 +593,14 @@ Readability.prototype = {
     }
 
     // If there's a separator in the title, first remove the final part
-    if (/ [\|\-\\\/>»] /.test(curTitle)) {
+    if (/ [\|\-\–\—\\\/>»] /.test(curTitle)) {
       titleHadHierarchicalSeparators = / [\\\/>»] /.test(curTitle);
-      let allSeparators = Array.from(origTitle.matchAll(/ [\|\-\\\/>»] /gi));
+      let allSeparators = Array.from(origTitle.matchAll(/ [\|\-\–\—\\\/>»] /gi));
       curTitle = origTitle.substring(0, allSeparators.pop().index);
 
       // If the resulting title is too short, remove the first part instead:
       if (wordCount(curTitle) < 3) {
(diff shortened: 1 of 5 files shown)


Readability.js
```diff
@@ -600,7 +600,7 @@ Readability.prototype = {
 
       // If the resulting title is too short, remove the first part instead:
       if (wordCount(curTitle) < 3) {
-        curTitle = origTitle.replace(/^[^\|\-\\\/>»]*[\|\-\\\/>»]/gi, "");
+        curTitle = origTitle.replace(/^[^\|\-\–\—\\\/>»]*[\|\-\\\/>»]/gi, "");
```
gijsk, line 603: I think we also need/want to update the regexes a few lines above this (the `.test` and `.matchAll` calls), right?

gijsk:
In terms of tests, I think it would be worth having something - especially as I am wondering if/when the change I outlined is sufficient or not. Steps to create a minimal testcase:

1. create a subfolder in `test`, maybe `title-em-dash` or similar
2. create a `source.html` file - you can copy https://github.com/mozilla/readability/blob/main/test/test-pages/title-and-h1-discrepancy/source.html 
3. adjust the content of the `<title>` and drop the `<h1>` in your new file
4. run something like `npm run generate-testcase title-em-dash` (using the directory name you picked in step 1)
5. check the resulting metadata and/or `expected.html` that have been generated for having the correct title information.
6. doublecheck that `npm test` passes successfully (if you need to rerun just your testcase, `npm test -- -g title-em-dash`, again with the directory name from step 1, will run only that test)

Could repeat for both en and em-dash depending on whether you think that makes sense. It doesn't look like we have explicit testing for the title stuff at the moment - when I worked on the changes in https://github.com/mozilla/readability/commit/1c4d63be3e7344c3dfdf76ebb05fd0d32de93eb3 I ended up breaking `archive-of-our-own` and one other testcase that I can't recall off the top of my head... but either way it would be good to have a minimal testcase as `Archive of our own` is kind of a huge page with a lot going on.

Readability.js
robinmetral: Yes, sorry for the oversight! Should be fixed. I'm also suggesting to move the separators into a separate expression to avoid making this mistake in the future (see my message below).

Readability.js
```diff
@@ -593,14 +593,25 @@ Readability.prototype = {
     }
 
     // If there's a separator in the title, first remove the final part
-    if (/ [\|\-\\\/>»] /.test(curTitle)) {
-      titleHadHierarchicalSeparators = / [\\\/>»] /.test(curTitle);
-      let allSeparators = Array.from(origTitle.matchAll(/ [\|\-\\\/>»] /gi));
+    const titleSeparators = /\|\-–—\\\/>»/;
```
gijsk, line 596: As we only ever use `.source`, we could perhaps do that here instead of requesting it repeatedly.

gijsk (approved):
Really sorry for the long delay here. Having some health things going on and my other work responsibilities meant that this got a bit lost. I think this all looks good but had one note. Will see if I can remember enough git fu to update that and then merge.

robinmetral:
Thanks a lot for the improvement and merge @gijsk—no worries for the delay, take care of yourself and I hope you'll get better soon!

PR #971, opened 2025-05-18
Add typescript definitions for Article and ReadabilityOptions objects.

mcepl:


index.d.ts
```diff
@@ -62,3 +62,39 @@ export class Readability<T = string> {
     publishedTime: string | null | undefined;
   };
 }
+
+// Assuming Article is the return type of Readability.prototype.parse()
+export type Article = ReturnType<Readability['parse']>;
+
+export interface ReadabilityOptions {
```
gijsk, line 69: Shouldn't this then be used in the definition of the constructor, above? It also seems like the definition here does not match the definition there, e.g. `nbTopCandidates` is missing here and `headless` is added...

index.d.ts
```diff
@@ -62,3 +62,39 @@ export class Readability<T = string> {
...
+   * @param el The HTMLElement to serialize.
+   * @returns The HTML string representation of the element's content.
+   */
+  serializer?: (el: Node) => string;
+  /**
+   * When `true`, Readability skips some browser-specific operations
+   * (e.g., scroll behavior modification) that might be problematic in
+   * non-browser environments like JSDOM. Defaults to `false`.
+   */
+  headless?: boolean;
```
gijsk, line 98: I don't see any code that actually uses this property? Where has this come from?

gijsk (changes requested):
Thanks for the PR! We don't use typescript in the repo so the definitions are only here as a convenience for consumers, which means I sometimes find it difficult to review them. It would help me understand the PR better if you could describe in a few words why/how these added definitions are helpful.

index.d.ts
mcepl: You are right. Removed.

mcepl:
Also, for explanation. I was converting https://git.sr.ht/~mcepl/readability-cli to TypeScript and it is helpful to have typed primary dependencies.

index.d.ts
```diff
@@ -14,20 +14,82 @@ export function isProbablyReaderable(
...
+   * potentially helping with documents that have many or few links.
+   * Defaults to 1.
+   */
+  linkDensityModifier?: number;
+  /**
+   * An object of callback functions that can be used to extend Readability's behavior.
+   * Currently, no specific callbacks are detailed or widely used.
+   * It's treated as a generic object (`Record<string, unknown>`).
+   */
+  callbacks?: Record<string, unknown>;
```
gijsk, line 82: This doesn't seem to exist?

index.d.ts
```diff
@@ -14,20 +14,82 @@ export function isProbablyReaderable(
...
+  /**
+   * The maximum number of elements to parse. If the document exceeds this,
+   * Readability will stop processing. Useful for performance on very large documents.
+   * Defaults to 0 (no limit).
+   */
+  maxElemsToParse?: number;
+  /**
+   * The number of top candidate nodes to consider when determining the main article content.
+   * A higher number might lead to better results but could increase processing time.
+   * Defaults to 5.
```
gijsk, line 31: Can the default be expressed in TS?

index.d.ts
```diff
@@ -14,20 +14,82 @@ export function isProbablyReaderable(
...
+  /**
+   * An object of callback functions that can be used to extend Readability's behavior.
+   * Currently, no specific callbacks are detailed or widely used.
+   * It's treated as a generic object (`Record<string, unknown>`).
+   */
+  callbacks?: Record<string, unknown>;
+  /**
+   * The base URL for resolving relative URLs. defaults to the document's `baseURI`.
+   */
+  url?: string;
```
gijsk, line 86: I think I've seen a different PR that tried to add this but I don't think it's been merged, unless I missed something?

index.d.ts
mcepl: Isn’t it what’s here https://github.com/mozilla/readability/blob/main/test/test-pages/ehow-2/source.html#L982 ?

index.d.ts
mcepl: I see, I’ll remove it.

index.d.ts
gijsk: Right, the JS there has nothing to do with the module.

gijsk: (it's in the `test-pages` directory which is just test pages from the internet; JS there doesn't interact with the module in any way)

gijsk (changes requested):
Unfortunately this still needs the callbacks bit dropping. I'd make the change myself but pull request changes by maintainers are not turned on. :-)

gijsk (approved):
Thanks!

gijsk:
> > Unfortunately this still needs the callbacks bit dropping. I'd make the change myself but pull request changes by maintainers are not turned on. :-)
> 
> Is it something I could change?

I think it's an option when you create the pull request; unsure if you can change it after - see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork , specifically item #7.

Issue #834, opened 2024-01-14
Can we get a list of images in the article as part of the json response?


ralyodio:
Would be easier than trying to parse html for them

inhumantsar:
i have a use case for this on a downstream project. happy to take this on

gijsk:
Would this just be a list of URLs? Anyway, that seems reasonable.

ralyodio:
Yeah I think that would work.

ralyodio:
Looking forward to it. I got another project where images would be handy.

PR #975, opened 2025-07-21
Extract image URL

LordEidi:
I noticed that although the meta data of websites is used for title and other info, there is no extraction for the image URL (nor favicon). This pull request tries to get the image URL from a given website. 

Please note that based on the nature of how the regex regarding metadata extraction works, the order of the new _og:image_ and _og:image:alt_ search in the regex is important.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -455,23 +455,6 @@ Readability.prototype = {
    * @return void
    */
   _fixRelativeUris(articleContent) {
-    var baseURI = this._doc.baseURI;
-    var documentURI = this._doc.documentURI;
-    function toAbsoluteURI(uri) {
-      // Leave hash links alone if the base URI matches the document URI:
-      if (baseURI == documentURI && uri.charAt(0) == "#") {
-        return uri;
-      }
-
-      // Otherwise, resolve against base URI:
-      try {
-        return new URL(uri, baseURI).href;
-      } catch (ex) {
-        // Something went wrong, just return the original:
-      }
-      return uri;
-    }
-
     var links = this._getAllNodesWithTag(articleContent, ["a"]);
     this._forEachNode(links, function (link) {
       var href = link.getAttribute("href");
@@ -495,7 +478,7 @@ Readability.prototype = {
             link.parentNode.replaceChild(container, link);
           }
         } else {
-          link.setAttribute("href", toAbsoluteURI(href));
+          link.setAttribute("href", this._toAbsoluteURI(href));
         }
       }
     });
@@ -515,19 +498,17 @@ Readability.prototype = {
       var srcset = media.getAttribute("srcset");
 
       if (src) {
-        media.setAttribute("src", toAbsoluteURI(src));
+        media.setAttribute("src", this._toAbsoluteURI(src));
       }
 
       if (poster) {
-        media.setAttribute("poster", toAbsoluteURI(poster));
+        media.setAttribute("poster", this._toAbsoluteURI(poster));
       }
 
       if (srcset) {
         var newSrcset = srcset.replace(
           this.REGEXPS.srcsetUrl,
-          function (_, p1, p2, p3) {
-            return toAbsoluteURI(p1) + (p2 || "") + p3;
-          }
+          (_, p1, p2, p3) => this._toAbsoluteURI(p1) + (p2 || "") + p3
         );
 
         media.setAttribute("srcset", newSrcset);
@@ -1774,7 +1755,7 @@ Readability.prototype = {
 
     // property is a space-separated list of values
     var propertyPattern =
-      /\s*(article|dc|dcterms|og|twitter)\s*:\s*(author|creator|description|image:alt|image|published_time|modified|title|site_name)\s*/gi;
+      /^\s*(article|dc|dcterms|og|twitter)\s*:\s*(author|creator|description|image|published_time|modified|title|site_name)\s*$/i;
 
     // name is a single value
     var namePattern =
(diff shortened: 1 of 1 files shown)


LordEidi:
Forgot to mention: This Pull Request addresses #834. At least in parts.

Readability.js
```diff
@@ -1854,6 +1854,12 @@ Readability.prototype = {
     // get site name
     metadata.siteName = jsonld.siteName || values["og:site_name"];
 
+    // get image thumbnail
+    metadata.image = values["og:image"] || values.image || values["twitter:image"];
+
+    // get favicon
+    // metadata.favicon = this._getArticleFavicon()
```
gijsk, line 1861: Can we drop this change? It doesn't seem to be doing anything.

Readability.js
```diff
@@ -1854,6 +1854,12 @@ Readability.prototype = {
     // get site name
     metadata.siteName = jsonld.siteName || values["og:site_name"];
 
+    // get image thumbnail
+    metadata.image = values["og:image"] || values.image || values["twitter:image"];
```
gijsk, line 1858: Do we also need to check `jsonld` for this? Would more work be required to extract this info from the JSON-LD bits?

What about the other prefixes, can images occur there?

Readability.js
```diff
@@ -1771,7 +1771,7 @@ Readability.prototype = {
 
     // property is a space-separated list of values
     var propertyPattern =
-      /\s*(article|dc|dcterm|og|twitter)\s*:\s*(author|creator|description|published_time|title|site_name)\s*/gi;
+      /\s*(article|dc|dcterm|og|twitter)\s*:\s*(author|creator|description|image:alt|image|published_time|title|site_name)\s*/gi;
```
gijsk, line 1774: In the PR comment you wrote:

> Please note that based on the nature of how the regex regarding metadata extraction works, the order of the new og:image and og:image:alt search in the regex is important.

However, I'm a bit confused why `og:image:alt` is in here at all - we don't seem to use it later on? Is this just to avoid treating `og:image:alt` as equivalent/overriding `og:image` if both are present, because the `og:image` regex would still match the `og:image:alt` instance, only excluding the "alt" part?

A code comment about why this is here may help future readers. :-)

gijsk (changes requested):
Thanks for the patch!

I left some minor comments inline.

More importantly, I think some bits are missing here, unfortunately. I think we should have at least a little bit of automated test coverage. I don't have enough visibility into whether this should include JSON-LD; if there's a good reason to push that into a follow-up ticket/PR then that's fine but perhaps it could be included here if it's straightforward to do / if there are already examples of that in the existing test pages.

On the whole this part of the API is unfortunately not documented nor included in the type annotations; fixing that would also be useful but not required for this PR.

Readability.js
LordEidi: I just pushed another change, containing an extraction for favicons.

Readability.js
LordEidi: This needs a bit of explanation.

Looking into some BBC page, you find these lines (without context):

```
<meta data-rh="true" property="og:image" content="https://ichef.bbci.co.uk/ace/branded_news/1200/cpsprodpb/b98b/live/2554ced0-678d-11f0-8dbd-f3d32ebd3327.jpg"/>
<meta data-rh="true" property="og:image:alt" content="Bandmates Tony Iommi, Ozzy Osbourne and Geezer Butler pose in front of yellow and white Grammy logos while holding a grammy award for Best Metal Performance at the 2014 Grammy awards"/>
```

If you look at the regex on line 1776 in Readability.js (before my change), you notice that it does _not_ contain a check for neither the start nor the end of the property name.

```
var propertyPattern =
      /\s*(article|dc|dcterm|og|twitter)\s*:\s*(author|creator|description|image|published_time|title|site_name)\s*/gi;

```

If you check the lines from above for og:image, you will get two hits. one for every line. Because both lines contain og:image. While the first one would be a "direct" / complete hit, the second hit is not, it only matches a substring. Since there is no check for beginning/end of the property name.

In lines 1784 and following there is a loop over all meta elements and ultimately over all properties that return a hit for the regex. If there is a hit, the content is then placed in the field named after the hit. 

The first line works, we get a hit for og:image, place the uri in the content as value for the property og:image.

The second line hits as well. But since we use the part of the regex that hits as a name for the property to place the content of the property into, we replace the value from the step above with (in this case) a note about Ozzy. Correct were to place that note into the property it belongs into: og:image:alt. 

Placing the _image:alt_ part right before the _image_ part in the regex, we make sure that the first one hits first and the value is placed in the correct property (og:image:alt) and not into og:image.

True, this is some kind of hack. But if you want to fix this, you will have to completely refactor the whole _getArticleMetadata function. I tried to avoid that since this would be more intrusive and a lot more work to do. For me as a newcomer regarding Readability.js sources, probably not the best place to start with. 

About your note re comment. The explanation needs some space. What about just linking this PR instead?

gijsk (changes requested):
Thanks for the iteration here, and especially catching that `dcterms` typo, that's very unfortunate.

I noticed there's still an issue with the favicon code (see inline comments). In a roundabout way, `eslint` also noticed this and complained.

Unfortunately our infra is set up such that it then (after the eslint failure) didn't run formatting (which would also have picked up a few things) or tests... it would be interesting to see if any test output/metadata changes as a result of this.

It would also be good to have at least some test coverage for the new properties you're adding here - I would imagine some existing testcases will already have favicons that you can use to add this test coverage (probably by adding checks in `test/test-readability.js` that already runs for all the pages in `test/test-pages`, we wouldn't necessarily need more testcases.

Readability.js
gijsk: Linking to the PR would be fine, thanks!

Readability.js
```diff
@@ -1854,24 +1857,95 @@ Readability.prototype = {
...
+      // make sure the type is correct and element contains a href attribute
+      var rel = element.hasAttribute("rel") ? element.getAttribute("rel") : "";
+      if (rel === "icon" && element.hasAttribute("href")) {
+        favicon = element.getAttribute("href");
+
+        var type = element.hasAttribute("type") ? element.getAttribute("type") : "";
+        if(type === "image/svg+xml")
+        {
+          // svg wins as best quality format
+          return this._toAbsoluteURI(favicon);
```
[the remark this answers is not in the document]
gijsk: I think this is still broken, unfortunately? The code wants to use the `return this._toAbsoluteURII(favicon)` as the return value for `_getArticleFavicon`. But this is inside the (function) callback provided to `this._forEachNode`, and the return value is not used by `_forEachNode`, see https://github.com/mozilla/readability/blob/1f0ec42686c89a4a1c71789849f7ffde349ab197/Readability.js#L349 . So it's just going to be dropped on the floor.

To make this work, switch from using `_forEachNode` to using a manual loop, `for (var i = 0; i < metaElements.length; i++)` etc. - this is fine in this loop as we're not modifying the DOM while we're looping over these elements.

PR #979, opened 2025-08-25
Consider blockquote in article scoring

tuyafeng:
Some pages have long quoted sections but only a few lines of commentary or intro. Right now, `isProbablyReaderable` ignores `blockquote`, which can make the examples below score 0 (the default options), whereas including `blockquote` would give a score of 9.

Example from [W3Schools](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_blockquote_test):

```html
<!DOCTYPE html>
<html>
<body>

<h1>The blockquote element</h1>

<p>Here is a quote from WWF's website:</p>

<blockquote cite="http://www.worldwildlife.org/who/index.html">
For 50 years, WWF has been protecting the future of nature. The world's leading conservation organization, WWF works in 100 countries and is supported by 1.2 million members in the United States and close to 5 million globally.
</blockquote>

</body>
</html>
```

I’m thinking it might make sense to factor blockquote into the scoring. If not, I’d love to hear the reasoning—thanks!

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -63,7 +63,7 @@ function isProbablyReaderable(doc, options = {}) {
   };
   options = Object.assign(defaultOptions, options);
 
-  var nodes = doc.querySelectorAll("p, pre, article");
+  var nodes = doc.querySelectorAll("p, pre, blockquote, article");
 
   // Get <div> nodes which have <br> node(s) and append them into the `nodes` variable.
   // Some articles' DOM structures might look like


gijsk:
Closing given the lack of response here, to try to get a grip on the pile of PRs here.

tuyafeng:
Sorry for the late reply. I completely forgot about this PR, and a lot has happened over the past six months.

I’m still not sure whether this change is the best approach. It does make the detection process slightly slower, but it seems necessary for some websites:

https://simonwillison.net/2026/Jul/3/josh-w-comeau/
https://simonwillison.net/2026/Jul/8/kenton-varda/
https://www.sgfox.cc/archives/via-shisui.html

Maybe we could introduce an option in the `options` object to control this behavior (for example, an `includeTags` field). This would make the implementation more flexible and leave room for future extensions.

PR #981, opened 2025-09-06
Updated instructions for permission errors for the tidy binary

hollisticated-horse:
Changed the mention to include Linux as a platform where the issue may arise with the `tidy` binary and EACCESS errors.


gijsk (approved):
Thanks!

PR #982, opened 2025-09-06
WIP permalink anchor stripping

### LINKED ISSUE #965: Anchor permalinks display issues
When parsing this page: [https://unplannedobsolescence.com/blog/less-htmx-is-more/](https://unplannedobsolescence.com/blog/less-htmx-is-more/), the permalink images are huge.

Image

I feelt that they should be either properly styled and sized, or completly removed.

hollisticated-horse:
This PR aims to fix #965. 

The initial method would be to remove the SVG files or element, and replace it with either a `#`.
The presence of a hidden span with permalink could also be a pattern for recognition.

However, I am unable to create a test from a source file locally. I run it with the node command, but the content is not generated...

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -828,6 +828,7 @@ Readability.prototype = {
     this._clean(articleContent, "select");
     this._clean(articleContent, "button");
     this._cleanHeaders(articleContent);
+    this._cleanHeadingPermalinks(articleContent);
 
     // Do these last as the previous stuff may have removed junk
     // that will affect these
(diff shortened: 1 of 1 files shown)


gijsk (changes requested):
Thanks for working on this and apologies for the slow response! I've been sidetracked by various other things including some time off.

The code change looks reasonable though I would probably not replace the SVG with anything at all?

> However, I am unable to create a test from a source file locally. I run it with the node command, but the content is not generated...

Can you say more about this? What exactly are you running and what is the result?

Off-hand I would expect `npm run generate-testcase <slug>` to regenerate the testcase, and this WFM on my machine. Do you get an error instead?

I think updating the testcases is the only other thing that would stop this from being merged. 🙂

hollisticated-horse:
> Thanks for working on this and apologies for the slow response! I've been sidetracked by various other things including some time off.

Yup, I understand completely. I would have eventually closed this PR in six months time :grin: 

> The code change looks reasonable though I would probably not replace the SVG with anything at all?

My rational for replacing the svg with a "#" is to keep the anchor. I believe I saw some tests that expected anchors to stay and be displayed as #.


> Can you say more about this? What exactly are you running and what is the result?

I'll get back to you on that, I seem to remember having managed to create the test case. 

> I think updating the testcases is the only other thing that would stop this from being merged. 🙂

That part scares me a bit. Can I modify expected content in those test cases that fail ? What are the guidelines for that ?

PR #985, opened 2025-09-16
Fix paragraph wrapping logic and improve DOM mock

samuelhuang:
Fix paragraph wrapping logic and improve DOM mock

The core logic for wrapping phrasing content in paragraphs was flawed,
leading to incorrect HTML output in several edge cases. This change
refactors that logic and improves the test environment's DOM mock to
be more robust and compliant.

The original implementation had two main bugs:
1. It failed to trim trailing whitespace from a paragraph if the
   phrasing content was at the very end of a container `<div>`.
2. It would leave leading whitespace nodes behind as direct children
   of the `<div>` instead of discarding them.

This was fixed in `Readability.js` by replacing the complex `while`
loop with a robust "collect and transform" pattern using a
`DocumentFragment`. This new logic correctly gathers all consecutive
phrasing content, trims any leading or trailing whitespace, and
only then wraps the result in a `<p>` tag.

This refactoring exposed several limitations and latent bugs in the
lightweight DOM mock, `JSDOMParser.js`. The following improvements
were made:
* Added `DocumentFragment` support by implementing the object and the
  `doc.createDocumentFragment()` method.
* Refactored DOM insertion logic (`appendChild()`, `insertBefore()`,
  `replaceChild()`) into a single, efficient `_insertNodesBefore()`
  private helper to ensure correctness and consistency.
* Corrected a bug in the `remove()` method where it was incorrectly
  modifying properties on non-element nodes.
* Made `replaceChild()` robust by correctly handling the edge case
  of a node being replaced by itself.

Finally, the test suite in `test-jsdomparser.js` was updated to
validate these improvements by adding new test cases for the "move"
behavior of `insertBefore()`, a regression test for the bug
found and fixed in the `remove()` method, and a test for the
`replaceChild(node, node)` edge case.


JSDOMParser.js
```diff
@@ -347,26 +347,111 @@
...
+        if (nodes[i].parentNode) {
+          nodes[i].remove();
+        }
       }
 
-      var last = this.lastChild;
-      if (last) {
-        last.nextSibling = child;
+      var refIndex = referenceNode
+        ? this.childNodes.indexOf(referenceNode)
```
brandonawylie, line 363: Are these the top-level children? What if the referenceNode is a sub-child?

JSDOMParser.js
```diff
@@ -347,26 +347,111 @@
       return this.children[this.children.length - 1] || null;
     },
 
-    appendChild(child) {
-      if (child.parentNode) {
-        child.remove();
+    _insertNodesBefore(nodes, referenceNode) {
```
brandonawylie, line 350: Can you add some documentation for this method? What nodes and referenceNode is and what happens if referenceNode is null or not found?

JSDOMParser.js
```diff
@@ -347,26 +347,111 @@
...
-      if (child.nodeType === Node.ELEMENT_NODE) {
-        child.previousElementSibling =
-          this.children[this.children.length - 1] || null;
-        this.children.push(child);
-        child.previousElementSibling &&
-          (child.previousElementSibling.nextElementSibling = child);
+
+      // Find the element reference node.
+      var refElem = referenceNode;
+      while (refElem && refElem.nodeType !== Node.ELEMENT_NODE) {
```
brandonawylie, line 371: I don't understand this while loop, what are you doing here?

brandonawylie (commented):
I think overall JSDOMParser could use a little more inline documentation and clarity, but your changes look good. Let's send this over to gijsk to see what they think.

JSDOMParser.js
samuelhuang: Added documentation to reflect the changes. I also changed the interface, so callers can do more checking (and emit the right errors).

JSDOMParser.js
```diff
@@ -347,26 +347,129 @@
...
+
+        // Store the previous element sibling before more DOM modifications.
+        var prevElem = afterElem
+          ? afterElem.previousElementSibling
+          : this.lastElementChild;
+
+        var afterElemIndex = afterElem ? this.children.indexOf(afterElem) : -1;
+        var elemInsertionPoint =
+          afterElemIndex === -1 ? this.children.length : afterElemIndex;
+        Array.prototype.splice.apply(
```
gijsk, line 429: Same here.

Readability.js
```diff
@@ -1174,23 +1174,39 @@ Readability.prototype = {
...
-                p = doc.createElement("p");
-                node.replaceChild(p, childNode);
-                p.appendChild(childNode);
+              var fragment = doc.createDocumentFragment();
+              // Collect all consecutive phrasing content into a fragment.
+              do {
+                nextSibling = childNode.nextSibling;
+                fragment.appendChild(childNode);
+                childNode = nextSibling;
+              } while (childNode && this._isPhrasingContent(childNode));
```
gijsk, line 1187: I'm curious why you decided to do this this way. The previous code set up a `p` outside the loop. Now there is a loop inside the loop, that still "advances" the outer loop (by manipulating `nextSibling` and `childNode`). That feels a little fragile / hard to follow, and ISTM you could use the fragment strategy but the same logic/structure that existed before. I'd find that easier to read as it maintains having one pass/loop through the nodes, rather than the nested loop.

Here's a gist to illustrate what I mean: https://gist.github.com/gijsk/c3fcb07d3d193a88626e455adafca49c (untested, written in github, ymmv etc.)

test/test-jsdomparser.js
```diff
@@ -223,6 +223,172 @@ describe("Test JSDOM functionality", function () {
...
+    var newEl = doc.createElement("hr");
+    body.insertBefore(newEl, foo);
+    nodeExpect(p.nextSibling, newEl);
+    nodeExpect(newEl.nextSibling, foo);
+    nodeExpect(foo.previousSibling, newEl);
+    nodeExpect(newEl.previousSibling, p);
+    nodeExpect(p.nextElementSibling, newEl);
+    nodeExpect(newEl.nextElementSibling, foo);
+    nodeExpect(foo.previousElementSibling, newEl);
+    nodeExpect(newEl.previousElementSibling, p);
```
gijsk, line 244: Can we also add a test where the previous/next sibling vs element sibling are different?

JSDOMParser.js
```diff
@@ -347,26 +347,129 @@
...
+
+      // Store the previous sibling before we modify the DOM.
+      var prevSibling = afterSibling
+        ? afterSibling.previousSibling
+        : this.lastChild;
+
+      // Insert nodes into childNodes.
+      var insertionPoint = index === -1 ? this.childNodes.length : index;
+      Array.prototype.splice.apply(
+        this.childNodes,
```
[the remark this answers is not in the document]
samuelhuang: Yup, this is doable in spread syntax. But since the file uses old JS constructs (e.g., `.prototype` instead of `class`), so I think it would be better to follow style? Will switch to spread syntax if you insist.

test/test-jsdomparser.js
samuelhuang: Done.

Readability.js
samuelhuang: The helper approach is fine; a nitty consideration would be that code would be moved farther away if wrapped in `function` or even a lambda outside the loop?

My main preference is re. state machine: Previously the state was explicit, and one run loop exists. For the proposed code, state is implicitly represented by where the current instruction is, which I think is more clear. Though caveat is the usage of `do...while`, and having 2 rather verbose trim loops.

gijsk:
> Question: I have the 4 CLs with short comments merged onto main. But my intent is for these to be ephemeral progress in my clone, and then have everything squashed into 1 CL, with the first comment of this thread to be the detailed commit comment?

In future, if this is important to you, please make this clearer ahead of time.

I intended to use "squash", I must have misclicked, and that's on me - but even if I hadn't, github would not have offered to do what you asked for, but just concatenated the commit messages of your 4 commits. I would probably have altered the message to avoid that, but I doubt I would have included all of your initial comment. In any case, as I have now found in practice, asking github for the raw markdown HTML-escapes some of the comment, which is unfortunate. I noticed one of those and fixed manually, but missed another...

> How do we fix this (if possible), and do squash in the future? Thanks!

I made an exception and force pushed to main. I'd much prefer not to do this in future.

samuelhuang:
Thanks for the fix! I'm used to Gerrit and didn't know about the GitHub process. So in the future perhaps I can do the following in my cloned repo?
* Make branch `feature`, set its commit comment to the full one I want on the trunk. When I update I force push.
* Make branch `feature_h` to track change history of `feature`, with short comments.

gijsk:
Honestly, I'm not sure - I'm also not used to github, I'm used to phabricator...

For what you describe, which branch would you push to github? I'd try to just force-push the PR branch every time; I'm assuming github knows how to show the interdiffs regardless, and if not I guess that'll be my lookout... If you want the interdiffs yourself locally and git does not provide them, something like [jj](https://jj-vcs.github.io/jj/latest/tutorial/) can help - that's what I use, anyway.

samuelhuang:
In the case I described I'd just send `feature` out for merge request, and if there's any change, I'd force push onto `feature` my clone, so when merge happens it will just be one commit? Though you've squashed my changes and applied the big comment. Aside from the forced push part (which we should try to avoid), would this be hard to do?

PR #993, opened 2025-12-03
Remove deeply nested divs around img/noscript combinations

gijsk:
Alternative to #989 .

This seems to also work, and also positively impacts some existing testcases. @samuelhuang could you confirm if this looks reasonable to you? There are minor markup changes in the `expected.html` for the new `allrecipes-1` testcase that you added (that strip the extraneous containers around the `img`).


gijsk:
(looks like maybe I should update the extended commit msg for [03abc3f](https://github.com/mozilla/readability/pull/993/commits/03abc3f12306127098d6445d0adc58c7d5980e37) still, before merging this)

samuelhuang:
NIT: Add function-level comment?

Optional NIT: Perhaps `_removeDeeplyNestedImageDivs()` to have `unwrap()` somewhere?

Optional NIT: The outer `while (cond1)` loop and its immediately preceding `if (cond2) { break; }` can be merged into `while (cond1 && !cond2) {}`

Optional NIT: The unwrap loop (gut out `parent` content to its its previous siblings, then removing `parent`) seem to appear elsewhere as well; maybe extract as helper?

Otherwise LGTM.

samuelhuang:
Hi, an updates in landing this change? Thanks!

Issue #863, opened 2024-04-26
H1, H2, H3, etc, always get converted to H2 and document loses structure


yagudaev:
I was also surprised by how Readability handles headings.

The demo below converts a page to markdown. It first uses readability to eliminate other necessary content.

However, the heading information h1, h2, h3, is destroyed and is all h2 at the end.

![CleanShot 2024-04-18 at 16 29 51@2x](https://github.com/mozilla/readability/assets/1386966/577fc5fd-e107-4895-a038-c83d4d93c5be)

Is there a way to turn this off?

[...]

PR #1006, opened 2026-04-17
Optionally keep original title headers for main content extraction accuracy

mcPear:
### Summary

Reader extraction currently rewrites all in-article `h1` elements to `h2` so the article title can remain the sole top-level heading in the reader UI. Moreover, removes the first similar heading spotted after the title. These normalizations improve classic “reader mode” presentation but **weaken the semantic outline** of the page: crawlers, SEO tooling, and systems that infer structure from HTML (including retrieval and “reverse engineering” of how a page is organized) rely on **stable heading levels** that match the publisher’s markup.

This change **preserves the original heading tag names and levels** in the extracted content wherever we are not explicitly removing noise, so the serialized article HTML stays closer to the source document’s hierarchy. All that is **gated behind an option.**

### What changes (high level)

- Stop blanket `h1` → `h2` replacement in article content
- Stop duplicate-title header removal
- Add unit tests

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -64,6 +64,12 @@ function Readability(doc, options) {
   this._disableJSONLD = !!options.disableJSONLD;
   this._allowedVideoRegex = options.allowedVideoRegex || this.REGEXPS.videos;
   this._linkDensityModifier = options.linkDensityModifier || 0;
+  /**
+   * If true, keep the first in-article H1/H2 that duplicates the article title
+   * and leave H1 tags in the extracted content. Defaults to false (strip the
+   * duplicate title header and normalize remaining H1 elements to H2).
+   */
+  this._keepOriginalTitleHeaders = !!options.keepOriginalTitleHeaders;
 
   // Start with all flags set
   this._flags =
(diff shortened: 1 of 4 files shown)


gijsk (changes requested):
Thanks for submitting a PR.

I'm confused about this being labeled a "chore". And I think that the relabeling of h1 to h2 and removing headers that match the titles should be split into different options. The h1-to-h2 replacement would fix #863 . I would be tempted to make not swapping it the default, though it would potentially break existing consumers - I can fix Firefox but not sure about others.

gijsk (changes requested):
Apologies for the very unreasonable delay here.

I'm fairly confused at this point. The updated patch here added a lot more code which is a bit hard to read, and also doesn't seem consistent with what I asked in my previous review. Instead, it seems to change the approach used to keep the headers. Why is that?

mcPear:
Sorry, I had to adjust the logic to fix corner cases we found in our app. I wasn't aware you will see it as kinda response to this conversation. I'm closing it as for now.

PR #1009, opened 2026-05-06
Revise LICENSE.md to include full Apache License 2.0

bohemian-miser:
Updated the Apache License text to include detailed terms and conditions for use, reproduction, and distribution.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,13 +1,204 @@
 Copyright (c) 2010 Arc90 Inc
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
 
-   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


gijsk:
The text in the license in the PR explicitly says that "how to license" is to append the text at the bottom... which is already there. So I don't see why this would be an improvement.

bohemian-miser:
Those instructions about applying the license are in reference to creating the license headers of individual source files. 

If you take a look at section 4.a. of the Apache 2.0 license, it clarifies that a copy of the license needs to travel with the software whenever it is redistributed. The instructions at https://www.apache.org/licenses/LICENSE-2.0#apply say "Include a copy of the Apache License, typically in a file called LICENSE, in your work, and consider also including a NOTICE file that references the License."

bohemian-miser:
@gijsk please see my previous comment. The license states that the full text is required.

PR #1010, opened 2026-05-11
Include Full Apache-2.0 License text

bohemian-miser:
Section 4.a. of the Apache 2.0 license, states that a full copy of the license needs to travel with the software whenever it is redistributed.

I pointed this out in the previous pull request, but the PR was already closed and ignored https://github.com/mozilla/readability/pull/1009 

The confusion came from the appendix regarding license headers in source code and was misread as applying to the project.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,13 +1,204 @@
 Copyright (c) 2010 Arc90 Inc
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
 
-   http://www.apache.org/licenses/LICENSE-2.0
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


gijsk (changes requested):
Thanks for the renewed PR.

For context, I did not intentionally ignore your PR or comments, I have simply been busy and didn't get around to yours yet.

My understanding here is that if we're updating things, ideally the LICENSE file should contain the contents of https://www.apache.org/licenses/LICENSE-2.0.txt verbatim with no changes, and a NOTICE file should be used to include copyrights (instead of leaving them at the top of the license file).

Would you mind updating the PR accordingly?

bohemian-miser:
LICENSE.md is now a verbatim copy of the Apache-2.0 license and I've added a NOTICE file

gijsk (approved):
Apologies for the delay here, I have been swamped with various other things.

Issue #997, opened 2025-12-20
PHRASING_ELEMS missing <u>, <s>, and <bdi> elements causes paragraph splitting


gregko:
## Problem

The `PHRASING_ELEMS` array is missing several HTML5 phrasing content elements: `U`, `S`, and `BDI`. 

This causes Readability to incorrectly treat these inline elements as block-level, splitting paragraphs around them.

## Example

**Input HTML:**
```html
<p>Seigneur, mon Die<u>u</u> et mon salut</p>
```

Output (incorrect):

```html
<p>Seigneur, mon Die</p><u>u</u><p> et mon salut</p>
```

The &lt;u&gt; tag is being treated as a block element, breaking the paragraph into three parts.

Solution
Add the missing phrasing content elements to PHRASING_ELEMS:

```
 PHRASING_ELEMS: [
   // ... existing elements ...
  "BDI",  // Bidirectional text isolation
  "S",    // Strikethrough
  "U",    // Underline
   // ...
 ]
```
Per [HTML spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html), these are all valid phrasing content elements that should be kept inline within paragraphs.

PR #1011, opened 2026-05-19
fix(Readability): treat S and BDI as phrasing content

SAY-5:
`PHRASING_ELEMS` is missing the HTML5 phrasing elements `S` and `BDI`, so Readability treats them as block-level and splits paragraphs around them (e.g. `<div>foo <s>bar</s> baz</div>` becomes `<p>foo </p><s>bar</s><p> baz</p>`). This adds both to `PHRASING_ELEMS` with a regression test.

Towards #997. I left `U` out for now because adding it changes the extracted output of the hukumusume test page, so that one is better handled separately.


gijsk (changes requested):
Thanks for the patch.

The testcase is a bit strange. We would typically add a minimal testcase to the testcases folder instead, with source and expected output.

I also think that we should include `u` and update the testcases accordingly (`generate-testcase.js` can do this for you if you delete the expected.html file first and then run it with the testcase slug as an argument). Or at least - I would like to see what changes that causes, before deciding whether any of these changes make sense. In isolation it is hard to see what this is really accomplishing.

SAY-5:
Rebased on main so the harness runs again, and hukumusume is green on both parsers now.

Serializing the source through jsdom didn't actually change that region, so I dug further: the mismatch wasn't the parsers disagreeing on the source, it was that Readability's output there contains a p inside a p (the outer font/span run wraps, and the inner u > font > span run wraps again), and no HTML parser can round trip that, so expected.html could never match on the jsdom leg. The two br tags inside the inner `u > font size="-2"` were the only reason the inner run got its own paragraph, so I dropped them from source.html and regenerated. Full suite is 1997 passing locally.

PR #1013, opened 2026-06-21
Fix: handle JSON-LD @type as array in _getJSONLD

JSap0914:
## Problem

The [JSON-LD specification](https://www.w3.org/TR/json-ld/#specifying-the-type) allows the `@type` property to be **either** a plain string (`"NewsArticle"`) **or** an array of strings (`["NewsArticle", "Article"]`). The second form is common in the wild — for example when a page is both a `BlogPosting` and an `Article`.

`Readability._getJSONLD` calls `.match()` directly on the `@type` value in three places. `Array.prototype.match` does not exist, so when `@type` is an array the call throws a `TypeError`. The surrounding `try/catch` swallows the error silently, leaving the JSON-LD block completely ignored. Every metadata field — title, byline, excerpt, siteName, publishedTime — then falls back to (usually noisier) scraped HTML values.

## Fix

Add a small `matchesArticleType(type)` helper inside `_getJSONLD` that normalises `@type` to an array before testing each element against `REGEXPS.jsonLdArticleTypes`. Replace the three existing `.match()` call-sites that operated directly on `@type` with this helper.

## Test

Add a new test page `test/test-pages/jsonld-array-type/` whose JSON-LD block uses `@type: ["NewsArticle", "Article"]`. Without the fix, all five metadata fields fail; with the fix, all pass.

```
Before: 10 failing (title, byline, excerpt, siteName, publishedTime — jsdom + JSDOMParser)
After:  15 passing (all new tests green, all 1984 existing tests still pass)
```

> AI-assisted contribution.

### DIFF
--- a/Readability.js
+++ b/Readability.js
@@ -1659,6 +1659,17 @@ Readability.prototype = {
     var scripts = this._getAllNodesWithTag(doc, ["script"]);
 
     var metadata;
+    var self = this;
+
+    // @type can be a string or an array of strings per the JSON-LD spec.
+    function matchesArticleType(type) {
+      var types = Array.isArray(type) ? type : [type].filter(Boolean);
+      return types.some(function (t) {
+        return (
+          typeof t === "string" && !!t.match(self.REGEXPS.jsonLdArticleTypes)
+        );
+      });
+    }
 
     this._forEachNode(scripts, function (jsonLdElement) {
       if (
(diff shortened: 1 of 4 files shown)


Readability.js
```diff
@@ -1659,6 +1659,17 @@ Readability.prototype = {
...
 
     var metadata;
+    var self = this;
+
+    // @type can be a string or an array of strings per the JSON-LD spec.
+    function matchesArticleType(type) {
+      var types = Array.isArray(type) ? type : [type].filter(Boolean);
+      return types.some(function (t) {
+        return (
+          typeof t === "string" && !!t.match(self.REGEXPS.jsonLdArticleTypes)
```
gijsk, line 1669: Doesn't seem like we need both `.filter` here and then a typeof check below - can we drop the filter?

gijsk (changes requested):
The PR description says that the array variation is "common" - can you give some high profile examples where this happens? I don't see it on some news sites that I spot-checked - they all use a single string.

I also note some superfluous work that the patch is doing inline.

PR #1015, opened 2026-07-07
Fix #1012: Prevent aggressive stripping of below/bellow header content wrappers

### LINKED ISSUE #1012: knowablemagazine.org - Does not show the article, republish message displayed after enabling reader mode
**Environment:**
Operating system: Linux/Windows 10
Firefox version: Firefox 152/154

**Steps to reproduce:**
1. Navigate to: https://knowablemagazine.org/content/article/mind/2026/why-we-crave-social-interaction?utm_source=firefox-newtab-en-us
2. Enable reader mode.

**Expected Behavior:**
Article displayed in reader mode.

**Actual Behavior:**
Displays only a republish message.

Screenshot: 

Image

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=2045019

hanumanraj07:
Fixes #1012

**Description**
Currently, Readability fails to parse articles on sites like `knowablemagazine.org` where the main content is wrapped inside a container with a class such as `bellowheadercontainer`. 

The `unlikelyCandidates` regex aggressively strips out elements containing the word `header` without checking for word boundaries. As a result, structural containers using class names like `below-header` or `bellowheader` are prematurely removed from the DOM, causing Readability to completely miss the main article content and fallback to parsing unrelated footer elements.

This PR adds `below` and its common typo `bellow` to the `okMaybeItsACandidate` safeguard regex. This ensures that classes indicating placement beneath a header (which frequently wrap the main article content) bypass the `unlikelyCandidates` purge and receive a proper chance to be scored.

**Testing**
- All 1,984 automated tests pass successfully. 
- Verified that `okMaybeItsACandidate` correctly safeguards the article container to resolve the parsing failure in #1012.

### DIFF
--- a/Readability-readerable.js
+++ b/Readability-readerable.js
@@ -24,7 +24,7 @@ var REGEXPS = {
   // Readability.js. Please keep both copies in sync.
   unlikelyCandidates:
     /-ad-|ai2html|banner|breadcrumbs|combx|comment|community|cover-wrap|disqus|extra|footer|gdpr|header|legends|menu|related|remark|replies|rss|shoutbox|sidebar|skyscraper|social|sponsor|supplemental|ad-break|agegate|pagination|pager|popup|yom-remote/i,
-  okMaybeItsACandidate: /and|article|body|column|content|main|mathjax|shadow/i,
+  okMaybeItsACandidate: /and|article|body|column|content|main|mathjax|shadow|below|bellow/i,
 };
 
 function isNodeVisible(node) {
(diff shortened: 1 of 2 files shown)


gijsk (changes requested):
Thanks for the PR.

Unfortunately, it doesn't seem reasonable to treat "below" as semantically indicating "this is important article text" - even worse for "bellow".

Is there no other markup that can be used to make this determination? And/or, is there evidence that this class name structure is common for other sites than `knowablemagazine.org`, due to common CMSes or similar? 

Also, please add a testcase to the testcases folder for this so we don't regress it in future - you can use the `generate-testcase.js` script to do this easily.

gijsk (changes requested):
Doesn't look like the PR changed since last time? I already left a comment before.

hanumanraj07:
I'm closing this PR. After further investigation, adding below and bellow to okMaybeItsACandidate is too generic and doesn't semantically indicate the main article content.

We tried an alternative approach (adding word boundaries to header in unlikelyCandidates like \bheader\b) which correctly safeguards bellowheadercontainer, but this unfortunately broke several existing test cases (like the SRE book) where valid sub-headers were no longer stripped out.

Given that knowablemagazine.org hides its real content behind Cloudflare's CAPTCHA block and doesn't appear to have any other standard semantic markup on the article container, this seems to be a site-specific parsing failure that is not safely fixable on a global level. Thanks for your review!

Issue #1016, opened 2026-07-09
understandingwar.org shows end notes instead of main body


gijsk:
Moved from: https://bugzilla.mozilla.org/show_bug.cgi?id=2051603

**Environment:**
Operating system: Windows
Firefox version: Firefox 152.0.3 (release)

**Preconditions:**
- Clean profile

**Steps to reproduce:**
1. Navigate to: https://understandingwar.org/research/russia-ukraine/russian-offensive-campaign-assessment-june-27-2026/
2. Trigger reader mode from the toolbar and observe

**Expected Behavior:**
Reader mode renders the page correctly

**Actual Behavior:**
Reader mode is broken

**Notes:**
- Reproducible on the latest Firefox Release and Nightly
- Reproducible regardless of the ETP setting
- Works as expected using Chrome

---

Created from webcompat-user-report:c6b45216-30e4-4a32-9819-72e2fd0bfdd1

PR #1022, opened 2026-07-25
Fix plain text links in citations

aakashsbhatia2:
This fixes issue - https://github.com/mozilla/readability/issues/1016

It seems like the issue is in the scoring algorithm. The end links are scored higher than the article content causing it to take precedence. I updated `_getLinkDensity()` to consider urls not wrapped in `<a>` tags as well. Added the source as a test case


Readability.js
```diff
@@ -2140,7 +2143,7 @@ Readability.prototype = {
    * @param Element
    * @return number (float)
    **/
-  _getLinkDensity(element) {
+  _getLinkDensity(element, includePlainTextUrls) {
     var textLength = this._getInnerText(element).length;
```
gijsk, line 2147: Let's maybe cache the return of `getInnerText` which may not be cheap for large nodes, given we now potentially reuse it further down?

gijsk (changes requested):
Thanks for the PR!

This looks interesting and the change makes sense to me. My main question is whether we really need the argument - do we not want to count the plaintext links for all `_getLinkDensity` callers? If not, why not? :-)

gijsk:
Oh, and one minor thing - can we strip scripts from the input HTML to avoid accidentally running them when checking testcases? Thank you!

(if the generate testcase script doesn't do this, it wouldn't hurt to change that... though obviously that's a separate problem)

